PR #3, opened 2017-05-09
avoid ClassCastException if value is BigDecimal

kaikreuzer:


cdjackson:
This won't work though will it?

Firstly, why wouldn't this be a String? I thought the configuration ensures the type is correct?

Second, if this is a BigDecimal, then it will be incorrectly processed as the assumption at the moment is it's a 128bit value defined as a hex value. If it's a number, then it will be decimal, and incorrectly processed.

I did it this way so that the user can specify this as a hex value which is more inkeeping with this definition - a decimal value would be difficult to specify. Maybe there should be a format applied to parameters (I don't think this is available now).

kaikreuzer:
> Firstly, why wouldn't this be a String? I thought the configuration ensures the type is correct?

Good question - I came across a CCE here and definitely had a BigDecimal (which I didn't set manually)...

kaikreuzer:
I guess your issue is that you set this config param as a long and not as a String:
```
                    configuration.put(CONFIGURATION_EXTENDEDPANID, extendedPanId);
```

cdjackson:
Where did you find this - is that in the coordinator handler?  The only place I seem to set it is done like this -:

            configuration.put(CONFIGURATION_EXTENDEDPANID, String.format("%016X", currentExtendedPanId));


Maybe I fixed this in the past few days and haven’t pushed it yet?

cdjackson:
I've repushed the repo - I think this issue is resolved by the change I mentioned earlier in that the parameter is saved as a string now.

cdjackson:
I'll close this since it won't merge now anyway. If the problem is still there let me know...

PR #4, opened 2017-05-09
Changed status detail to NONE as BRIDGE_OFFLINE is not valid (since t…

kaikreuzer:
…his Thing does not depend on another Bridge)

Btw, I do not fully understand all the logic about setting the coordinator to offline and doing heavy configuration changes. Looks pretty risky to me as it messes with the Thing lifecycle a lot...


cdjackson:
I will close this since I re-pushed the repo and now this has conflicts. I'll update the status shortly.

kaikreuzer:
ok, thanks.

PR #10, opened 2017-05-26
Fix `ZigBeeCoordinatorHandler.initialize()`

DuncanvR:
Running the binding today, I encountered a few issues during the initialisation phase. My CC2531EMK Coordinator thing wouldn't come online, so I investigated and found two bugs. The first two commits in this PR fix those; the other two commits just improve the code a little.

The first issue I saw was a `ClassCastException` at `ZigBeeCoordinatorHandler.java:122`. In my setup, the configuration item `CONFIGURATION_EXTENDEDPANID` wasn't a string, but a `BigDecimal` -- even though it should be a string. It turned out to be due to the assignment during the initialisation, right after generating a new extended PAN id. Including the `String.format()` call fixes this.

The second issue I saw was due to the comparison of the network key array with a temporary array. The `equals()` method of an array only checks whether the two are the same, just like `==` would (see [this SO post](https://stackoverflow.com/questions/8777257/equals-vs-arrays-equals-in-java)). Using `Arrays.equals()` instead will compare the array contents.

Furthermore I found that the `initialize()` would update the configuration every time it was called, which I think is unnecessary. Adding a simple condition on `initializeNetwork` around it should avoid that.

Cheers!

Signed-off-by: Duncan van Roermund <<foss@duncanvr.eu>>

### DIFF
--- a/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeCoordinatorHandler.java
+++ b/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeCoordinatorHandler.java
@@ -11,6 +11,7 @@ package org.openhab.binding.zigbee.handler;
 import static org.openhab.binding.zigbee.ZigBeeBindingConstants.*;
 
 import java.math.BigDecimal;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Hashtable;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


cdjackson:
Thanks - I think this all looks ok but I'll take a better look later. Can you please update both the PR and commit with the signature -:

https://github.com/openhab/openhab2-addons/blob/master/CONTRIBUTING.md#sign-your-work

DuncanvR:
Just in case you missed it: I've added the sign-off to the commits and PR.

cdjackson:
Closing this since I think it's covered by other PRs. If I missed anything, please feel free to rebase.

PR #13, opened 2017-06-21
adapted code to match latest ESH APIs

kaikreuzer:

### DIFF
--- a/src/main/java/org/openhab/binding/zigbee/ZigBeeBindingConstants.java
+++ b/src/main/java/org/openhab/binding/zigbee/ZigBeeBindingConstants.java
@@ -16,7 +16,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.TimeZone;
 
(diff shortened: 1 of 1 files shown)


kaikreuzer:
Hey @cdjackson - I am aware that you will push a huge update any time soon, but as I am preparing the inclusion of this binding in the official build, please allow me small fixes for issues that block me from doing so.

cdjackson:
@kaikreuzer if you don't mind I will revert this as this has been completely rewritten and this will now cause conflicts? Is that ok?

kaikreuzer:
Why should you revert it? If your new version does not include that part, it is absolutely fine. All I wanted to achieve is that the CURRENT version in the repo is at least buildable, so it should not be reverted.

cdjackson:
Ok, I expect it to cause conflicts which I will need to resolve as all this is completely changed now.

kaikreuzer:
No worries, just resolve any conflict in the code by choosing YOUR version - just make sure that the result builds nicely :-)

PR #26, opened 2017-11-06
Command Listiners and an Occupancy sensor implementation

saukijan:
Small changes to how discovery service creates a discovery result (PaperUI is not updated with changes to a thing with the same UID) 
Changes to Binding constants (removed unused channel constants) 
Added an Occupancy sensor converter 
Added Occupancy channel definition in _channels.xml 
Added CC2540 flashing instruction RF05 evaluation board (Windows OS only)  in readme.md
Tested the binding with  FLS-PP Ip  light control unit

And did some clean up from my old, uncommitted project...

Signed-off-by: Dovydas Girdvainis <Dovydas.Girdvainis@hahn-schickard.de>


saukijan:
Changed the _classpaths_, but the binaries still differ because of my small changes to them, will this cause a problem for you? If so I could do a commit with your new libraries referenced in the project. 

**ZigBeeCoordinatorHandler** and **ZigBeeCoordinatorHandler** conflicts are weird, at first glance I see no changes, and I do not remember messing with them at all. It seems like the whole code in these 2 classes has been offset by 1 line.

cdjackson:
Thanks for this.  I think this will be very hard to merge - at least without significant work it won't be possible.

Currently I'm updating the binding to use the new library which has a lot of changes to incorporate a number of new features. It might be possible to merge your changes before the updates are complete, but it might also be better to do it after now.

In any case, this PR can't be merged until the following are resolved -:
* ensure that the PR only contains your changes - currently many files show that every line has changed. This is probably as you haven't run the project formatter to clean it up and it should only highlight changes in lines you have changed.
* ensure that absolute paths to libraries are removed
* remove any conflicts
* commits need to be signed
* PR needs to be signed

cdjackson:
> ZigBeeCoordinatorHandler and ZigBeeCoordinatorHandler conflicts are weird, at first glance I see no changes, and I do not remember messing with them at all. It seems like the whole code in these 2 classes has been offset by 1 line.

I have a feeling that you might not be using Eclipse? Maybe your IDE has reformatted the files to use tabs instead of spaces (or something like that).

saukijan:
1. Signed-off the PR and last commits (I could try signing-off the older commits, but will that make a big difference?) 
2. Fixed the paths in the classpath (as stated before)
3. Removed source code conflicts

The library binaries are a bit of a pain since they include those small changes we discussed before, and since then you refactored the library (you got rid of ZigBeeDevice class), I am afraid these changes will have to be merged afterward, once I see how you plan to implement the new library into the binding.

PR #34, opened 2017-11-24
Change registration of discovery to handler factory

### LINKED ISSUE #31: ZigBeeCoordinatorHandler must not use bundleContext
The `bundleContext` field will be removed from the BaseThingHandler at some point in time and it is not intended to be used by bindings.
The ZigBeeCoordinatorHandler uses it in order to register a discovery service. This must be changed to registering it in the handler factory instead as it is e.g. done in the Hue binding (and many others) as well.

cdjackson:
Moves discover handler registration into handler factory.
To avoid problems with ordering, the discovery handler is created in the bridge handler so that it can be activated from the bridge handler once the zigbee network is instantiated.

Closes #31 


cdjackson:
Thanks Kai,
Yes - it's problematic trying to make this fit :(

The problem I had implementing the same as Hue is that DiscoveryHandler gets activated before the BridgeHandler is initialized, but the DiscoveryHandler needs to register a listener with the bridge which it can't do until after the BridgeHandler.initialize() is called.

One option is to have a method in the DiscoveryHandler get called once the bridge changes state (ONLINE/OFFLINE), or after activation or something similar so that it can register with the bridge...

There are a few other options I looked at, but they are all a bit messy.

What do you think?

kaikreuzer:
>  the DiscoveryHandler needs to register a listener with the bridge which it can't do until after the BridgeHandler.initialize() is called.

Why can't this registration be done before the initialize? It would be similar to injecting services in a DS component before it gets started. All you would need to do is to add it to some collection and make use of it as soon as the handler is initialized.

cdjackson:
> Why can't this registration be done before the initialize? 

Because it needs to initialise the ZigBee library which is done in the initialize method.  I could implement a secondary listener in the BridgeHandler, so that the discover service registers with that and then the bridge handler effectively proxies  the notifications, but that adds another level just to get around this issue. Another option is to store a reference to the discovery service and register it once the initialise method is called....

cdjackson:
> All you would need to do is to add it to some collection and make use of it as soon as the handler is initialized.

Sorry - misread this part... This is effectively my second option above and yes, I could do this.  It just feels a bit wrong, but if that's the recommended approach then I'll go with that.

kaikreuzer:
Sounds like a good option to me, which nicely decouples the services, while not adding much coding overhead - so clearly my favorite, yes!

cdjackson:
Ok, I’ll look at that then. It just feels quite convoluted, but at least it outwardly fits the model...

cdjackson:
Updated...

src/main/java/org/openhab/binding/zigbee/handler/ZigBeeCoordinatorHandler.java
```diff
@@ -243,10 +241,6 @@ public void initialize() {
 
     @Override
     public void dispose() {
```
kaikreuzer: Wouldn't it be cleaner to also unregister the listeners from the networkManager here?

src/main/java/org/openhab/binding/zigbee/handler/ZigBeeCoordinatorHandler.java
```diff
@@ -87,6 +85,8 @@
 
     protected ZigBeeKey networkKey;
 
+    private List<ZigBeeNetworkNodeListener> listeners = new CopyOnWriteArrayList<ZigBeeNetworkNodeListener>();
```
kaikreuzer: You might rather want to use a Set instead of a List in order to avoid duplicates.

cdjackson:
Yep - updated as suggested.

src/main/java/org/openhab/binding/zigbee/handler/ZigBeeCoordinatorHandler.java
```diff
@@ -243,15 +240,17 @@ public void initialize() {
...
-        // Remove the discovery service
-        discoveryService.deactivate();
-        discoveryRegistration.unregister();
-
         // If we have scheduled tasks, stop them
         if (restartJob != null) {
             restartJob.cancel(true);
         }
 
+        synchronized (listeners) {
```
kaikreuzer: move this behind line 255, so that you can be sure that networkManager!=null

cdjackson:
Done... (hopefully ;) ).

PR #41, opened 2017-12-01
Add color_mode channel and minor code fixes to all converters

### LINKED ISSUE #39: Add Channel for color state
Hue bulbs can operate in two color modes: Hue / Saturation / Brightness or Color Temperature / Brightness. To coordinate with other lights which onyl show white, it would be helpful to have a channel which reflects the operating mode, since this is not reflected in the available channels. The original Hue hub API seems to have a state.colormode [hs|ct] paramter for this, so it is probably transmitted somewhere.
I suggest to call it colormode with possible values ct or hs, for maximum compatibility with people who might have gotten this value from the hue hub.

cdjackson:
Closes #39

### DIFF
--- a/ESH-INF/thing/_channels.xml
+++ b/ESH-INF/thing/_channels.xml
@@ -3,15 +3,31 @@
 	xmlns:thing="http://eclipse.org/smarthome/schemas/thing-description/v1.0.0"
 	xsi:schemaLocation="http://eclipse.org/smarthome/schemas/thing-description/v1.0.0 http://eclipse.org/smarthome/schemas/thing-description-1.0.0.xsd">
 
-	<!-- Color Channel -->
-	<channel-type id="color_color">
-		<item-type>Color</item-type>
-		<label>Color</label>
-		<description>The color channel allows to control the color of a light.
+    <!-- Color Channel -->
+    <channel-type id="color_color">
+        <item-type>Color</item-type>
+        <label>Color</label>
+        <description>The color channel allows to control the color of a light.
             It is also possible to dim values and switch the light on and off.
         </description>
-		<category>ColorLight</category>
-	</channel-type>
+        <category>ColorLight</category>
+    </channel-type>
+
+    <!-- Color Channel -->
+    <channel-type id="color_mode">
+        <item-type>Number</item-type>
+        <label>Color</label>
+        <description>The color mode channel allows to control the color mode of a light.
+        </description>
+        <category>ColorLight</category>
+        <state readOnly="true">
+            <options>
+                <option value="0">Hue and Saturation</option>
+                <option value="1">X and Y</option>
+                <option value="2">Color Temperature</option>
+            </options>
+        </state>
+    </channel-type>
 
 	<!-- Color Temperature Channel -->
 	<channel-type id="color_temperature">
--- a/src/main/java/org/openhab/binding/zigbee/ZigBeeBindingConstants.java
+++ b/src/main/java/org/openhab/binding/zigbee/ZigBeeBindingConstants.java
@@ -45,6 +45,7 @@ public class ZigBeeBindingConstants {
     public static final String CHANNEL_SWITCH_ONOFF = "switch_onoff";
 
     public static final String CHANNEL_COLOR_COLOR = "color_color";
+    public static final String CHANNEL_COLOR_MODE = "color_mode";
     public static final String CHANNEL_COLOR_TEMPERATURE = "color_temperature";
 
     public static final String CHANNEL_TEMPERATURE_VALUE = "sensor_temperature";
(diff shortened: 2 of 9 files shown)


ESH-INF/thing/_channels.xml
```diff
@@ -3,15 +3,31 @@
...
+    <!-- Color Channel -->
+    <channel-type id="color_mode">
+        <item-type>Number</item-type>
+        <label>Color</label>
+        <description>The color mode channel allows to control the color mode of a light.
+        </description>
+        <category>ColorLight</category>
+        <state readOnly="true">
+            <options>
+                <option value="0">Hue and Saturation</option>
```
kaikreuzer: What happens if you send a command to the color_temperature channel when this mode is selected?

ESH-INF/thing/_channels.xml
```diff
@@ -3,15 +3,31 @@
...
+                <option value="1">X and Y</option>
```
kaikreuzer: How do you send X/Y commands at all?

ESH-INF/thing/_channels.xml
```diff
@@ -3,15 +3,31 @@
...
+                <option value="2">Color Temperature</option>
```
kaikreuzer: What happens if you send a command to the color channel when this mode is active?

ESH-INF/thing/_channels.xml
cdjackson: It is up to the bulb. Some bulbs I have update the temperature, some don't. Temperature and colour aren't linked are they? What does Hue do?

ESH-INF/thing/_channels.xml
cdjackson: At the moment it's not supported, but it will soon be added. The binding will decide based on what the bulb supports which system to use. (some) Tradfri uses XY, and Hue uses HS. The binding will discover what the bulb supports and use the appropriate system.

ESH-INF/thing/_channels.xml
cdjackson: I'm not sure what you mean? What "mode" is active? This channel is an indicator of the mode of the bulb - it can't be controlled (maybe I didn't make it read-only. If not, I'll update this).

The way ZigBee bulbs work is that bulbs will switch this mode indicator based on the mode they are in, which in turn is based on the commands they are sent.

ESH-INF/thing/_channels.xml
kaikreuzer: > but it will soon be added

No, please don't! Checkout the Tradfri binding: Sure, the device wants XY, but this conversion is done by the binding - the channel is still proper HSB and nothing else. The same must be done here, otherwise you completely ignore the ESH architecture.

ESH-INF/thing/_channels.xml
kaikreuzer: And here's again where you need to build the abstraction: If the user sends color command, it is expected to be executed. It is your job in the binding to make this sure. If you need to read and change the color_mode first, so be it. But this must be hidden in the binding and not be exposed to the user.

ESH-INF/thing/_channels.xml
cdjackson: Yes- agreed - the binding does the change. This is 100% for sure. IT will expose a standard HSB - no problem. BUT, the device needs to use XY - it doesn't work otherwise. The binding changes the HSB into XY and sends it to the bulb. The bulb will still report XY in its color mode though - I can't change that (unless we just remove the mode status channel).

Sorry - I clearly miss something, or maybe you don't understand what this mode channel is? It's ONLY a status channel to indicate what the BULB is doing - not ESH. ESH, and the binding, will always expose the standard channel types.

ESH-INF/thing/_channels.xml
cdjackson: Yes - again, as stated elsewhere this is what the binding does. It doesn't stop the bulb reporting what mode it is using though does it?

I think you don't understand this - the binding is presenting all the standard features to the user in the standard way - this mode status is on the bulb interface only, so it's indicating what the bulb is doing only.

ESH-INF/thing/_channels.xml
kaikreuzer: The binding is the only one who needs to know what the bulb is doing - therefore you can remove the channel.
As I said, it is perfectly fine for the binding itself to internally read and manage that information, so that it knows, how to communicate with the bulb. But this must not be exposed to anyone else through an ESH channel.

cdjackson:
@kaikreuzer too many emails now and discussions are crossing. I'll summarise here what happens.

The binding receives a color command - ALWAYS a standard HSB Color item type - ALWAYS. The binding converts this into the format the bulb wants - HSB, XY, CT, and sends it to the bulb. The bulb then reports back what mode it is in (again, HSB, XY, CT). It's not possible to set this mode manually - it's a status only. The user asked to have this channel - apparently other bindings have it, but if it's not desirable, I can remove it.

The user will NEVER be presented with an XY channel. The binding will discover what commands the bulb supports (either HSB or XY) and it will convert the incoming Color item into the required format - the user wont worry about this at all.

Does that make sense?

ESH-INF/thing/_channels.xml
kaikreuzer: The bulbs isn't "doing" anything - it has a (visible) state, which is some color, which should be reflected as the state of the color channel - that's all the user needs to know.

ESH-INF/thing/_channels.xml
cdjackson: Ok, I added the channel at user request (@AfromD) - maybe he can explain why it's useful?

PR #65, opened 2017-12-27
Added support for lights supporting only XY color

### LINKED ISSUE #64: Color lights supporting only XY command set do not work (i.e. IKEA Tradfri)
The current color channel converter implementation only supports Hue/Saturation commands and attributes, so some lights do not work.

I have submitted a pull request with the required changes, referencing to this issue.

puzzle-star:
The current color channel converter implementation only supports Hue/Saturation commands and attributes, so some lights do not work.

This pull request addresses this (closes #64)

Signed-off-by: Pedro Garcia <pg@puzzle-star.com> (github: puzzle-star)

### DIFF
--- a/src/main/java/org/openhab/binding/zigbee/converter/ZigBeeConverterColorColor.java
+++ b/src/main/java/org/openhab/binding/zigbee/converter/ZigBeeConverterColorColor.java
@@ -45,6 +45,8 @@ public class ZigBeeConverterColorColor extends ZigBeeBaseChannelConverter implem
     private ZclLevelControlCluster clusterLevelControl;
     private ZclOnOffCluster clusterOnOff;
 
+    private boolean delayedColorChange = false; // Wait for brightness transition before changing color
+
     private boolean initialised = false;
     private PercentType lastBrightness = PercentType.HUNDRED;
     private ScheduledExecutorService colorUpdateScheduler;
@@ -99,6 +101,7 @@ public class ZigBeeConverterColorColor extends ZigBeeBaseChannelConverter implem
             } else if (clusterColorControl.getSupportedAttributes().contains(ZclColorControlCluster.ATTR_CURRENTX)) {
                 logger.debug("{}: Device supports XY color set of commands", endpoint.getIeeeAddress());
                 supportsHue = false;
+                delayedColorChange = true; // By now, only for XY lights till this is configurable
             } else {
                 logger.warn("{}: Device does not support RGB color", endpoint.getIeeeAddress());
                 return;
@@ -279,21 +282,33 @@ public class ZigBeeConverterColorColor extends ZigBeeBaseChannelConverter implem
 
                 try {
                     if (command instanceof HSBType) {
+                        HSBType current = currentHSB;
                         HSBType color = (HSBType) command;
                         PercentType brightness = color.getBrightness();
 
-                        if (brightness.intValue() != currentHSB.getBrightness().intValue()) {
-                            changeBrightness(brightness);
-                            // Wait for transition to complete
-                            // Some lights do not like receiving a level/color change command
-                            // while the previous transition is in progress...
-                            // Thread.sleep(1000);
+                        boolean changeColor = true;
+                        if (delayedColorChange) {
+                            // Color conversion (HUE -> XY -> HUE) makes this necessary due to rounding & precision
+                            int changeSensitivity = supportsHue ? 0 : 1;
+                            changeColor = Math
+                                    .abs(current.getHue().intValue() - color.getHue().intValue()) > changeSensitivity
+                                    || Math.abs(current.getSaturation().intValue()
+                                            - color.getSaturation().intValue()) > changeSensitivity;
                         }
 
-                        if (supportsHue) {
-                            changeColorHueSaturation(color);
-                        } else {
-                            changeColorXY(color);
+                        if (brightness.intValue() != currentHSB.getBrightness().intValue()) {
+                            changeBrightness(brightness);
+                            if (changeColor && delayedColorChange) {
+                                Thread.sleep(1100);
+                            }
+                        }
+
+                        if (changeColor) {
+                            if (supportsHue) {
+                                changeColorHueSaturation(color);
+                            } else {
+                                changeColorXY(color);
+                            }
                         }
                     } else if (command instanceof PercentType) {
                         changeBrightness((PercentType) command);
@@ -356,7 +371,7 @@ public class ZigBeeConverterColorColor extends ZigBeeBaseChannelConverter implem
     }
 
     private void updateColorXY(PercentType x, PercentType y) {
-        HSBType color = ColorHelper.fromXY(x.floatValue() / 100.0f, y.floatValue() / 100.0f, 1.0f);
+        HSBType color = ColorHelper.fromXY(x.floatValue() / 100.0f, y.floatValue() / 100.0f);
         logger.debug("{}: Update Color XY ({}, {}) -> HSV ({}, {}, {})", endpoint.getIeeeAddress(), x.toString(),
                 y.toString(), color.getHue(), color.getSaturation(), currentHSB.getBrightness());
         updateColorHSB(color.getHue(), color.getSaturation());
--- a/src/main/java/org/openhab/binding/zigbee/converter/ZigBeeConverterColorColor.java
+++ b/src/main/java/org/openhab/binding/zigbee/converter/ZigBeeConverterColorColor.java
@@ -25,11 +25,12 @@ import com.zsmartsystems.zigbee.zcl.ZclAttribute;
 import com.zsmartsystems.zigbee.zcl.ZclAttributeListener;
 import com.zsmartsystems.zigbee.zcl.clusters.ZclColorControlCluster;
 import com.zsmartsystems.zigbee.zcl.clusters.ZclLevelControlCluster;
-import com.zsmartsystems.zigbee.zcl.protocol.ZclClusterType;
+import com.zsmartsystems.zigbee.zcl.clusters.ZclOnOffCluster;
 
 /**
  *
  * @author Chris Jackson - Initial Contribution
+ * @author Pedro Garcia - Added CIE XY color support
  *
  */
 public class ZigBeeConverterColorColor extends ZigBeeBaseChannelConverter implements ZclAttributeListener {
@@ -38,8 +39,23 @@ public class ZigBeeConverterColorColor extends ZigBeeBaseChannelConverter implem
     private HSBType currentHSB;
     private ZclColorControlCluster clusterColorControl;
     private ZclLevelControlCluster clusterLevelControl;
+    private ZclOnOffCluster clusterOnOff;
 
     private boolean initialised = false;
+    private PercentType lastBrightness = PercentType.HUNDRED;
+    private Object colorUpdateSync = new Object();
+
+    private boolean supportsHue = false;
+    private float lastHue = -1.0f;
+    private float lastSaturation = -1.0f;
+    private boolean hueChanged = false;
+    private boolean saturationChanged = false;
+
+    private boolean supportsXY = false;
+    private float lastX = -1.0f;
+    private float lastY = -1.0f;
+    private boolean xChanged = false;
+    private boolean yChanged = false;
 
     @Override
     public void initializeConverter() {
@@ -51,34 +67,87 @@ public class ZigBeeConverterColorColor extends ZigBeeBaseChannelConverter implem
 
         clusterColorControl = (ZclColorControlCluster) endpoint.getInputCluster(ZclColorControlCluster.CLUSTER_ID);
         if (clusterColorControl == null) {
-            logger.error("Error opening device control controls {}", endpoint.getIeeeAddress());
+            logger.error("Error opening device color controls {}", endpoint.getIeeeAddress());
             return;
         }
 
         clusterLevelControl = (ZclLevelControlCluster) endpoint.getInputCluster(ZclLevelControlCluster.CLUSTER_ID);
         if (clusterLevelControl == null) {
-            logger.error("Error opening device level controls {}", endpoint.getIeeeAddress());
-            return;
+            logger.warn("Device does not support level control {}", endpoint.getIeeeAddress());
         }
 
-        clusterColorControl.bind();
-        clusterLevelControl.bind();
+        clusterOnOff = (ZclOnOffCluster) endpoint.getInputCluster(ZclOnOffCluster.CLUSTER_ID);
+        if (clusterOnOff == null) {
+            logger.warn("Device does not support on/off control {}", endpoint.getIeeeAddress());
+        }
 
-        // Add a listener, then request the status
-        clusterColorControl.addAttributeListener(this);
-        clusterLevelControl.addAttributeListener(this);
-
-        clusterColorControl.getCurrentHue(0);
-        clusterColorControl.getCurrentSaturation(0);
-        clusterLevelControl.getCurrentLevel(0);
-
-        // Configure reporting - no faster than once per second - no slower than 10 minutes.
+        // Discover whether the device supports HUE/SAT or XY color set of commands
         try {
-            clusterColorControl.setCurrentHueReporting(1, 600, 1).get();
-            clusterColorControl.setCurrentSaturationReporting(1, 600, 1).get();
-            clusterLevelControl.setCurrentLevelReporting(1, 600, 1).get();
-        } catch (ExecutionException | InterruptedException e) {
-            logger.debug("Exception configuring color reporting", e);
+            if (!clusterColorControl.discoverAttributes(false).get()) {
+                logger.warn("{}: Cannot determine whether device supports RGB color. Assuming it supports HUE/SAT",
+                        endpoint.getIeeeAddress());
+                supportsHue = true;
+            } else if (clusterColorControl.getSupportedAttributes().contains(ZclColorControlCluster.ATTR_CURRENTHUE)) {
+                logger.debug("Device supports Hue/Saturation color set of commands {}", endpoint.getIeeeAddress());
+                supportsHue = true;
+            } else if (clusterColorControl.getSupportedAttributes().contains(ZclColorControlCluster.ATTR_CURRENTX)) {
+                logger.debug("Device supports XY color set of commands {}", endpoint.getIeeeAddress());
+                supportsXY = true;
+            } else {
+                logger.warn("Device does not support RGB color {}", endpoint.getIeeeAddress());
+                return;
+            }
+        } catch (InterruptedException | ExecutionException e) {
+            logger.warn(
+                    "{}: Exception checking whether device endpoint supports RGB color. Assuming it supports HUE/SAT",
+                    endpoint.getIeeeAddress(), e);
+            supportsHue = true;
+        }
+
+        // Bind to attribute reports, add listeners, then request the status
+        // Configure reporting - no faster than once per second - no slower than 10 minutes.
+        clusterColorControl.bind();
+        clusterColorControl.addAttributeListener(this);
+        if (supportsHue) {
+            clusterColorControl.getCurrentHue(0);
+            clusterColorControl.getCurrentSaturation(0);
+            try {
+                clusterColorControl.setCurrentHueReporting(1, 600, 1).get();
+                clusterColorControl.setCurrentSaturationReporting(1, 600, 1).get();
+            } catch (ExecutionException | InterruptedException e) {
+                logger.debug("Exception configuring color reporting", e);
+            }
+        } else if (supportsXY) {
+            clusterColorControl.getCurrentX(0);
+            clusterColorControl.getCurrentY(0);
+            try {
+                clusterColorControl.setCurrentXReporting(1, 600, 1).get();
+                clusterColorControl.setCurrentYReporting(1, 600, 1).get();
+            } catch (ExecutionException | InterruptedException e) {
+                logger.debug("Exception configuring color reporting", e);
+            }
+        }
+
+        if (clusterLevelControl != null) {
+            clusterLevelControl.bind();
+            clusterLevelControl.addAttributeListener(this);
+            clusterLevelControl.getCurrentLevel(0);
+            try {
+                clusterLevelControl.setCurrentLevelReporting(1, 600, 1).get();
+            } catch (ExecutionException | InterruptedException e) {
+                logger.debug("Exception configuring level reporting", e);
+            }
+        }
+
+        if (clusterOnOff != null) {
+            clusterOnOff.bind();
+            clusterOnOff.addAttributeListener(this);
+            clusterOnOff.getOnOff(0);
+            try {
+                clusterOnOff.setOnOffReporting(1, 600).get();
+            } catch (ExecutionException | InterruptedException e) {
+                logger.debug("Exception configuring on/off reporting", e);
+            }
         }
 
         initialised = true;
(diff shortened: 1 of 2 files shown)


src/main/java/org/openhab/binding/zigbee/converter/ZigBeeConverterColorColor.java
```diff
@@ -106,72 +266,186 @@ public void run() {
...
+
+                        if (brightness.intValue() != currentHSB.getBrightness().intValue()) {
+                            changeBrightness(brightness);
+                            // Wait for transition to complete
+                            // Some lights do not like receiving a level/color change command
+                            // while the previous transition is in progress...
+                            Thread.sleep(1000);
+                        }
+
+                        if (supportsHue) {
```
cdjackson: As mentioned earlier, I don't know if there's a need for both "supports" flags and we have an ambiguous case here - what if both are false? I think one flag is better - or an enum to define the mode?

src/main/java/org/openhab/binding/zigbee/converter/ZigBeeConverterColorColor.java
```diff
@@ -106,72 +266,186 @@ public void run() {
...
+                            changeColorHueSaturation(color);
+                        } else if (supportsXY) {
+                            changeColorXY(color);
+                        }
                     } else if (command instanceof PercentType) {
-                        level = ((PercentType) command).intValue();
+                        PercentType brightness = (PercentType) command;
+                        changeBrightness(brightness);
```
cdjackson: This could just be -:
```
changeBrightness((PercentType )brightness);
```
and avoid the extra variable.

src/main/java/org/openhab/binding/zigbee/converter/ZigBeeConverterColorColor.java
```diff
@@ -106,72 +266,186 @@ public void run() {
...
                     } else if (command instanceof OnOffType) {
-                        level = (OnOffType) command == OnOffType.ON ? 100 : 0;
-                    } else {
-                        return;
+                        OnOffType onoff = (OnOffType) command;
+                        changeOnOff(onoff);
```
cdjackson: This could just be 
```
changeOnOff((OnOffType)onoff);
```
to avoid the extra variable.

src/main/java/org/openhab/binding/zigbee/converter/ZigBeeConverterColorColor.java
```diff
@@ -106,72 +266,186 @@ public void run() {
...
                 try {
                     if (command instanceof HSBType) {
-                        HSBType hsb = (HSBType) command;
-                        clusterColorControl
-                                .moveToHueAndSaturationCommand((int) (hsb.getHue().doubleValue() * 254.0 / 360.0 + 0.5),
-                                        (int) (hsb.getSaturation().doubleValue() * 254.0 / 100.0 + 0.5), 10)
-                                .get();
-                        level = hsb.getBrightness().intValue();
+                        HSBType color = (HSBType) command;
+                        PercentType brightness = color.getBrightness();
```
cdjackson: These two lines are only used within the next if statement so they should be moved into the conditional block. You also have created two variables that aren't used - maybe you can just use the following below -:
```
changeBrightness(((HSBType)command).getBrightness());
```

src/main/java/org/openhab/binding/zigbee/converter/ZigBeeConverterColorColor.java
```diff
@@ -106,72 +266,186 @@ public void run() {
...
+
+                        if (brightness.intValue() != currentHSB.getBrightness().intValue()) {
+                            changeBrightness(brightness);
+                            // Wait for transition to complete
+                            // Some lights do not like receiving a level/color change command
+                            // while the previous transition is in progress...
+                            Thread.sleep(1000);
```
cdjackson: I'm not sure that this will work properly under all conditions? This will sleep this thread, but I don't think there's anything in the binding will stop another command coming in in another thread and sending another brightness command.

I think a better implementation would be to record the last brightness time, and then check that before sending an update. I think in the near future we will want to make this configurable so that lamps that do allow it, can support it (don't worry about that right now though as I'm working on configuration at the moment).

src/main/java/org/openhab/binding/zigbee/converter/ZigBeeConverterColorColor.java
```diff
@@ -87,14 +156,105 @@ public void initializeConverter() {
...
+
+    private void changeColorXY(HSBType color) throws InterruptedException, ExecutionException {
+        PercentType xy[] = ColorHelper.toXY(color);
+
+        HSBType oldHSB = currentHSB;
+        currentHSB = new HSBType(color.getHue(), color.getSaturation(), oldHSB.getBrightness());
+
+        logger.debug("{}: Change Color HSV. {}, {}, {}", endpoint.getIeeeAddress(), color.getHue(),
+                color.getSaturation(), oldHSB.getBrightness());
+        logger.debug("{}: Calculated XY. {}, {}", endpoint.getIeeeAddress(), xy[0], xy[1]);
```
cdjackson: I'd suggest to make these a single log entry as really what you're wanting to see is the conversion and in the changeColorHueSaturation method, you aren't logging the "Change color HSV" line....

src/main/java/org/openhab/binding/zigbee/converter/ZigBeeConverterColorColor.java
```diff
@@ -106,72 +266,186 @@ public void run() {
...
     }
 
+    private void updateOnOff(OnOffType onoff) {
+        updateBrightness(lastBrightness);
+    }
+
+    private void updateBrightness(PercentType brightness) {
+        HSBType oldHSB = currentHSB;
+        HSBType newHSB = new HSBType(oldHSB.getHue(), oldHSB.getSaturation(), brightness);
+        currentHSB = newHSB;
```
cdjackson: I don't think we need the extra step here - I think you can just use this? -:
```
currentHSB - new HSBType(...).
```

src/main/java/org/openhab/binding/zigbee/converter/ZigBeeConverterColorColor.java
```diff
@@ -106,72 +266,186 @@ public void run() {
...
+        lastBrightness = brightness;
+        updateChannelState(newHSB);
+    }
+
+    private void updateColorHSB(DecimalType hue, PercentType saturation) {
+        HSBType oldHSB = currentHSB;
+        HSBType newHSB = new HSBType(hue, saturation, oldHSB.getBrightness());
```
cdjackson: I don't think we need the extra step here - I think you can just use this? -:
```
currentHSB - new HSBType(...).
```

src/main/java/org/openhab/binding/zigbee/converter/ZigBeeConverterColorColor.java
```diff
@@ -38,8 +39,23 @@
     private HSBType currentHSB;
     private ZclColorControlCluster clusterColorControl;
     private ZclLevelControlCluster clusterLevelControl;
+    private ZclOnOffCluster clusterOnOff;
 
     private boolean initialised = false;
+    private PercentType lastBrightness = PercentType.HUNDRED;
+    private Object colorUpdateSync = new Object();
+
+    private boolean supportsHue = false;
```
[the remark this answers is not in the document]
puzzle-star: I added it like this because at some moment it maight be worth adding CT to this channel, instead of keeping it in a different one (in the end everything is in the color control cluster), but I will change it to its original implementation (I only had put supportsHue at the beginning).

src/main/java/org/openhab/binding/zigbee/converter/ZigBeeConverterColorColor.java
puzzle-star: So do we leave it as is by now, or do you prefer to synchronize this part of the code ?

src/main/java/org/openhab/binding/zigbee/converter/ZigBeeConverterColorColor.java
puzzle-star: If we remove oldHSB we will need synchronization in all the code that might change currentHSB, as currentHSB might change between getHue() and getSaturation(), leading to unconsistent / unpredicatable resulting color.

As newHSB is used also later, if we use currentHSB intead woukld also lead to inconsistencies between the received parameter (ie brightnes here), and currentHSB if modified, requiring again synchronization in all the code modifying current HSB.

I would prefer to keep the extra step and both temp variables than synchronizing all currentHSB changes just for this. What do you think ?

src/main/java/org/openhab/binding/zigbee/converter/ZigBeeConverterColorColor.java
puzzle-star: Same as the previous one...

src/main/java/org/openhab/binding/zigbee/converter/ZigBeeConverterColorColor.java
```diff
@@ -106,72 +266,186 @@ public void run() {
...
-                    updateChannelState(currentHSB);
+
+        synchronized (colorUpdateSync) {
+            try {
+                if (attribute.getCluster().getId() == ZclOnOffCluster.CLUSTER_ID
+                        && attribute.getId() == ZclOnOffCluster.ATTR_ONOFF) {
+                    Boolean value = (Boolean) attribute.getLastValue();
+                    OnOffType onoff = value ? OnOffType.ON : OnOffType.OFF;
+                    updateOnOff(onoff);
+                } else if (attribute.getCluster().getId() == ZclLevelControlCluster.CLUSTER_ID
```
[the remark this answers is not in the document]
puzzle-star: Will do. I do not normally like nested ifs for code readability if the repeated condition is simple to compute, but I understand it will be more efficient anyway.

src/main/java/org/openhab/binding/zigbee/converter/ZigBeeConverterColorColor.java
```diff
@@ -106,72 +266,186 @@ public void run() {
...
+                    updateColorHSB();
+                } else if (xChanged && yChanged) {
+                    updateColorXY();
+                } else if (hueChanged || saturationChanged || xChanged || yChanged) {
+                    // Wait some time and update anyway if only one attribute in each pair is updated
+                    new Thread() {
+                        @Override
+                        public void run() {
+                            try {
+                                sleep(500);
```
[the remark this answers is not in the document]
puzzle-star: Not really.

Within the relevant synchronized portions of code, the [hue, saturation, x, y]Changed set of variables are set to false when the update happens, so after the second attribute is received, the delayed thread will see them as false and just skip the update. Normally it will be the second attribute in each pair forcing the update inmediately, unless the attribute is reported 500ms later (this is just a safeguard in case only one is updated). 500ms seems a good compromise, as both attributes should be reported within this timeframe (in my IKEA lights even both are reported in the same message).

src/main/java/org/openhab/binding/zigbee/converter/ZigBeeConverterColorColor.java
```diff
@@ -51,34 +67,87 @@ public void initializeConverter() {
...
-            logger.error("Error opening device level controls {}", endpoint.getIeeeAddress());
-            return;
+            logger.warn("Device does not support level control {}", endpoint.getIeeeAddress());
         }
 
-        clusterColorControl.bind();
-        clusterLevelControl.bind();
+        clusterOnOff = (ZclOnOffCluster) endpoint.getInputCluster(ZclOnOffCluster.CLUSTER_ID);
+        if (clusterOnOff == null) {
+            logger.warn("Device does not support on/off control {}", endpoint.getIeeeAddress());
```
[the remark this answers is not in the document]
puzzle-star: Same as above, in terms of war vs debug

src/main/java/org/openhab/binding/zigbee/converter/ZigBeeConverterColorColor.java
```diff
@@ -51,34 +67,87 @@ public void initializeConverter() {
...
+                        endpoint.getIeeeAddress());
+                supportsHue = true;
+            } else if (clusterColorControl.getSupportedAttributes().contains(ZclColorControlCluster.ATTR_CURRENTHUE)) {
+                logger.debug("Device supports Hue/Saturation color set of commands {}", endpoint.getIeeeAddress());
+                supportsHue = true;
+            } else if (clusterColorControl.getSupportedAttributes().contains(ZclColorControlCluster.ATTR_CURRENTX)) {
+                logger.debug("Device supports XY color set of commands {}", endpoint.getIeeeAddress());
+                supportsXY = true;
+            } else {
+                logger.warn("Device does not support RGB color {}", endpoint.getIeeeAddress());
```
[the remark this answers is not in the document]
puzzle-star: I am warning here because we arrived at this point of code assuming the device supports color, and it doesn't. So we created the channel and it will not work. Should'nt we warn instead of debug here ?

src/main/java/org/openhab/binding/zigbee/converter/ZigBeeConverterColorColor.java
```diff
@@ -51,34 +67,87 @@ public void initializeConverter() {
...
-            logger.error("Error opening device control controls {}", endpoint.getIeeeAddress());
+            logger.error("Error opening device color controls {}", endpoint.getIeeeAddress());
             return;
         }
 
         clusterLevelControl = (ZclLevelControlCluster) endpoint.getInputCluster(ZclLevelControlCluster.CLUSTER_ID);
         if (clusterLevelControl == null) {
-            logger.error("Error opening device level controls {}", endpoint.getIeeeAddress());
-            return;
+            logger.warn("Device does not support level control {}", endpoint.getIeeeAddress());
```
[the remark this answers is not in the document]
puzzle-star: I put it as warn as the OH impelemtation assumes a Color channel supports level and switching, while the channel does not really.
Wouldn't it make sense to warn about it when this happens ?

src/main/java/org/openhab/binding/zigbee/converter/ZigBeeConverterColorColor.java
```diff
@@ -51,34 +67,87 @@ public void initializeConverter() {
 
         clusterColorControl = (ZclColorControlCluster) endpoint.getInputCluster(ZclColorControlCluster.CLUSTER_ID);
         if (clusterColorControl == null) {
-            logger.error("Error opening device control controls {}", endpoint.getIeeeAddress());
+            logger.error("Error opening device color controls {}", endpoint.getIeeeAddress());
```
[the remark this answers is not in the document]
puzzle-star: Will change it to standard logging (this was part of the original implemetation anyway).

BTW, i think this should be an error. When we arrive at this point of code, the channel is created because it detected color is supported, and here it happens we cannot get the cluster. This should never happen, and if it does the channel will not work. Error or Warn ? (I do not think it should be debug here)

src/main/java/org/openhab/binding/zigbee/converter/ZigBeeConverterColorColor.java
cdjackson: Ok, I didn't consider the thread synch issues here so I guess it's ok to leave this. Maybe add a comment to remind anyone reviewing later.

src/main/java/org/openhab/binding/zigbee/converter/ZigBeeConverterColorColor.java
cdjackson: Ok

src/main/java/org/openhab/binding/zigbee/converter/ZigBeeConverterColorColor.java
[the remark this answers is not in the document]
cdjackson: Ok, if CT gets added here then we can update, but at the moment CT is still a separate channel (I think) in other bindings, and currently in the ZigBee binding a converter can't handle more than 1 channel so it can't handle CT and Color in one converter.

src/main/java/org/openhab/binding/zigbee/converter/ZigBeeConverterColorColor.java
puzzle-star: 'brightness' is used both in the if statement and in the conditional block. 'color' is used in the next if statement (even if only a typecasting, it is used twice, plus the assignment in 'brightness').

Wouldn't it make sense to keep them then ?

src/main/java/org/openhab/binding/zigbee/converter/ZigBeeConverterColorColor.java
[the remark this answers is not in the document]
cdjackson: Yes - I know some logging in the current code doesn't confirm to the "standard" - I've changed most of it in my next PR, but would like to standardise this as much as possible as it really makes things easier for debugging (lesson learnt from the ZWave binding).

Agreed - error is fine here.

src/main/java/org/openhab/binding/zigbee/converter/ZigBeeConverterColorColor.java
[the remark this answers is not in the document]
cdjackson: Ok, let's leave this one as warn as it's kind of needed in this converter for HSB.

src/main/java/org/openhab/binding/zigbee/converter/ZigBeeConverterColorColor.java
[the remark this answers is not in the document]
cdjackson: Is On/Off NEEDED? In the past it wasn't as we used the MoveWithOnOff method in the level control. This still looks to be the case, so On/Off will still work, but using the level control (right?). If so, I would say this is debug as no need to confuse the user when ON/OFF will work just fine.

src/main/java/org/openhab/binding/zigbee/converter/ZigBeeConverterColorColor.java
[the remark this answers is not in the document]
cdjackson: Ok.

src/main/java/org/openhab/binding/zigbee/converter/ZigBeeConverterColorColor.java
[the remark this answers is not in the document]
puzzle-star: Agreed, as the implemetation will default to level control if supported and if it is not then it will already warn about it (as per above). So moving this to debug

src/main/java/org/openhab/binding/zigbee/converter/ZigBeeConverterColorColor.java
cdjackson: I actually misunderstood the purpose of this...

Do you have bulbs that don't like color setting when brightness is being changed? What happens if you send these close together? What happens if you just send two brightness commands close together - is that ok or does it also cause problems?

Does the same issue occur if brightness is moved after color?

I don't really like the sleep here as I'm sure it must look a bit strange on some bulbs and if someone is moving a color wheel, then for sure we will violate this 1 second anyway and the problem will happen unless we delay everything which is (IMHO) not a nice option as we effectively slow down changes from a UI is someone is "spinning" a color wheel.

puzzle-star: My IKEA lights abort the brightness transition at whichever point it is when a color change is requested. The same if the order is reverted, with the color. Will experiment a bit more to see if I can workaround this.

Anyway, this delay should only happen when both color and brightness are changed within the same command (HSB Type):

    if (brightness.intValue() != currentHSB.getBrightness().intValue()) {

So when changing through a color wheel this branch is skipped as we only change color, not brightness, and we compare it to current value and skip the change (and delay) if brightness does not change

The 1 second delay will only happen then if we receive a command which changes color and brightness at the same time (which I do not think is so common, in terms of UI: might happen in rules), so it should have no impact on the normal behaviour of changes through UI (color wheel, etc)

I do not like the sleep either, and it could be made configurable when config is available, only for lights requiring it.

src/main/java/org/openhab/binding/zigbee/converter/ZigBeeConverterColorColor.java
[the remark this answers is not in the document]
cdjackson: I suspect this isn't a very good way to acheive what you want. You will potentially create a lot of threads if the attributes update often (which might happen when someone is playing around with settings). I would suggest to use a ScheduledExecutorService - create this in your initialisation method, then when you want to create a new timer here, you can do so and can cancel anything that is already waiting to run.

```
        ScheduledExecutorService scheduler = Executors.newSingleThreadScheduledExecutor();

        ScheduledFuture<?> timer = scheduler.schedule(task, 500, TimeUnit.MILLISECONDS);
        timer.cancel(true);
```

What do you think?

src/main/java/org/openhab/binding/zigbee/converter/ZigBeeConverterColorColor.java
cdjackson: What happens if you use a faster transition time? eg set transition time to 0 in this case and avoid the sleep? Would that work?

src/main/java/org/openhab/binding/zigbee/converter/ZigBeeConverterColorColor.java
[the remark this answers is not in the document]
puzzle-star: Done & testing

src/main/java/org/openhab/binding/zigbee/converter/ZigBeeConverterColorColor.java
puzzle-star: I does not work either. It looks like the IKEA lights do not like receiveing both commands just one  immediately after the other. A small delay (tried with 100ms) is still necessary. When I was looking for DH examples for SmartThings, I say a similar issue (and workaround) on other device handlers, so I think this kind of bug may not be exclusive of IKEA lights, anyway...

Immediate transition time is, anyway, somehow "awful to see", used to having the lights change smoothly.

So what I have done:
- Kept the transition time
- Added the delay only if there is a real change in both brightness and color (so no effect on user experience when using the UIs)
- Added a boolean setting (by now set to true) so that this behaviour can be activated on demand once there is a way to provide conguration

Just testing, and will now send the PR with all the changes

puzzle-star: Or, considering XY support "experimental", I am setting the "delayedColorChange" boolean to true only for XY lights, so that we no not change the behaviour for working Hue/Sat lights. Would this make sense for you ?

cdjackson: I don't mind either way. Is this issue only an XY issue, or also an HS issue? I assumed both?

At the end of the day, if there are "problems" after we merge this, then we'll sort them out. If it's not 100% perfect, then we'll learn from it and find the best way even if it takes a couple of tries and some wider user feedback...

puzzle-star: A transition time of 0 still requires some 100ms delay between the two commands, or the first one is acked, but silently aborted when receiving the second one. When there is a transition, I see the light start the level change, and then revert to the original brightness when it receives the color command. Nice bug...

I do not think the issue is only related to XY, but as there is surely people already using the channel for hue/sat lights, and nobody complained about it, to avoid impact on users for who it is working OK now, it may be better to apply this to the only XY light we have to test (for which it fails without the delay), and later on let this be configured if needed.

cdjackson:
Ok - fine with me.

src/main/java/org/openhab/binding/zigbee/converter/ZigBeeConverterColorColor.java
```diff
@@ -106,72 +280,206 @@ public void run() {
...
+                        colorUpdateTimer = null;
+                    }
+
+                    if (hueChanged && saturationChanged) {
+                        updateColorHSB();
+                    } else if (xChanged && yChanged) {
+                        updateColorXY();
+                    } else {
+                        // Wait some time and update anyway if only one attribute in each pair is updated
+                        colorUpdateTimer = colorUpdateScheduler.schedule(new Runnable() {
```
cdjackson: This will still end up with the thread being created (I think). You probably ought to cancel the timer - I think this could be done.

I would think adding the following code just above here at line 451 should do it. This way the timer is cancelled, and will only ever run if we end up in the default statement.

```
if(colorUpdateTimer != null) {
    colorUpdateTimer.cancel();
    colorUpdateTimer = null;
}
```

puzzle-star: Hi Chris, it is already there, exactly between lines 447 and 450 ::)

cdjackson: Hmmm - my bad! Thanks.

cdjackson:
Sorry - one last thing before I can merge - can you [sign your code](https://docs.openhab.org/developers/contributing/contributing#sign-your-work) - at least the last commit if possible, and also the PR.

puzzle-star:
Just sent a new  commit with the signoff (sure there is a better way to do so, but... at least the last commit is signed off now).

For the sign-off of the PR, is it just adding the same "Signed-off-by" to the PR message ? I have just done so, but if it is to be done differently, please let me know

cdjackson:
I think that's fine - thanks.

cdjackson:
Sorry for the delay getting this merged - I spent the last day looking at some ZWave issues but have now given this a quick test with a hue lamp. All good so will merge.

Thanks.
Chris

PR #66, opened 2017-12-29
Additions: Configuration, Polling, Channel Consolidation

cdjackson:
This is a mostly minor refactoring and some additional features -:

* Simplifies converters a little.
* Adds channel configuration. This adds a configuration class that is cluster specific to allow re-use between converters which are channel specific.
* Adds polling. The binding detects if bind/reporting fails and sets the polling period to a higher rate (currently not user configurable).
* Channel consolidation is now performed to remove channels that are not required (eg removing switch if level is supported...).
* Thing definition is only changed if the channels have changed from the definition at startup.
* Battery percent converter is added (untested).
* ZigBee library is bumped to 1.0.3-SNAPSHOT. This fixes some issues with ZDO message addressing (eg bind/unbind).


weakfl:
I've changed the channels for my relay items from switch to level to keep things working when this PR gets merged. However, the level channel is noticeably slower when turning a relay off (on is still immediate).

cdjackson:
So the relay device is supporting the level control as well as on/off? Slightly strange (or does it also allow to be used as a dimmer?), but I’ll look at changing the way this works a little.

weakfl:
Yes, the relays have two channels, but the level channel is just turning the relay on/off.
I guess it must have something to do with code reuse in the BJ firmware. Real level support is probably restricted to the [6715-U inserts](http://www.busch-jaeger-catalogue.com/2CKA006710A0007,artikel.html). I don't know why else the [6711-U inserts](http://www.busch-jaeger-catalogue.com/2CKA006710A0006,artikel.html) would have a level channel.

cdjackson:
Ok - at the moment you have both level and on/off controls right?
What I’ve done is to change this so that it uses the on/off cluster if you send an on/off command. So, in PaperUI you will only see a slider/dimmer, but if you are using sitemaps or HABpanel you should be able to define this channel to use a switch widget, and if you do that it will use the on/off cluster the same as now.

weakfl:
> Ok - at the moment you have both level and on/off controls right?

Exactly.

And that's perfect, thanks. I've configured the items as switches in the sitemap anyway, doesn't make much sense to have a dimmer if it doesn't dimm :)

PR #95, opened 2018-01-15
Improve property discovery performance

cdjackson:
This should improve the startup performance, and should significantly help on large networks, and networks with battery devices. It now doesn't request the device properties if they are already set.

I've also removed some unused properties in the BindingConstants where they have been changed to use standard property names.


cdjackson:
#95

PR #136, opened 2018-02-06
Fix SpotBugs Errors in ZigBeeNetworkStateSerializerImpl

brodo:
Hi again,
this is another [SpotBugs](https://github.com/spotbugs/spotbugs) fix. Also, on line 114, I changed `file.getParent().mkdirs()` to `file.mkdirs()`. We had another file not found exception and that fixed it...

```
RV: Method ignores exceptional return value (RV_RETURN_VALUE_IGNORED_BAD_PRACTICE)
This method returns a value that is not checked. The return value should be checked since it can indicate an unusual or unexpected function execution. For example, the File.delete() method returns false if the file could not be successfully deleted (rather than throwing an Exception). If you don't check the result, you won't notice if the method invocation signals unexpected behavior by returning an atypical return value.
```

Cheers
Julian

Signed-off-by: Julian Dax <julian.dax@itemis.com>


cdjackson:
@brodo I didn't check this properly before merging - sorry. 

With this change we now create a folder which is the name of the XML file which then breaks the serialisation completely.

Additionally, I don't know that I understand why you needed to add the ```mkdirs``` here in the first place since it should be created in the ```openStream``` method.

Can you comment please?

Issue #139, opened 2018-02-06
The binding cannot be compiled against OSGi 4.2 (due to used generic type parameter introduced in 4.3)


hsudbrock:
In OSGi 4.3, a generic type parameter was introduced for the class `org.osgi.framework.ServiceRegistration` (cf. the API docs for 4.2, https://osgi.org/javadoc/r4v42/index.html, and for 4.3, https://osgi.org/javadoc/r4v43/core/index.html).

Since this generic type parameter is used in the binding, the binding cannot be compiled against a target platform that only supports OSGi 4.2.

While seeing the benefit of using the generic type parameter, I propose to remove it to become compatible with OSGi 4.2, the minimum framework version supported by Eclipse SmartHome (cf. https://www.eclipse.org/smarthome/documentation/development/guidelines.html Part C).

PR #140, opened 2018-02-06
Remove generic type parameter from ServiceRegistration, for compatibility with OSGi version 4.2

hsudbrock:
This PR removes all three occurrences of the generic type parameter for the `ServiceRegistration` interface, thereby fixing #139 and making the binding compile-time compatible with OSGi version 4.2 (the minimal OSGi version supported by Eclipse SmartHome).


cdjackson:
Thanks.

PR #142, opened 2018-02-07
Fix SpotBugs Errors: Delete ZigBeeActivator

brodo:
This is a Spotbugs fix which requires some bigger changes.
It relates to the following error:

```
ST: Write to static field from instance method (ST_WRITE_TO_STATIC_FROM_INSTANCE_METHOD)
This instance method writes to a static field. This is tricky to get correct if multiple instances are being manipulated, and generally bad practice.
```

As the ZigBeeActivator saves the bundle context I saw no other way to archive this then to just don't save it at all and replace calls to `ZigBeeActivator.getContext()` calls to `FrameworkUtil.getBundle()`. When this is done I don't longer see a need for the activator in the first place.

Signed-off-by: Julian Dax <julian.dax@itemis.com>


cdjackson:
If #143 didn't cover the changes here then please update.

Thanks.

cdjackson:
As above, closing this as the issue should be covered by #143.

PR #152, opened 2018-02-14
Add Busch-Jaeger 6715 U to list of supported devices

weakfl:
Chris, had some time today to install a Busch-Jaeger 6715 U LED-Dimmer insert. It's working great, so it can be added to the supported devices if you like 😉


cdjackson:
Thanks :)

Issue #47, opened 2017-12-02
Refactor each dongle into separate bundle


cdjackson:
The binding needs to be split into separate packages to make it more manageable/extensible, and reduce installation footprint. There should be a single bundle containing the majority of the code (ZigBee stack, thing handler, converters...) and another bundle with just the dongle driver JAR, the bridge handler and XML.

@kaikreuzer your thoughts would be very welcome with the following...

My initial thinking was to use fragments for the dongle bundles - this should solve getting all the bits in one place, the XML file should be read fine. The HandlerFactory could have a method to directly register a new handler (eg ```registerBridge(BridgeTypeUID, BridgeHandler)```)...  Given the discovery service is instantiated from the factory, we can pass in the same list of bridge UIDs...

I think this ought to work ok and hopefully you won't tell me it violates too many OSGi principals ;)

This is a similar issue to the discussion we had a few months back for BLE and I think you said you were looking at other possible extensions to the framework to support this sort of structure (??) so any further suggestions welcome.

I'm not in a huge hurry to do this, but when you find an elusive "few minutes" I'd welcome your thoughts...

kaikreuzer:
Sounds indeed very much like the same requirement as we had for BLE.
Please do not go for fragments - the very same can be nicely done with proper bundles.

I had done all this refactoring in the bluetooth PR, so please have a look at https://github.com/eclipse/smarthome/pull/4535.

The only issue at the moment is that the "device" thing types currently need a reference to all existing dongle-bridge-ids in their XML (like [here](https://github.com/eclipse/smarthome/pull/4535/files#diff-c04aad8af96b12ec8f1a43f5da679403R8)), while it would be nicer to simply say you require "some zigbee bridge". But this is imho ok for now and once the framework supports this, it will be a very simply change in the bindings.

You might also want to think about the "north side" of the binding as well: While in theory your "main" bundle can support any Zigbee device by reading its clusters and supporting them, I can imagine (from what I have learned about Zigbee in practice) that you might require special "hacks" for certain devices or also implementations for proprietary clusters of some devices. So it probably makes sense to go the same way as the Bluetooth binding here as well: The "base" bundle supports everything straight-forward, but it allows additional bundles to bring specific handlers and thing type definitions for special devices.

Let me know if you have any questions about this approach (that are not answered by the code in the Bluetooth PR)!

cdjackson:
[...]

PR #159, opened 2018-02-26
Split dongles into separate bundles

cdjackson:
This splits each dongle into a separate bundles as per #47. Currently it's tested with the Telegesis dongle only but all dongles have been refactored.
The library jars are included in the PR at this stage so that it can be tested. These just need to be included into the TP.
There's also some tidying needed of the readme files etc.


kaikreuzer:
I am a bit uncertain on the effects of this restructuring. So far we had this repo designed in a way that it includes the bundle in the repo root, so that it nicely blends into https://github.com/openhab/openhab-bundles. Did you check whether this aggregated build still works as expected?
Also, can it easily be imported in the IDE and used from there as before?
Furthermore, is there any impact on the openhab-docs project, which aggregates the documentation from the different repos?

tomhoefer:
My expectation would be to have at least separated bundles for:
- ZSS CL
- ZSS Ember
- ZSS Telegesis
- General OH Binding
- OH Binding specific logic for Ember
- OH Binding specific logic for Telegesis

kaikreuzer:
@cdjackson & @tomhoefer As promised, I have tried to create the project structure that I have suggested last week. You can find the result at https://github.com/openhab/zigbee.

@cdjackson I think it nicely addresses your concerns that people should not be bothered by any ESH specifics if they are not interested in it, but want to use the "raw" library for their own purposes. If you look at the repo, it imho looks pretty clean and straight forward and if you run `mvn clean install` on it, it will produce you the artifacts through plain Maven as expected:

```
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary:
[INFO] 
[INFO] zigbee ............................................. SUCCESS [  1.897 s]
[INFO] zigbee-core ........................................ SUCCESS [  8.018 s]
[INFO] zigbee-cc2531 ...................................... SUCCESS [  1.049 s]
[INFO] zigbee-xbee ........................................ SUCCESS [  0.622 s]
[INFO] zigbee-ember ....................................... SUCCESS [  1.252 s]
[INFO] zigbee-telegesis ................................... SUCCESS [  0.528 s]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 13.787 s
[INFO] Finished at: 2018-03-12T21:14:51+01:00
[INFO] Final Memory: 31M/645M
[INFO] ------------------------------------------------------------------------
```
For the use in ESH, all that one need to do is to run `mvn -f pom-esh.xml clean install` instead and the build will magically produce the full ESH binding bundles through Tycho:
```
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary:
[INFO] 
[INFO] ZigBee Binding Parent POM .......................... SUCCESS [  0.727 s]
[INFO] ZigBee Binding ..................................... SUCCESS [  5.109 s]
[INFO] cc2531 Dongle Support for ZigBee Binding ........... SUCCESS [  1.315 s]
[INFO] Ember Dongle Support for ZigBee Binding ............ SUCCESS [  1.363 s]
[INFO] Telegesis Dongle Support for ZigBee Binding ........ SUCCESS [  0.527 s]
[INFO] XBee Dongle Support for ZigBee Binding ............. SUCCESS [  0.569 s]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 41.058 s
[INFO] Finished at: 2018-03-12T21:15:50+01:00
[INFO] Final Memory: 67M/772M
[INFO] ------------------------------------------------------------------------
```

The projects can be imported in the ESH IDE as PDE projects and they fully compile, so it is very easy to work on all parts of the code without having to build and package any intermediate binaries. If being used as pure Maven projects (e.g. in IntelliJ), the ESH parts will simply be ignored by the IDEs and builds.

I'd think that this is a well working compromise that allows us to minimise the roundtrips during development and it is a good preparation towards moving the code base to the Eclipse Foundation as planned.

cdjackson:
Hi Kai,
Thanks for looking at this. As I said last week, I’m not keen on adding the source of the libraries directly into the binding. I don’t believe it’s necessary, or even a good idea and even like this, I think the framework is hidden and people will not (easily) find a standalone library they can use external to OH/ESH - even if it is possible. 

Additionally, it is much easier (IMHO) to debug the library separately to the binding and with this merge we’ve lost this functionality. You seem to have the impression that there is continual “back and forward” during development, but that’s not the case - all development of library features is fully tested and debugged first without the complexity of the binding and ESH. Once tested and the ZigBee functions are working, the binding is then debugged as needed. Trying to debug low level ZigBee functionality directly in the binding would (IMHO) be a lot more difficult.

I would prefer to keep to what we originally discussed, and that is to keep the two repositories separate - even if they are both under OH repository if that’s necessary. Can I suggest we discuss when you and @tomhoefer are available.

Cheers
Chris

tomhoefer:
Hi all,

I am fine to disucss this topic - I see also a benefit by Kai´s proposal

BR

Thomas

cdjackson:
Thanks Thomas. I'd be interested to hear the advantages you see. Maybe for OH only, there could be some advantages, along with the negatives, but in the broader picture, for me it is highly negative.

Please let me know when you want to discuss.

tomhoefer:
I know that I originally mentioned that a split into the parts listed above is required. However I felt immediately familiar with the structure in the new repo. As a newbie I understand the structure at first sight:
- zigbee-core
- zigbee-cc2531
- zigbee-ember
- zigbee-telegesis
- zigbee-xbee
So core will contain the CL stuff and everything else will contain the impls for the different dongles. Inside the dongles we can make the general framework code more visible. In our builds we can also build the different artefacts. Let us discuss this topic so that everybody likes the final solution

cdjackson:
Fundamentally, this is structure exacly the same as I proposed in this PR (the names are different, but it's the same split). 

My major concern is the inclusion of the source directly into openHAB binding rather than using the external library. This I can't really work with - the library is separate and being used by other systems. We don't normally include library source into a binding directly so I don't expect we should do it here. It makes working with the system a lot more difficult (for me at least) and makes it difficult for people outside of ESH environment to use.

If we really MUST start incorporating the source within the binding, then I will manage the source externally, and periodically synchronise. This really isn't something I want - I much prefer a modular system - isn't that the right way to manage libraries?

tomhoefer:
I checked the new structure and have a few comments:
- It seems that the binding bundles cannot be "compiled" if we havent imported the org.openhab.binding.zigbee.zcl bundle into the workspace
- We should avoid committing binary jars into the repo
- Is it possible to avoid using the m2e nature as it results in errors in the pom xmls?

Since we cannot add the zcl bundles to the target platform (as they are updated too often) we must build the bundles from source. In the context of the openhab build we should contribute the zcl framework bundles to the newly created zigbee repository. Then we could simply add the projects to the workspace so that everything can be resolved and it could be added to openhab-bundles so that the headless maven build is also able to compile everything.

kaikreuzer:
I'd agree that having all the sources in the workspace in order to compile is much better than including (and frequently updating) the binaries in the source repo.
Once the zcl libs are stable so that an update every few months is sufficient, we could put them in the target platform, just as other dependencies like jupnp, etc. But if I understood you right, this is currently not an option.

cdjackson:
> It seems that the binding bundles cannot be "compiled" if we havent imported the org.openhab.binding.zigbee.zcl bundle into the workspace
As with any dependencies, they need to be somewhere. Either in the TP or another JAR. You said that you wanted them in the TP, so I have kept them in a separate bundle to allow the implementation to define the packaging.

> We should avoid committing binary jars into the repo
For now, I wanted to get something committed. It is possible to copy them using maven at compile time, but I thought for simplicity this was easiest for now. Again, if the packaging puts the dependencies into the TP, then this is not needed at all.

> Is it possible to avoid using the m2e nature as it results in errors in the pom xmls?
If there’s another option for managing maven dependencies in Eclipse, then sure. But I’m not sure if that’s possible is it?  The errors can just be ignored with eclipse settings.


> Since we cannot add the zcl bundles to the target platform (as they are updated too often)
> 
What does “too often” mean? I normally update them every couple of weeks, but I don’t really think they should be updating particularly often. As we have discussed, the current implementation for most devices should be stable.


> we must build the bundles from source. In the context of the openhab build we should contribute the zcl framework bundles to the newly created zigbee repository.
> 
If we really MUST add the source, which I strongly disagree with, then I will need to manage the existing library as well as the OH library. If that is what you require then this makes things difficult, but that’s for me to manage I guess.


I know I’ve said this before, but I really can’t understand the need to consume the ZCL project into OH. Dependency management is normal work and we shouldn’t simply consume dependant libraries into openHAB as source (IMHO of course ;) ).

tomhoefer:
> As with any dependencies, they need to be somewhere. Either in the TP or another JAR. You said that you wanted them in the TP, so I have kept them in a separate bundle to allow the implementation to define the packaging.

Or in an own workspace project as shown originally by @kaikreuzer 

> What does “too often” mean?

I understood @kaikreuzer typically changes the TP only once or twice a year.

> I know I’ve said this before, but I really can’t understand the need to consume the ZCL project into OH. 

I thought that we have the same understanding that the core framework project is hopefully becoming a eclipse project in the future and that we should show the corresponding motivation for this. Hence having a project in the org domain is a better indication than the commercial com domain (this might be a reason for the alliance, too).

What shall we do? I would like to find a solution so that everybody is happy and that nobody is blocked for future development. I am also not an expert of the openhab build infrastructure - so we need also the agreement of Kai so that the final solution can easily be built

cdjackson:
> Or in an own workspace project as shown originally by @kaikreuzer

I don't understand this though - the only difference is that Kai has simply placed the source from the cluster library into the binding. To me this isn't the right method to manage dependencies.

> I understood @kaikreuzer typically changes the TP only once or twice a year.

Ok, I wasn't assuming that the ZCL libs would be added to the TP in OH in any case. This is why I added a separate bundle to manage the libs. However I understood that in your system, you would manage the libs directly through the TP and you didn't want the JARs in the bundles (hence the separation of the libs and the bundle).

> I thought that we have the same understanding that the core framework project is hopefully becoming a eclipse project in the future

But that's not what is proposed now. Now we are simply consuming the library source into the OH binding. As I've explained, that makes things pretty difficult to manage. As I explained, I'd be happy to take this forward as an Eclipse project if needed - that said, I'm not really sure that the namespace makes so much difference - I'm sure legally it's the license that matters ;)

> What shall we do? I would like to find a solution so that everybody is happy and that nobody is blocked for future development

I have a lot of time suggesting solutions that are unacceptable. I don't believe that all dependencies in OH can be managed by copying source code into the OH bundles so maybe you or Kai can provide some information on how this works for other bundles. 

I don't believe that dependency management should be such a big issue and I'd certainly not anticipated being blocked with this while splitting the dongle implementations, but here we are ;)

tomhoefer:
> However I understood that in your system, you would manage the libs directly through the TP and you didn't want the JARs in the bundles (hence the separation of the libs and the bundle).

At the end we will always compile from source in our system (even the core framework against our special target platform). 

> But that's not what is proposed now. Now we are simply consuming the library source into the OH binding.

No - the proposal is that the core framework is added to the newly created zigbee project.

kaikreuzer:
> No - the proposal is that the core framework is added to the newly created zigbee project.

Right. More or less what you have after [this](https://github.com/openhab/zigbee/commit/bb0947304f2bc5bf41a44013c7e6008c16120d6d) change: https://github.com/openhab/zigbee (this isn't building, so don't even try).

>  I'd be happy to take this forward as an Eclipse project if needed - that said, I'm not really sure that the namespace makes so much difference

It does! As discussed, this openhab/zigbee repo is only meant as a working place from which I'd like to create a PR towards ESH from asap - otherwise I'm pretty sure that time will fly by and we won't get any legal clarifications off the ground.

cdjackson:
> But that's not what is proposed now. Now we are simply consuming the library source into the OH binding.
> 
> No - the proposal is that the core framework is added to the newly created zigbee project.

Ok, then I am quite confused as to what the other repo is because it definitely contains both the binding and cluster library source code. I understood from Kais comment that this was a consolidated repository to contain both sources in a single place (and this is what I see).

To me, if we are going to move the repository, we should move it as it currently stands. Otherwise it's really going to cause problems - for sure the other repo from Kai is not usable for me.

However, If we just move the repository, it doesn't solve any of the issues that you have regarding the management of the dependency, so we are no further along :( .

> It does!

I don't really understand why - there are many OS projects that have a .com namespace. Anyway -we don't need to worry about that ;) Let's find a way to solve the dependency issue that we have.

tomhoefer:
> ok, then I am quite confused as to what the other repo is because it definitely contains both the binding and cluster library source code.

This was only a proposal made by @kaikreuzer how the organization of the source code could be easily managed. I think Kai will agree to clear the current content of the repo so that we can add the core framework.

> To me, if we are going to move the repository, we should move it as it currently stands. 

Agree - refactoring can be done later

cdjackson:
> This was only a proposal made by @kaikreuzer how the organization of the source code could be easily managed. I think Kai will agree to clear the current content of the repo so that we can add the core framework.

Ok, that's great - I didn't realise this was the case. We are still left with the question of how to handle the dependencies since the location of the repository won't really change this  ;)

kaikreuzer:
> I think Kai will agree to clear the current content of the repo so that we can add the core framework.

As you can see in my https://github.com/openhab/org.openhab.binding.zigbee/pull/159#issuecomment-375623928 above, I have already cleared the repo to just contain the core framework and nothing else. And as I mentioned there, this was just meant to look at and not to use. So yes, clearly the repo should be wiped and started with a clean history.

> I don't really understand why - there are many OS projects that have a .com namespace

If you didn't seem to have had a close look at the repo - the code specifically had the org.eclipse namespace as this is important for code that's panned to be contributed to Eclipse :-)
There is therefore one thing to remember: You won't be able to move ALL content from the current repo to openhab/zigbee, but only the stuff with a clean contribution history, which is under EPL - the rest (the auto-coder, if I got you right?) would have to stay outside.

cdjackson:
> If you didn't seem to have had a close look at the repo - the code specifically had the org.eclipse namespace as this is important for code that's panned to be contributed to Eclipse :-)

Yes, I had seen that, but was confused with the fact that both sets of source were in the same structure... Anyway, let's move on ;)

I would really like to get the split bundles merged - even if we haven't moved the cluster library yet. I think we need to address the autocoder as without that, life will be messy and very difficult.  I'll discuss with Thomas getting this done.

If we can agree how to handle the dependencies, then it shouldn't matter too much if we point them at my existing repo, or the OH repo (right?). Then in a month or so I can maybe look at sorting out the whole repo (ie autocoder) move so that everything can migrate across. At the moment I'm completely blocked without a lot of messing around as the workspace is completely different so any changes now need to be manually moved. So, I'd really appreciate it if we can agree how to handle the dependencies, and then sort out the repo move in a slightly slower time (which is what myself and Thomas discussed last week and I think was agreed).

Thanks.

tomhoefer:
> I would really like to get the split bundles merged - even if we haven't moved the cluster library yet. I think we need to address the autocoder as without that, life will be messy and very difficult. I'll discuss with Thomas getting this done.

Sure

> If we can agree how to handle the dependencies, then it shouldn't matter too much if we point them at my existing repo, or the OH repo (right?). 

Yes - but as already mentioned I am not the expert of the openhab build infrastructure. At least I will be able to manage it for our system until it is going to be contributed to openhab with corresponding eclipse namespace

So if we are all happy to contribute of the core framework to a separate openhab zigbee repo and if Kai can manage the dependencies for openhab then I am finally ok with this change :)

kaikreuzer:
> it shouldn't matter too much if we point them at my existing repo, or the OH repo (right?).

Well, they are not pointing to any repo, my idea was to have it included in the same Maven reactor, so that the dependencies are found. As you are not using Tycho, but the Felix Bundle Plugin for the OSGi build, I am not at all certain anymore whether this can work.

> Then in a month or so I can maybe look at sorting out the whole repo (ie autocoder) move so that everything can migrate across.

If this is realistic and you say that the core framework libs are rather stable already, I'd think the easiest way to move forward is to include the external bundles in the target platform (https://github.com/openhab/openhab-deps-repo). If you could tell me which version (1.0.7, I suppose?) should be used, I can take care of this part. All you would then have to change in this PR is to completely remove the lib folder(s) and the m2e natures from the projects.

kaikreuzer:
@cdjackson Short ping to make sure that you noticed my comment above.

cdjackson:
Sorry Kai, I didn't see a notification for this.

> my idea was to have it included in the same Maven reactor, so that the dependencies are found.

Can you explain your idea - does it also work in Eclipse IDE?

> If you could tell me which version (1.0.7, I suppose?) 

I would probably update to 1.0.8 to pickup some small fixes in the XBee driver, but I'd like to understand your alternative idea if there's a better way...

kaikreuzer:
> Can you explain your idea

Well, you do not need to care about dependencies, if all that you depend on is part of your own build - and that's what openhab-bundles is currently doing.

> does it also work in Eclipse IDE?

Yes, it works if you have all the projects in your workspace (i.e. the same as having all in the Maven reactor).

> I would probably update to 1.0.8 

Ok, if you want me to add this to the target platform, just ping me once it is available.

cdjackson:
@kaikreuzer I've released 1.0.8 this afternoon and just retested the binding (also with ZB3 :) ). I've removed the dependency JAR, so if y'all are happy, then we can look to get this all merged.

Let me know if there's anything else you want me to do...

Thanks

kaikreuzer:
Thanks @cdjackson, I have created https://github.com/openhab/openhab-distro/pull/674, which adds the libs to the target platform in version 1.0.8.

What is still missing in this PR is the removal of all Maven specific stuff - like m2e natures on projects, dependencies in the poms and definition of maven repo for zsmartsystems. None of that is needed as the resolution is done through p2 and Tycho.

cdjackson:
Thanks @kaikreuzer. I've updated the POM to remove all the dependencies. This works fine in the IDE, but somewhere maven is still picking up 1.0.23 and the build is failing. If I manually change the POM to 1.0.24 and build locally, it works fine.

kaikreuzer:
@cdjackson It is not just the poms, it is also the m2e nature on the projects that need to be removed as I said. Otherwise there are error markers on the project in the IDE.

Once you have adapted that, is it ok from your end to merge? I.e. is all code up to date, so that I can adapt the distro to pick up these separate bundles instead of the old single one?

Wrt the Maven build: Yes, I still have to update the dependencies reference in the parent pom - will do so soon.

kaikreuzer:
FTR: https://github.com/openhab/openhab-core/pull/327

cdjackson:
Hopefully I've removed all the m2e bits from the project.

Yes, once you're happy with the build then it can be merged.

kaikreuzer:
Ok, perfect, many thanks! Even Jenkins is now happy 👍

kaikreuzer:
Will now try to update the distro accordingly...

cdjackson:
Thanks Kai.

kaikreuzer:
Ok, I hope that https://github.com/openhab/openhab-distro/pull/676 will make the Zigbee binding work smoothly in the distro - for openHAB, we are simply adding all the bundles, when the user wishes to install the binding. Not really ideal to have 8 bundles for a single binding, but you didn't like my 4-bundle-solution... 🙄

Issue #178, opened 2018-04-10
Support discovery of USB dongles


hsudbrock:
**Context**
With https://github.com/eclipse/smarthome/pull/5315, Eclipse SmartHome supports the discovery of USB dongles which provide a serial port. To participate in this discovery, bindings can provide an instance of `org.eclipse.smarthome.config.discovery.usbserial.UsbSerialDiscoveryParticipant`. This `UsbSerialDiscoveryParticipant` will be called for newly discovered USB dongles with serial port, getting information about the USB dongle, the serial port, and the USB interface attached to the serial port. The `UsbSerialDiscoveryParticipant` can then provide `DiscoveryResult`s for supported USB dongles.

**Issue**
The ZigBee binding should participate in this discovery, and provide `DiscoveryResult`s for ZigBee USB dongles.

hsudbrock:
I would like to provide a first PR for this issue, enabling the discovery of the USB dongles that are available to me:
* bitronVideo ZigBee dongle (Ember chip, "BV 2010/10"); USB vendor ID 10c4, USB product ID 8b34
* Qivicon ZigBee dongle (Telegesis chip); USB vendor ID 10c4, USB product ID 8293
* Qivicon ZigBee dongle (Telegesis chip); USB vendor ID 10c4, USB product ID 89fb

cdjackson:
Thanks Henning - I was hoping you were planning to add this :)

puzzle-star:
Hi @hsudbrock,

I have seen that the discovery of the USB devices, and naming the different dongles via udev, do not seem to interface very well (at least for me).

The discovery participant is returning the /dev/ttyUSBx (for instance) and using the discovered device name as representationProperty. But this name might change between reboots, so normally, when using different dongles, we symlink to it (i.e. /dev/telegesis-1 -> /dev/ttyUSB0) via udev rules.

With the USB discovery, if for some reason the devname changes between reboots, the dongle becomes unusable. On the other side, if after discovering it I manually change the serial port of the device to the persistent name symlink, it will keep on appearing as new device in the inbox (yes, I can hide it, but it is annoying and maybe misleading for current users).

Changing the representationProperty to the serial number (or maybe vip_pid_serial) might prevent it from appearing again in the inbox (not sure), but would still require to manually change the serial port to the symlink.

Would it make sense to change (this is ESH, not OH):

```
Path devicePath = Paths.get(devDirectory).resolve(serialPortName);`
```

[...]

PR #179, opened 2018-04-12
[178] Support Discovery of USB dongles (for some Telegesis and one Ember dongle)

hsudbrock:
> Please note: This PR depends on the recently added USB-Serial discovery from Eclipse SmartHome (see https://github.com/eclipse/smarthome/pull/5315). In consequence, the build of this PR will fail until the new bundles for this discovery are contained in the stable p2 repository for Eclipse SmartHome, and this PR should not be merged before this has happened. I will add a comment to this PR once this has happened. For building the PR successfully now, replace the p2 repository URL in the ZigBee binding parent pom from https://openhab.jfrog.io/openhab/eclipse-smarthome-stable with http://download.eclipse.org/smarthome/updates-nightly.

This PR adds two discovery participants for the USB-serial discovery of Eclipse SmartHome, one for the telegesis bundle, and one for the ember bundle. These discovery participants can discover three Telegesis dongles (from the Qivicon brand) and one Ember dongle (from Bitron Video brand).

Discovery of USB dongles is held rather simple - the discovery participants check whether USB vendor and product ID match the vendor and product ID of a known ZigBee dongle. If yes, a discovery result is created using the dongle's serial number in the thing ID, and configuring the baud rate to a hard-coded baud rate with which the dongle is known to function.

I would be happy about any feedback on this PR!

Fixes https://github.com/openhab/org.openhab.binding.zigbee/issues/178.


cdjackson:
Thanks Henning.

One question without having looked at the code (and given I probably can't test this as I think it only works on Linux) -:

> and configuring the baud rate to a hard-coded baud rate with which the dongle is known to function.

I don't think there's any way to know this for sure - at least not without doing some sort of active scan? I guess we can't be sure it will function, so presumably the user can always change the baud rate to the value that they know works?

cdjackson:
> But there might be some device out there where a different baud rate is needed.

Yes, for sure this is the case. The Ember dongle will definitely require this as there are 2 common baud rates used, and other configuration parameters will also need to change. Namely, the flow control settings also change - I expect it's about 50/50 for the different configurations that are commonly used.

org.openhab.binding.zigbee.ember/src/main/java/org/openhab/binding/zigbee/ember/internal/discovery/ZigBeeEmberUsbSerialDiscoveryParticipant.java
```diff
@@ -0,0 +1,88 @@
...
+ * Currently, this {@link UsbSerialDiscoveryParticipant} supports the ZigBee dongle 'BV AV2010/10' from Bitron Video.
+ *
+ * @author Henning Sudbrock - initial contribution
+ */
+@Component(service = UsbSerialDiscoveryParticipant.class)
+public class ZigBeeEmberUsbSerialDiscoveryParticipant implements UsbSerialDiscoveryParticipant {
+
+    private static final Set<ThingTypeUID> SUPPORTED_THING_TYPES = new HashSet<>(asList(THING_TYPE_EMBER));
+
+    public static final int SILICON_LABS_USB_VENDOR_ID = 0x10c4;
```
tomhoefer: just a minor comment - such constants have been documented for the Telegesis dongle

org.openhab.binding.zigbee.telegesis/src/main/java/org/openhab/binding/zigbee/telegesis/internal/discovery/ZigBeeTelegesisUsbSerialDiscoveryParticipant.java
```diff
@@ -0,0 +1,104 @@
...
+    private static final Set<ThingTypeUID> SUPPORTED_THING_TYPES = new HashSet<>(asList(THING_TYPE_TELEGESIS));
+
+    public static final int SILICON_LABS_USB_VENDOR_ID = 0x10c4;
+
+    /**
+     * For the {@link ZigBeeTelegesisUsbSerialDiscoveryParticipant#SILICON_LABS_USB_VENDOR_ID}, this USB product ID
+     * is used for the Telegesis ZigBee dongle model ETRX3USB+8M, as well as for one version of the QIVICON ZigBee stick
+     * (which is also produced by Telegesis).
+     */
+    public static final int QIVICON_DONGLE_V1_USB_PRODUCT_ID = 0x8293;
```
tomhoefer: I am a bit confused - the black Telegesis dongle is also supported, isn´t it? If yes shouldnt we rename the prefix to Telegesis?

tomhoefer:
in general this PR looks good to me - just two minor comments

org.openhab.binding.zigbee.ember/src/main/java/org/openhab/binding/zigbee/ember/internal/discovery/ZigBeeEmberUsbSerialDiscoveryParticipant.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    @Override
+    public @Nullable DiscoveryResult createResult(UsbSerialDeviceInformation deviceInformation) {
+        if (isBitronVideoDongle(deviceInformation)) {
+            return DiscoveryResultBuilder.create(createBitronVideoDongleThingType(deviceInformation))
+                    .withLabel(createBitronVideoDongleLabel(deviceInformation))
+                    .withRepresentationProperty(CONFIGURATION_PORT)
+                    .withProperty(CONFIGURATION_PORT, deviceInformation.getSerialPort())
+                    .withProperty(CONFIGURATION_BAUD, BITRON_VIDEO_2010_10_BAUD_RATE).build();
```
cdjackson: It may also be worth setting the flow control property as well (Should be XON/OFF for this version).

hsudbrock: Ah, I wasn't aware of that option yet, thanks! I will add this with my next commit to this PR (I tested this setting with the Ember dongle, and it worked fine so far).

@cdjackson Do you have any infos on flow control for the telegesis dongles? I have looked at https://www.silabs.com/documents/public/reference-manuals/TG-PM-0518-ETRX357USB.pdf which mentions that nRTS and nCTS lines are connected to the USB bridge chip - from which I (as a layman) would deduce that hardware flow control could be used - or am I mistaken completely here?

org.openhab.binding.zigbee.telegesis/src/main/java/org/openhab/binding/zigbee/telegesis/internal/discovery/ZigBeeTelegesisUsbSerialDiscoveryParticipant.java
hsudbrock: You are right - we could add TELEGESIS to the constant name (as this product ID is used both by one version of the Qivicon dongle, and by the dongle with Telegesis label). I will add this with my next commit to this PR.

org.openhab.binding.zigbee.ember/src/main/java/org/openhab/binding/zigbee/ember/internal/discovery/ZigBeeEmberUsbSerialDiscoveryParticipant.java
cdjackson: > I tested this setting with the Ember dongle, and it worked fine so far

I think the lower speed that is used in the XON/OFF flow control means that it's not a big issue this way around when running on a reasonably fast computer. But the other way around (ie using XON/OFF instead of CTS/RTS) can completely stop the dongle from working. This is what is used when the Ember runs at 115k2.

> which mentions that nRTS and nCTS lines are connected to the USB bridge chip - from which I (as a layman) would deduce that hardware flow control could be used - or am I mistaken completely here?

I would agree with your assessment. Currently I think I don't use flow control. I think I read somewhere that it is optional, but the Telegesis runs at such a slow speed (19k2) that it shouldn't matter I think (except on a really slow computer!)

hsudbrock: Thanks for explaining! I wil thenl leave it as is for now (i.e., use the default flow control setting for the telegesis discovery result, which is 'no flow control'). We can still change this if issues arise somewhere.

hsudbrock:
I have tested changing the baud rate of a ZigBee coordinator after initial configuration using Paper UI - and it did not work in my test. It appears to me that this is not possible, because the method `ZigBeeCoordinatorHandler#handleConfigurationUpdate` ignores some configuration parameters (e.g., the serial port, the baud rate, the zigbee channel, the pan id, and several others).

As far as I see it, this would need to be changed before being able to change baud rate (and other config parameters) after initial thing creation. Or do I overlook something, @cdjackson ?

cdjackson:
I would have thought that the handler should be reinitialised if the serial port config changes, but this doesn't sound like it's implemented. I'll add that separately.

hsudbrock:
> I woulsd have thought that the handler should be reinitialised if the serial port config changes, but this doesn't sound like it's implemented. I'll add that separately.

Thanks, @cdjackson !

hsudbrock:
The USB-Serial discovery has now been deployed to https://openhab.jfrog.io/openhab/eclipse-smarthome-stable; this is, hence, no longer an impediment for merging this PR.

hsudbrock:
I rebased this PR to be able to use the newly introduced flow control constants from https://github.com/openhab/org.openhab.binding.zigbee/pull/184; I added a commit for using these constants.

org.openhab.binding.zigbee.ember/src/main/java/org/openhab/binding/zigbee/ember/internal/discovery/ZigBeeEmberUsbSerialDiscoveryParticipant.java
tomhoefer: ok

cdjackson:
You can change the baud rate etc - you ‘just’ need to restart the bundle (not nice I know, but it should work).

cdjackson:
Looks like this needs rebasing then I guess it can be merged now that #187 is merged.

hsudbrock:
I have rebased the PR, so this can be merged now.

cdjackson:
Thanks @hsudbrock - @tomhoefer I guess you will merge?

tomhoefer:
Thanks @cdjackson - I have no preference who is going to merge actually a PR 

I will merge this PR now

kaikreuzer:
Looks as if this broke the build: https://openhab.ci.cloudbees.com/job/openHAB-Distribution/1261/console
I guess we need an additional dependency in https://github.com/openhab/openhab-distro/blob/master/features/addons/src/main/feature/feature.xml#L6.

cdjackson:
I thought this was meant to be part of ESH, and therefore included in OH core? Is it just that OH hasn’t been updated since this was merged into ESH?

hsudbrock:
@kaikreuzer @cdjackson I added a PR to the `openhab-distro` repository: https://github.com/openhab/openhab-distro/pull/686
With this PR, feature verification for openhab-distro does not produce an error anymore (at least on my machine).

PR #169, opened 2018-04-01
Static thing definitions

cdjackson:
This adds the ability to define static things using the standard XML descriptions. If a static description is defined, then it will be used by the binding, otherwise the binding will create channels based on the clusters etc that it finds (ie as it currently does).  If a static description is found, then all definitions come from this definition and no service discovery is performed. This allows the description to add or remove channels as desired.

A ```ZigBeeThingTypeMatcher``` class does the matching of ZigBee devices to thing types. The matcher will search the thing registry for all ZigBee things, and requires an exclusive match for all properties defined in the thing description. So if there are two things that provide a match, then no match will be returned. The input to the matcher is the set of device properties retrieved from the BASIC cluster (manufacturer name, model name, various versions) and any of these can be used to produce a match. To avoid any confusion, properties in the thing that are to be used by the matcher start with ```typeMatcher```.

``` xml
        <properties>
            <property name="typeMatcher:vendor">Philips</property>
            <property name="typeMatcher:modelId">SML001</property>
        </properties>
```

A ```ZigBeeDeviceConfigHandler``` class supports configuration of devices. Currently this allows any attribute to be configured - it doesn't use the built in attribute definitions from the cluster library which means that non-standard attributes can be supported (eg to support a manufacturer specific cluster).

Currently attributes are described within the parameter name as there is no other alternative until [this PR ](https://github.com/eclipse/smarthome/pull/5216) is resolved and implemented. Once this issue is resolved in ESH, we can modify this to use the new system.

The elements of the attribute definition are attribute\_```endpoint```\_```clusterId```\_```attributeId```\_```dataType```.

``` xml
			<parameter name="attribute_02_0000_0033_18" type="integer">
				<label>LED Indication</label>
				<options>
					<option value="0">Disabled</option>
					<option value="1">Enabled</option>
				</options>
			</parameter>
```

Closes #101 
Requires #159 to be merged first.

Signed-off-by: Chris Jackson chris@cd-jackson.com

PR #182, opened 2018-04-15
Fix some static analyses issues

cdjackson:
This fixes some static analyses checks - I've avoid some that would cause obvious conflicts with other open PRs (namely #169).


tomhoefer:
lgtm

PR #197, opened 2018-05-06
Add UoM for Pressure and Temperature

cdjackson:
Refactors the temperature and pressure converters to use the ```QuantityType``` and adds tests.


org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterTemperature.java
```diff
@@ -79,7 +92,7 @@ public void attributeUpdated(ZclAttribute attribute) {
                 && attribute.getId() == ZclTemperatureMeasurementCluster.ATTR_MEASUREDVALUE) {
             Integer value = (Integer) attribute.getLastValue();
             if (value != null) {
-                updateChannelState(new DecimalType(BigDecimal.valueOf(value, 2)));
+                updateChannelState(new QuantityType<Temperature>(BigDecimal.valueOf(value, 2), SIUnits.CELSIUS));
```
hsudbrock: Minor comment: One could omit the type parameter for the `QuantityType` here for brevity, as it can be inferred by the type checker as `SIUnits.CELSIUS` is a `Unit<Temperature>`:

```
updateChannelState(new QuantityType<>(BigDecimal.valueOf(value, 2), SIUnits.CELSIUS));
```

(The same applies to the `QuantityType<Pressure>` in the converter for pressure above.)

This is probably a matter of personal preference, but if you like the shorter version as well, you could change it :)

org.openhab.binding.zigbee/src/test/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterAtmosphericPressureTest.java
```diff
@@ -0,0 +1,71 @@
...
+        ZigBeeEndpoint endpoint = Mockito.mock(ZigBeeEndpoint.class);
+        ZigBeeCoordinatorHandler coordinatorHandler = Mockito.mock(ZigBeeCoordinatorHandler.class);
+        Mockito.when(coordinatorHandler.getEndpoint(ArgumentMatchers.any(IeeeAddress.class), ArgumentMatchers.anyInt()))
+                .thenReturn(endpoint);
+
+        ZigBeeConverterAtmosphericPressure converter = new ZigBeeConverterAtmosphericPressure();
+        ArgumentCaptor<ChannelUID> channelCapture = ArgumentCaptor.forClass(ChannelUID.class);
+        ArgumentCaptor<State> stateCapture = ArgumentCaptor.forClass(State.class);
+        ZigBeeThingHandler thingHandler = Mockito.mock(ZigBeeThingHandler.class);
+        Channel channel = new Channel(new ChannelUID("a:b:c:d"), "");
```
hsudbrock: One could use `ChannelBuilder.create(new ChannelUID("a:b:c:d"), "").build()` here, to avoid the deprecation warning for the `new Channel(...)` call.

If you think that this is worth to change as well: The same applies to the second new unit test.

hsudbrock:
I was wondering whether it is sufficient to add the dimension information to the `item-type` in the XML declaration of the `channel-type`, or whether it must also be included in the item type when creating the channel via the `org.eclipse.smarthome.core.thing.binding.builder.ChannelBuilder`.  Currently, the call to the `ChannelBuilder` still passes `Number` as item type, and not `Number:Pressure` or `Number:Temperature` (e.g., in `ZigBeeConverterAtmosphericPressure` line 99/100).

Do you know whether it is sufficient to add he dimension information in the XML file?

cdjackson:
Good question - I'm not sure. I will have a play and see what happens in both instances. I know it generally seems to work as it is at the moment, although I've not tried to display in a different local.

cdjackson:
I've updated to add the UoM to the Item - it must be more correct to have it I think even if it does work without. WDYT?

hsudbrock:
After thinking about this a little, I find it a little strange that the `ChannelBuilder` takes both an `acceptedItemType` and a `channelTypeUID` (which references a channel containing an item type); it is unclear to me right now which item type is used, in case the two item types differ from each other.

So I agree that it sounds more correct to use the same item type in both places.

Could you also add the quantity type to the channel creation in the `ZigBeeConverterTemperature`? (I think it is right now only added for the pressure-related converter.)

cdjackson:
Thanks Henning.  I've added the type to the temperature now...

PR #204, opened 2018-05-11
updated and restructured poms

kaikreuzer:
The current [release build](https://openhab.ci.cloudbees.com/view/Sandbox/job/sandbox-openhab2-release/) keeps failing, which seems to be due to problems in the pom structure and parent references.

This PR tries to clean it up and to re-sync it with the poms that are in openhab2-addons.

### DIFF
--- a/org.openhab.binding.zigbee.cc2531/pom.xml
+++ b/org.openhab.binding.zigbee.cc2531/pom.xml
@@ -7,7 +7,7 @@
         <groupId>org.openhab.binding.zigbee</groupId>
         <artifactId>pom</artifactId>
         <version>2.3.0-SNAPSHOT</version>
-		<relativePath>../pom.xml</relativePath>
+        <relativePath>../pom.xml</relativePath>
     </parent>
     
     <groupId>org.openhab.binding.zigbee</groupId>
--- a/org.openhab.binding.zigbee.cc2531/pom.xml
+++ b/org.openhab.binding.zigbee.cc2531/pom.xml
@@ -5,14 +5,15 @@
     <modelVersion>4.0.0</modelVersion>
     <parent>
         <groupId>org.openhab.binding.zigbee</groupId>
-        <artifactId>org.openhab.binding.zigbee.parent</artifactId>
+        <artifactId>pom</artifactId>
         <version>2.3.0-SNAPSHOT</version>
+		<relativePath>../pom.xml</relativePath>
     </parent>
     
-    <groupId>org.openhab.binding</groupId>
+    <groupId>org.openhab.binding.zigbee</groupId>
     <artifactId>org.openhab.binding.zigbee.cc2531</artifactId>
     
-    <name>ZigBee CC2531 Coordinator Binding</name>
+    <name>ZigBee CC2531 Bridge</name>
     <packaging>eclipse-plugin</packaging>
     
 </project>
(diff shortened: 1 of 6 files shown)


hsudbrock:
If I am not mistaken, with the adapted poms the unit tests are no longer run (at least not with a `mvn clean install`).

I checked this by looking at the output of `mvn fr.jcgay.maven.plugins:buildplan-maven-plugin:list` before and after the change; before the change, `maven-surefire-plugin` is bound to maven's `test` phase; after the change, no plugins are bound to the `test` phase.

I  believe that  this happens as the `<build>` block, which configured compilation and execution of the unit tests before this change, is removed in the new version of the root pom of the binding.

Was there a reason that this is removed, or could it simply be re-added?

(I see that it could be beneficial to have the tests in extra test fragments, running them using tycho-surefire (similar to the way it is done in ESH). Is this the preferred way how tests should be run in openhab bindings, because it is more compatible with the openhab parent poms?)

kaikreuzer:
Thanks @hsudbrock, that's a valid concern.
Indeed, the tests weren't executed anymore as I aligned the pom with our other poms.
You are fully right that the tests should instead be placed in test fragments, just like we have it for any other binding. The reason that this is currently not the case here is that we had a the root of the repo being the root of the bundle source in the past (before it had been split). As this is no longer the case, there is no reason to keep the tests in the bundle, so it would be great if someone could spend the effort and refactor them into fragments.

For the time being, I have re-added the `<build>` section to the pom, so that the tests are executed again.

hsudbrock:
Thanks, @kaikreuzer (for re-adding junit, but also for updating the binding's poms in general).

To me, it looks good, but for one point where I am not sure whether this is intended: 

There is now a profile `releasebuild` which adds the ESH release p2 repo to the list of repos. I.e., when  activating this profile, both the ESH release and the ESH snapshot repository will be available; so tycho would still use the ESH snapshot bundles, as they usually have higher version.

I tested this on my machine by doing a `mvn clean install -Preleasebuild`, and in fact the maven output looks like the snapshot bundles are used (more concretely: the list  of available bundles computed by tycho contains  both snapshot and release bundles, as both p2 repos were added; the 'slice' computed by tycho, however, uses the snapshot bundles, and not  the release bundles for ESH).

Am I just using the `releasebuild` profile in an unintended way? Or  does actually something need to be changed in the poms to suppress the ESH snapshot p2 repository when that profile is activated?

kaikreuzer:
Hi @hsudbrock, as mentioned above, I have just synched the content of the pom with what we already have in [openhab-core](https://github.com/openhab/openhab-core/blob/master/poms/tycho/pom.xml#L226-L249) and [openhab2-addons](https://github.com/openhab/openhab2-addons/blob/master/pom.xml#L178-L202).

There were long discussions on how to structure that part (which I would have to dig for), but the fundamental reason is that the ESH release repo contains versions like `0.10.0.b1`, which are considered to be NEWER than any snapshot and thus they cause 0.10.0-SNAPSHOT builds to fail.
For release builds, you actually want to use exactly those and therefore the repo is added only in that case.

hsudbrock:
Thanks @kaikreuzer for the explanation. When I built the release version by simply activating the profile `releasebuild`, maven used a release p2 repo for ESH that did not contain the `0.10.0.b1` versions you mention (only `0.9.0` versions) - but this must then be something that is changed by the openhab release build (or just a misconfiguration on my machine).

cdjackson:
LGTM - thanks.

PR #205, opened 2018-05-11
Fix JavaDoc warnings

hsudbrock:
Fixes "more severe" JavaDoc warnings (in the sense of "those JavaDoc warnings that are still visible when setting doclint=none").

Signed-off-by: Henning Sudbrock <henning.sudbrock@telekom.de>

### DIFF
--- a/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/ZigBeeBindingConstants.java
+++ b/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/ZigBeeBindingConstants.java
@@ -141,12 +141,15 @@ public class ZigBeeBindingConstants {
 
     /**
      * Convert a map into a json encoded string.
+     * 
+     * @param properties a map with the to-be-converted properties.
+     * @return a String with a JSON representation of the properties.
      */
-    public static String propertiesToJson(Map<String, Object> object) {
+    public static String propertiesToJson(Map<String, Object> properties) {
         StringBuilder jsonBuilder = new StringBuilder();
         jsonBuilder.append("{");
         boolean first = true;
-        for (Map.Entry<String, Object> entry : object.entrySet()) {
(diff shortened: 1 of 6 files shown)


org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/ZigBeeBindingConstants.java
```diff
@@ -140,10 +140,7 @@ public static String getISO8601StringForDate(Date date) {
     }
 
     /**
-     * Convert a map into a json encoded string
-     *
-     * @param object
```
cdjackson: Wouldn't it be better to add the Javadoc here rather than delete the empty references (which I assume was the "severe" error?).

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeChannelConverterFactory.java
```diff
@@ -92,7 +92,6 @@ public ZigBeeChannelConverterFactory() {
      *
      * @param thingUID the {@link ThingUID} of the thing
      * @param endpoint the {@link ZigBeeEndpoint} to generate the channels for
-     * @return
```
cdjackson: Wouldn't it be better to add the Javadoc return type here rather than delete the reference?

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/ZigBeeBindingConstants.java
hsudbrock: To avoid misunderstandings: I don't consider any of these severe, I just fixed them to get rid of the warnings in Eclipse, which I like to have enabled :)

In this case (as in the case below) I found the parameters to be so clear from the context that I considered it to be safe to remove them. But I see your point as well, so I will push a commit shortly including the removed docs.

PR #208, opened 2018-05-15
ZigBeeDiscovery: Add RepresentationProperty

mdicke2s:
RepresentationProperty enables ESH to uniquely identify a device and it is used in the inbox auto ignore service

Details:
* https://www.eclipse.org/smarthome/documentation/development/bindings/thing-definition.html#representation-property
* https://github.com/eclipse/smarthome/pull/3899

Signed-off-by: Michael Reitler <github@madpage.de>

### DIFF
--- a/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/discovery/ZigBeeDiscoveryService.java
+++ b/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/discovery/ZigBeeDiscoveryService.java
@@ -165,6 +165,7 @@ public class ZigBeeDiscoveryService extends AbstractDiscoveryService
                             node.getIeeeAddress().toString());
                     DiscoveryResult discoveryResult = DiscoveryResultBuilder.create(thingUID)
                             .withThingType(thingTypeUID).withProperties(objProperties).withBridge(bridgeUID)
+                            .withRepresentationProperty(ZigBeeBindingConstants.CONFIGURATION_MACADDRESS)
                             .withLabel(label).build();
 
                     thingDiscovered(discoveryResult);
@@ -192,7 +193,9 @@ public class ZigBeeDiscoveryService extends AbstractDiscoveryService
                 Map<String, Object> objProperties = new HashMap<String, Object>(properties);
                 objProperties.put(ZigBeeBindingConstants.CONFIGURATION_MACADDRESS, node.getIeeeAddress().toString());
                 DiscoveryResult discoveryResult = DiscoveryResultBuilder.create(thingUID).withThingType(thingTypeUID)
-                        .withProperties(objProperties).withBridge(bridgeUID).withLabel(label).build();
+                        .withProperties(objProperties).withBridge(bridgeUID)
+                        .withRepresentationProperty(ZigBeeBindingConstants.CONFIGURATION_MACADDRESS)
+                        .withLabel(label).build();
 
                 thingDiscovered(discoveryResult);
 


hsudbrock:
I think that the ZigBee discovery service will never add discovery results for devices for which a thing already exists (cf. the code below the comment 'If this already exists as a thing, then no need to rediscover' in the method `ZigBeeDiscoveryService#nodeDiscover`, which also uses the IEEE address of the device). In consequence, the representation property introduced in this PR will currently not have any effect on the inbox auto ignore feature of ESH.

Is there a reason that the representation property is added to the discovery, and not directly to the thing type XML?

mdicke2s:
Generally that is correct, as long as you always use <thing-type-uid>:<bridge-uid>:<ieeeAddress> as ThingType (which applies if things are created from the binding's discovery results). I the case that a thing was created manually and it was given another name, the `ZigBeeDiscoveryService#nodeDiscover` will create a DiscoveryResult anyways. This is rather unlikely to happen - but possible.

A more frequent usage of the RepresentationProperty is within UIs that show inbox items. For instance PaperUI shows the following data per DiscoveryResult

1. Label
2. Description
3. Value of RepresentationProperty

Once we have https://github.com/openhab/org.openhab.binding.zigbee/pull/169 merged, the label will not contain the IEEE address any more and therefore a user would not be able to distinguish two devices of the same type in the inbox as shown in PaperUI

cdjackson:
Can a representation property be a configuration parameter? I assumed not, but I'm not overly familiar with the representation proberty.

hsudbrock:
Thanks @mdicke2s,  I was not aware that PaperUI displays the representation property. What about adding that representation property to the thing type XML, instead of to the discovery result?

cdjackson:
One other point -:

> Once we have #169 merged, the label will not contain the IEEE address any more and therefore a user would not be able to distinguish two devices of the same type in the inbox as shown in PaperUI

This is also not the case now - it's not related to #169. The label is made up of the manufacturer and the model ID. The IEEE address is only used if this information fails.

As I mentioned above, the IEEE address is a configuration parameter. I don't like this, but it's what was agreed previously for ZWave, so it's the same here as it was stated that users must be able to configure the address information. Given that's the case, is it possible to use a configuration parameter as the representation property?

mdicke2s:
> Is there a reason that the representation property is added to the discovery, and not directly to the thing type XML?

@hsudbrock I decided to put it in the code because then it is exactly in the same location where the property "zigbee_macaddress" is defined
Actually we may do both, that way we can be sure that also manually created things will have the RepresentationProperty set.
[https://www.eclipse.org/smarthome/documentation/development/bindings/discovery-services.html](url)
[https://www.eclipse.org/smarthome/documentation/development/bindings/thing-definition.html#representation-property](url)

@cdjackson In the DiscoveryService the property is set directly above:
`Map<String, Object> objProperties = new HashMap<String, Object>();
                    objProperties.put(ZigBeeBindingConstants.CONFIGURATION_MACADDRESS,
                            node.getIeeeAddress().toString());`
So it looks that with the name "zigbee_macaddress" we have both - a property and a configuration

cdjackson:
> In the DiscoveryService the property is set directly above:

In the code you referenced, this is the discovery result properties - and not the thing properties. I don't think the two are the same (unless it's changed at least).

> we have both - a property and a configuration

I don't believe so. 

My understanding of how this works is both parameters and properties are added to the discovery. When a discovery device is then added as a thing, the system looks at the configuration parameters, and adds the property as a config parameter if the name matches a config param, and as a property if it doesn't.

mdicke2s:
@cdjackson is right - we have "zigbee_macaddress" only as configuration and not as property (I mixed that up because in the DiscoveryService it is refered to as property). Therefore we cannot specify a RepresentationProperty easily. If it is agreed that it is required to be an configuration, I do not want to change that. Also it does not make sense to have this information redundant, because then we would have to take care of updates as well.

**Reference - a zigbee device's json representation** (from https://gateway:443/rest/things/zigbee%3Adevice%3Aea51709d%3A7cb03eaa00a90c8c)
```{
  "statusInfo": {
    "status": "OFFLINE",
    "statusDetail": "NONE"
  },
  "firmwareStatus": {
    "status": "UNKNOWN"
  },
  "editable": true,
  "label": "OSRAM CLA60 RGBW OSRAM",
  "bridgeUID": "zigbee:coordinator_telegesis:ea51709d",
  "configuration": {
    "zigbee_macaddress": "7CB03EAA00A90C8C"
  },
  "properties": {
    "zigbee_logicaltype": "ROUTER",
    "zigbee_powerlevel": "FULL",
    "addedAt": "1525844189553",
    "modelId": "CLA60 RGBW OSRAM",
    "zigbee_networkaddress": "55644",
    "zigbee_powersource": "MAINS",
    "zigbee_stkversion": "2",
    "zigbee_datecode": "20140331CNWT****",
    "zigbee_zclversion": "1",
    "zigbee_routes": "[]",
    "zigbee_lastupdate": "",
    "vendor": "OSRAM",
    "zigbee_powermode": "RECEIVER_ON_IDLE",
    "zigbee_powersources": "[MAINS, DISPOSABLE_BATTERY, RECHARGABLE_BATTERY]",
    "hardwareVersion": "1",
    "firmwareVersion": "01020510",
    "zigbee_neighbors": "[{\"joining\":\"ENABLED\",\"address\":\"0\",\"depth\":\"1\",\"lqi\":\"96\",\"macaddress\":\"000D6F000AF902EA\"}]",
    "zigbee_devices": "[]"
  },
  "UID": "zigbee:device:ea51709d:7cb03eaa00a90c8c",
  "thingTypeUID": "zigbee:device",
  "channels": []
}```

Issue #164, opened 2018-03-08
Add support for Develco han-interface - power meter readings


coandersen:
In Norway there will be installed 700 000+ power meters which gives owners access to real-time power consumption. Develco has developed a Zigbee-device which accesses the HAN-interface of the power meter. 

If possible I would like support for the device to be added to the Zigbee binding. Especially these attributes:

Id# | Name | Type | Man |   |   |   |  
-- | -- | -- | -- | -- | -- | -- | --
0x0305 | TotalReactivePower | Int32 | O |   |   |   |  
0x0304 | TotalActivePower | Int32 | O |   |   |   |  
0x0505 | RMSVoltagePhA | Uint16 | O | L1 | - | Volts | (V)
0x0508 | RMSCurrentPhA | Uint16 | O | L1 | - | Amps | (A)

[emizb-132-technical-manual-emi-norwegian-han.pdf](https://github.com/openhab/org.openhab.binding.zigbee/files/1794391/emizb-132-technical-manual-emi-norwegian-han.pdf)

cdjackson:
Any feedback on this?

coandersen:
Sorry, did not see this before now. Does this mean my device is supported by the binding? If so I will test it soon.

linuxtrap:
Hello
Did you continue work with this cdjackson? Would be nice with some Power readings to :)

PR #210, opened 2018-05-17
Add RMS current and voltage converters

cdjackson:
Addresses part of #164 

### DIFF
--- a/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterMeasurementRmsCurrent.java
+++ b/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterMeasurementRmsCurrent.java
@@ -1,0 +1,143 @@
+/**
+ * Copyright (c) 2014-2017 by the respective copyright holders.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.openhab.binding.zigbee.internal.converter;
+
+import java.util.concurrent.ExecutionException;
+
+import org.eclipse.smarthome.core.library.types.DecimalType;
+import org.eclipse.smarthome.core.thing.Channel;
+import org.eclipse.smarthome.core.thing.ThingUID;
+import org.openhab.binding.zigbee.ZigBeeBindingConstants;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.zsmartsystems.zigbee.CommandResult;
+import com.zsmartsystems.zigbee.ZigBeeEndpoint;
+import com.zsmartsystems.zigbee.zcl.ZclAttribute;
+import com.zsmartsystems.zigbee.zcl.ZclAttributeListener;
+import com.zsmartsystems.zigbee.zcl.clusters.ZclElectricalMeasurementCluster;
+import com.zsmartsystems.zigbee.zcl.clusters.ZclOnOffCluster;
+import com.zsmartsystems.zigbee.zcl.protocol.ZclClusterType;
+
+/**
+ * ZigBee channel converter for RMS current measurement
+ *
+ * @author Chris Jackson - Initial Contribution
+ *
+ */
+public class ZigBeeConverterMeasurementRmsCurrent extends ZigBeeBaseChannelConverter implements ZclAttributeListener {
+    private Logger logger = LoggerFactory.getLogger(ZigBeeConverterMeasurementRmsCurrent.class);
+
+    private ZclElectricalMeasurementCluster clusterMeasurement;
+
+    private Integer divisor;
+    private Integer multiplier;
+
+    @Override
+    public boolean initializeConverter() {
+        logger.debug("{}: Initialising electrical measurement cluster", endpoint.getIeeeAddress());
+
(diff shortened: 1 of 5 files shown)


org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterMeasurementRmsCurrent.java
```diff
@@ -0,0 +1,143 @@
...
+    @Override
+    public void disposeConverter() {
+        logger.debug("{}: Closing electrical measurement cluster", endpoint.getIeeeAddress());
+
+        clusterMeasurement.removeAttributeListener(this);
+    }
+
+    @Override
+    public void handleRefresh() {
+        clusterMeasurement.getRmsVoltage(0);
```
hsudbrock: I think that this should be `getRmsCurrent` (as this is in the converter for RMS current).

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterMeasurementRmsCurrent.java
```diff
@@ -0,0 +1,143 @@
...
+        if (cluster == null) {
+            return null;
+        }
+
+        try {
+            if (!cluster.discoverAttributes(false).get()
+                    && !cluster.isAttributeSupported(ZclElectricalMeasurementCluster.ATTR_RMSCURRENT)) {
+
+                return null;
+            } else if (cluster.getRmsVoltage(0) == null) {
```
hsudbrock: I think that this should be `getRmsCurrent` (as this is in the converter for RMS current).

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterMeasurementRmsCurrent.java
```diff
@@ -0,0 +1,143 @@
...
+        } catch (InterruptedException | ExecutionException e) {
+            logger.warn("{}: Exception discovering attributes in electrical measurement cluster",
+                    endpoint.getIeeeAddress(), e);
+            return null;
+        }
+
+        return createChannel(thingUID, endpoint, ZigBeeBindingConstants.CHANNEL_ELECTRICAL_RMSCURRENT,
+                ZigBeeBindingConstants.ITEM_TYPE_NUMBER, ZigBeeBindingConstants.CHANNEL_LABEL_ELECTRICAL_RMSCURRENT);
+
+        // return ChannelBuilder
```
hsudbrock: I would favour removing this commented code (but keep it if there's a reason for it).

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterMeasurementRmsVoltage.java
```diff
@@ -0,0 +1,143 @@
...
+        } catch (InterruptedException | ExecutionException e) {
+            logger.warn("{}: Exception discovering attributes in electrical measurement cluster",
+                    endpoint.getIeeeAddress(), e);
+            return null;
+        }
+
+        return createChannel(thingUID, endpoint, ZigBeeBindingConstants.CHANNEL_ELECTRICAL_RMSVOLTAGE,
+                ZigBeeBindingConstants.ITEM_TYPE_NUMBER, ZigBeeBindingConstants.CHANNEL_LABEL_ELECTRICAL_RMSVOLTAGE);
+
+        // return ChannelBuilder
```
hsudbrock: As above, I would favour removing this commented code (but keep it if there's a reason for it).

cdjackson: I left this in to make it easy once the static code is merged. This code was originally based off that, but as it's taking so long, I rebased it and I'd prefer if possible to just leave it and update if the static definitions are merged. If the static definitions are finally rejected, then I will update this to remove.

Or do you think we should just remove now?

cdjackson:
> whether we want to have unit tests also for converters

I do have some unit tests, but most of the converter is not so testable. The biggest part of the converter is the initialisation, and I'm not really sure there's a lot of benefit since most of the heavy lifting is done in the ZigBee libs.

> whether it would be sensible to extract common logic for converters into base classes

I originally looked at this, and have done it for IAS if you checked those converters, but IMHO there's not a lot that's really common here. Yes, they look the same, but the detailed methods being called are obviously different and I'm not sure that it's easily parameterisable.

I'm not really sure it can be sensibly subclassed very much at least - maybe there are a couple of small instances though so I'll have a look.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterMeasurementRmsVoltage.java
hsudbrock: Ah, I see; no, fine for me to keep it then.

hsudbrock:
I see your point regarding unit tests; I will have to think about a way this could be tested (but, as mentioned, out of scope of this PR).

cdjackson:
@hsudbrock I've updated following your comments - thanks.

PR #212, opened 2018-05-22
Add IAS Fire Indicator alarm

cdjackson:
This adds the FIRE IAS alarm. Note that it requires 1.0.11 of the zigbee libs (now available in the OH TP).

### DIFF
--- a/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterIasFireIndicator.java
+++ b/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterIasFireIndicator.java
@@ -16,12 +16,12 @@ import com.zsmartsystems.zigbee.ZigBeeEndpoint;
 import com.zsmartsystems.zigbee.zcl.clusters.iaszone.ZoneTypeEnum;
 
 /**
- * Converter for the IAS Standard CIE System sensor.
+ * Converter for the IAS fire indicator.
  *
  * @author Chris Jackson - Initial Contribution
  *
  */
-public class ZigBeeConverterIasCieSystem extends ZigBeeConverterIas {
+public class ZigBeeConverterIasFireIndicator extends ZigBeeConverterIas {
     @Override
     public boolean initializeConverter() {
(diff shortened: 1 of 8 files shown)


org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterIasFireIndicator.java
```diff
@@ -0,0 +1,41 @@
...
+import com.zsmartsystems.zigbee.zcl.ZclCommandListener;
+import com.zsmartsystems.zigbee.zcl.clusters.iaszone.ZoneTypeEnum;
+
+/**
+ * Converter for the IAS fire indicator.
+ *
+ * @author Chris Jackson - Initial Contribution
+ *
+ */
+public class ZigBeeConverterIasFireIndicator extends ZigBeeConverterIas implements ZclCommandListener {
```
hsudbrock: Should this class also be added to the `channelMap` in the `ZigBeeChannelConverterFactory`?

hsudbrock: I think that it is superfluous to add the `implements ZclCommandListener` here, as the superclass `ZigBeeConverterIas` already implements this interface. If you think that it is valuable for documentation purposes here, leave as is (my personal preference would be to remove it).

cdjackson: Yes, I think this might have got lost when resolving some merge errors when rebasing from the static things to the master...

cdjackson: This is no doubt a cut and paste error - I agree - it should be removed.

hsudbrock:
@cdjackson The build is green, but it appears that github did not notice that. I could merge manually - ok for you?

cdjackson:
Yep - ok with me - thanks Henning.

PR #217, opened 2018-06-07
Added some examples for coordinators that are (hopefully) supported

claell:
I tried to make mostly link to the manufacturers site. For the cheap solution I linked to Aliexpress directly so one can see real examples, although they might not be the best option all time.

Signed-off-by: Claudius Ellsel <claudius.ellsel@live.de> (github: claell)

### DIFF
--- a/org.openhab.binding.zigbee/README.md
+++ b/org.openhab.binding.zigbee/README.md
@@ -16,6 +16,7 @@ Coordinators need to be installed manually and the serial port must be set.
 |----------------------------|-------------|-------------------------------|
 |[Texas Instruments CC2531EMK](http://www.ti.com/tool/cc2531emk)|[TI2531](#ti2531-coordinator)|-|
 |A cheap CC2530, CC2531, CC2538 or CC2650 Stick from eBay, Aliexpress, etc. like [this](https://de.aliexpress.com/item/Drahtlose-Zigbee-CC2531-Sniffer-software-protokoll-analyse-Bareboard-Paket-Protokoll-Analyzer-Modul-Usb-schnittstelle-Dongle-Erfassen/32852226435.html) and a module to flash the firmware like [this](https://de.aliexpress.com/item/SmartRF04EB-CC1110-CC2530-ZigBee-Module-USB-Downloader-Emulator-MCU-M100/32673666126.html)|[TI2531](#ti2531-coordinator)|-|
+|[Nortek Security & Control HUSBZB-1](https://nortekcontrol.com/products/2gig/husbzb-1-gocontrol-quickstick-combo/)|[TI2531](#ti2531-coordinator)|Stick contains both Z-Wave and Zigbee|
 |[Bitron Video ZigBee USB Funkstick](http://www.bitronvideo.eu/index.php/produkte/smart-home-produkte/zb-funkstick/)|[Ember](#ember-ezsp-ncp-coordinator)|-|
 |[Cortet EM358 USB Stick](https://www.cortet.com/iot-hardware/cortet-usb-sticks/em358-usb-stick)|[Ember](#ember-ezsp-ncp-coordinator)|-|
 |[Telegesis ETRX357USB zigbee® USB Stick](https://www.silabs.com/products/wireless/mesh-networking/telegesis-modules-gateways/etrx3-zigbee-usb-sticks)|[Telegesis](#telegesis-etrx3)|-|
--- a/org.openhab.binding.zigbee/README.md
+++ b/org.openhab.binding.zigbee/README.md
@@ -14,11 +14,10 @@ Coordinators need to be installed manually and the serial port must be set.
 
 | Name and Link              | Coordinator | Comment                       |
 |----------------------------|-------------|-------------------------------|
-|[Texas Instruments CC2531EMK](http://www.ti.com/tool/cc2531emk)|[TI2531](#ti2531-coordinator)|-|
-|A cheap CC2530, CC2531, CC2538 or CC2650 Stick from eBay, Aliexpress, etc. like [this](https://de.aliexpress.com/item/Drahtlose-Zigbee-CC2531-Sniffer-software-protokoll-analyse-Bareboard-Paket-Protokoll-Analyzer-Modul-Usb-schnittstelle-Dongle-Erfassen/32852226435.html) and a module to flash the firmware like [this](https://de.aliexpress.com/item/SmartRF04EB-CC1110-CC2530-ZigBee-Module-USB-Downloader-Emulator-MCU-M100/32673666126.html)|[TI2531](#ti2531-coordinator)|-|
-|[Nortek Security & Control HUSBZB-1](https://nortekcontrol.com/products/2gig/husbzb-1-gocontrol-quickstick-combo/)|[TI2531](#ti2531-coordinator)|Stick contains both Z-Wave and Zigbee|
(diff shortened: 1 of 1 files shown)


org.openhab.binding.zigbee/README.md
```diff
@@ -38,6 +38,14 @@ The Ember EZSP NCP (Network Co-Processor) supports the Silabs EM358 or MightyGec
 
 The thing type is ```coordinator_telegesis```.
 
+#### Examples for coordinators
```
cdjackson: Thanks again for this.

Can I suggest that instead of putting this list at the end of the coordinators, that it might be better to put the examples of each coordinator into the coordinator section? At the moment, we have a list of coordinators as subsections, then another subsection with a list of coordinators - at the same header level. 

To me this doesn't seem right and it might be better to put the examples of each coordinator directly into the coordinator section. Another alternative could be to put the list at the top of the coordinator section (ie under the higher level "Coordinators" subsection. I would suggest possibly a table format as well, with the dongle name (ie the manufacturer, and the coordinator type, and then possibly a link for more information). Your current list is just a list of coordinators, but it doesn't tell people how to configure them (ie when they add the coordinator, they still don't know that a Qivicon stick uses the Telegesis driver).

Either way is fine with me, but somehow I think we need to link the dongle types, to the coordinator sections...

Cheers
Chris

claell:
> shouldnt we also add an Ember dongle?

I just added all sticks I found in the community forum, so the list is by far not complete, I guess. I will add the stick you linked. If you know some more let me know :smiley:

org.openhab.binding.zigbee/README.md
claell: I also thought about adding them directly for each section but thought that would make it harder to compare sticks. Also since I currently have not used the binding before, I did not know much about configuration. Then I will probably go with a table.

cdjackson:
Please feel free to look as well at https://github.com/zsmartsystems/com.zsmartsystems.zigbee#zigbee-dongles-and-chipsets <https://github.com/zsmartsystems/com.zsmartsystems.zigbee#zigbee-dongles-and-chipsets>

org.openhab.binding.zigbee/README.md
claell: Regarding the position: It does not really make much sense to have the examples before introducing the supported coordinator types. If they are listed in a table they have to come afterwards, so a headline is needed to separate them. The level seems to be correct, since it is a subsection of coordinators. One could only create another subsection the containing the three different coordinator type headlines.

cdjackson:
I disagree ;)

To me, a top section that provides a summary of all the coordinators supported is fine, then subsections giving the detail of each coordinator and how to configure it makes perfect sense (to me at least ;) ). The table at the top could list all supported sticks (eg Telegesis, Qivicon - stating that these are all Telegisis sticks). The user then goes to the detailed section on how to configure the Telegesis dongle. If you just put the list into the Telegesis section, then the user doesn’t know what the stick is, and has to read all coordinator sections to find the information. Likewise, putting it at the end as now also doesn’t (IMHO) lead the user through the process and seems back to front.

Ok, at the moment the coordinator sections are all small, so it doesn’t matter so much which way we do it - the information is short, so they will find it easily. However as the docs grow, this won’t always be the case. To me, the user looks in the coordinator section to find a list of coordinators, he decides Qivicon is a Telegesis coordinator, so goes down to the telegesis section to read all about it.

WDYT?

claell:
Makes sense to me thinking about it from a user perspective. Also no big change, so I will go for putting it at the beginning.

Some further question:

The site you linked shows more dongle types than in this Readme: https://github.com/zsmartsystems/com.zsmartsystems.zigbee#dongles

- ConBee / RaspBee
- XBee

Is this Readme also outdated regarding that or are the things subsumized under the Telegesis coordinator for example?

cdjackson:
The binding supports XBee (I thought the readme had XBee in it already, but it seems now), but it doesn’t support ConBee.

claell:
Alright. Currently there is no XBee section in the Readme. I can add it also, I guess the thing type is just `coordinator_xbee`?

claell:
Also I am not sure whether "Coordinators need to be installed manually and the serial port must be set." is still applying.

org.openhab.binding.zigbee/README.md
```diff
@@ -10,6 +10,19 @@ A ZigBee Coordinator is the network controller, and is therefore the heart of th
 
 Coordinators need to be installed manually and the serial port must be set.
 
+#### Examples for coordinators
+
+| Name and Link              | Coordinator | Comment                       |
+|----------------------------|-------------|-------------------------------|
+|[Texas Instruments CC2531EMK](http://www.ti.com/tool/cc2531emk)|[TI2531](#ti2531-coordinator)|-|
+|A cheap CC2530, CC2531, CC2538 or CC2650 Stick from eBay, Aliexpress, etc. like [this](https://de.aliexpress.com/item/Drahtlose-Zigbee-CC2531-Sniffer-software-protokoll-analyse-Bareboard-Paket-Protokoll-Analyzer-Modul-Usb-schnittstelle-Dongle-Erfassen/32852226435.html) and a module to flash the firmware like [this](https://de.aliexpress.com/item/SmartRF04EB-CC1110-CC2530-ZigBee-Module-USB-Downloader-Emulator-MCU-M100/32673666126.html)|[TI2531](#ti2531-coordinator)|-|
```
cdjackson: I wonder if this shouldn't just be put into the comments section of the TI2531 EMK. These Chinese boards are the same thing - If I remember correctly TI release the circuits etc, and the Chinese are just copying it.

org.openhab.binding.zigbee/README.md
```diff
@@ -10,6 +10,19 @@ A ZigBee Coordinator is the network controller, and is therefore the heart of th
...
+|[Nortek Security & Control HUSBZB-1](https://nortekcontrol.com/products/2gig/husbzb-1-gocontrol-quickstick-combo/)|[TI2531](#ti2531-coordinator)|Stick contains both Z-Wave and Zigbee|
```
cdjackson: This is incorrect - the HUSBZB contains the Ember chipset.

org.openhab.binding.zigbee/README.md
claell: Probably true. So I will put it into the comments and restructure it a little bit. I think also the original TI2531 needs the flash hardware, right? Then I will also point that out.

org.openhab.binding.zigbee/README.md
claell: You are right, thanks. Mixed up 

HUSBZB-1 | Yes (EZSP)

with `ZNP` for TI.

claell:
Ok, added the changes. Only problem now is that other chips like CC2530 that are only available in Sticks from China are now not listed separately but in the comment field of the CC2531EMK.

cdjackson:
I've not managed to get the 2530 devices to work - I've not tried super hard, but I don't think it's guaranteed they will work. In theory I think it should as they are running the same stack as 2531, but...

claell:
No problem, it will hopefully help other people with the same problem I had.

But there is one thing: The comment for the TI stick now contains "There are also cheap copys with of CC2530, CC2531, CC2538 or CC2650 Sticks available on eBay, Aliexpress, etc."
You said

> I've not managed to get the 2530 devices to work - I've not tried super hard, but I don't think it's guaranteed they will work. In theory I think it should as they are running the same stack as 2531, but...

So should the other CC chips better be left out?

cdjackson:
> On 8 Jun 2018, at 13:49, Claudius Ellsel <notifications@github.com> wrote:
> 
> So should the other CC chips better be left out?
> 
Maybe it should be changed, yes. It might work, but as far as I know, no-one has done so, and my attempts with a 2530 that I was sent were not successful. It may be that the firmware could be changed - I didn’t try - but it was a cheap chinese stick, and out of the box it did not work. Unfortunately this can be the case with the cheap sticks - there have been a few problems with them and it’s one of the main reasons I avoid recommending the 2531.

claell:
Ok, created a new PR for that.

PR #223, opened 2018-06-19
Delete network state file on bridge removal

### LINKED ISSUE #222: Network state file not deleted when bridge is removed
When you delete the bridge the network status file is not removed. When you then create a new bridge it will get a different thing uid and as a consequence a new file will be created. So the previous file will be kept forever. I order to keep the directory clean we would like to introduce a file deletion when the bridge is removed.

The first idea was
- In the drivers (zsmartsystems) we add a method 'delete()' to the interface ZigBeeNetworkStateSerializer.
- In ZigBeeCoordinatorHandler.dispose() after shutting down the network we check if the thing status is "REMOVED" and then delete the file.

Unfortunately when dispose() is called the bridge status is not "REMOVED" but "UNINITIALIZED". So this doesn't seem to be realizable.

The second approach
- In the drivers (zsmartsystems) we add a method 'delete()' to the interface ZigBeeNetworkStateSerializer.
- In ZigBeeCoordinatorHandler we override BaseThingHandler.handleRemoval(). Here we set a flag that a bridge removal is in progress and call super.handleRemoval().
- In ZigBeeCoordinatorHandler.dispose() after shutting down the network we will use the flag and then delete the file  

Is there a better way to delete the file when you remove the bridge?

TomTravaglino:
fixes #222 

### DIFF
--- a/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeCoordinatorHandler.java
+++ b/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeCoordinatorHandler.java
@@ -22,6 +22,7 @@ import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 
 import org.eclipse.jdt.annotation.NonNull;
+import org.eclipse.jdt.annotation.Nullable;
 import org.eclipse.smarthome.config.core.Configuration;
 import org.eclipse.smarthome.core.thing.Bridge;
 import org.eclipse.smarthome.core.thing.ChannelUID;
@@ -92,7 +93,7 @@ public abstract class ZigBeeCoordinatorHandler extends BaseBridgeHandler
     private Class<?> serializerClass;
     private Class<?> deserializerClass;
 
-    private ZigBeeNetworkStateSerializer networkStateSerializer;
+    private ZigBeeNetworkStateSerializerImpl networkStateSerializer;
 
     protected ZigBeeKey networkKey;
 
@@ -112,6 +113,8 @@ public abstract class ZigBeeCoordinatorHandler extends BaseBridgeHandler
     private ScheduledFuture<?> restartJob = null;
 
     private ZigBeeNetworkMeshMonitor meshMonitor = null;
+
+    private boolean bridgeRemoved = false;
 
     public ZigBeeCoordinatorHandler(Bridge coordinator) {
         super(coordinator);
@@ -272,6 +275,11 @@ public abstract class ZigBeeCoordinatorHandler extends BaseBridgeHandler
 
             // Shut down the ZigBee library
             networkManager.shutdown();
+
+            // If the bridge has been removed then delete the network state file
+            if (bridgeRemoved) {
+                networkStateSerializer.delete();
+            }
         }
         logger.debug("ZigBee network [{}] closed.", thing.getUID());
     }
@@ -281,6 +289,7 @@ public abstract class ZigBeeCoordinatorHandler extends BaseBridgeHandler
         super.thingUpdated(thing);
         logger.debug("Updating coordinator [{}]", thing.getUID());
     }
+
 
     /**
      * Common initialisation point for all ZigBee coordinators.
@@ -323,7 +332,7 @@ public abstract class ZigBeeCoordinatorHandler extends BaseBridgeHandler
         logger.debug("Initialising ZigBee coordinator");
 
         String networkId = getThing().getUID().toString().replaceAll(":", "_");
-        ZigBeeNetworkStateSerializer networkStateSerializer = new ZigBeeNetworkStateSerializerImpl(networkId);
+        networkStateSerializer = new ZigBeeNetworkStateSerializerImpl(networkId);
 
         // Configure the network manager
         networkManager = new ZigBeeNetworkManager(zigbeeTransport);
@@ -423,6 +432,13 @@ public abstract class ZigBeeCoordinatorHandler extends BaseBridgeHandler
             restartJob = null;
             initialiseZigBee();
         }, 1, TimeUnit.SECONDS);
+    }
+
+    @Override
+    public void handleRemoval() {
+        // this flag is used in dispose() in order to do some special stuff when the bridge has been removed
+        bridgeRemoved = true;
+        super.handleRemoval();
     }
 
     @Override
--- a/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeCoordinatorHandler.java
+++ b/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeCoordinatorHandler.java
@@ -22,7 +22,6 @@ import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 
 import org.eclipse.jdt.annotation.NonNull;
-import org.eclipse.jdt.annotation.Nullable;
 import org.eclipse.smarthome.config.core.Configuration;
 import org.eclipse.smarthome.core.thing.Bridge;
 import org.eclipse.smarthome.core.thing.ChannelUID;
@@ -46,7 +45,6 @@ import com.zsmartsystems.zigbee.ZigBeeNetworkManager.ZigBeeInitializeResponse;
 import com.zsmartsystems.zigbee.ZigBeeNetworkMeshMonitor;
 import com.zsmartsystems.zigbee.ZigBeeNetworkNodeListener;
 import com.zsmartsystems.zigbee.ZigBeeNetworkStateListener;
-import com.zsmartsystems.zigbee.ZigBeeNetworkStateSerializer;
 import com.zsmartsystems.zigbee.ZigBeeNode;
 import com.zsmartsystems.zigbee.app.iasclient.ZigBeeIasCieApp;
 import com.zsmartsystems.zigbee.app.otaserver.ZigBeeOtaServer;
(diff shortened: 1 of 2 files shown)


org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/ZigBeeNetworkStateSerializerImpl.java
```diff
@@ -180,4 +180,16 @@ public void deserialize(final ZigBeeNetworkManager networkState) {
         logger.debug("Loading ZigBee network state: Done.");
     }
 
+    public synchronized void delete() {
+        final File file = new File(networkStateFilePath + "/" + networkStateFileName + networkId + ".xml");
+        boolean networkStateExists = file.exists();
+        if (networkStateExists) {
+            if (file.delete()) {
+                logger.debug("Deleting ZigBee network state: Done.");
```
hsudbrock: Suggestion: Align the debug messages for good and error case somewhat more; in particular, use the same name for the file in both (currently: 'ZigBee network state' vs. 'network state file').

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeCoordinatorHandler.java
```diff
@@ -272,6 +275,11 @@ public void dispose() {
 
             // Shut down the ZigBee library
             networkManager.shutdown();
+
+            // If the bridge has been removed then delete the network state file
```
hsudbrock: Suggestion: Remove the comment - as it basically just says what the following three lines of code say. What do you think?

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/ZigBeeNetworkStateSerializerImpl.java
```diff
@@ -180,4 +180,16 @@ public void deserialize(final ZigBeeNetworkManager networkState) {
         logger.debug("Loading ZigBee network state: Done.");
     }
 
+    public synchronized void delete() {
```
hsudbrock: Wouldn't it be worth to contribute this method to the interface in the core framework?

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeCoordinatorHandler.java
```diff
@@ -113,6 +114,8 @@
 
     private ZigBeeNetworkMeshMonitor meshMonitor = null;
 
+    private boolean bridgeRemoved = false;
```
hsudbrock: This variable is set in method `dispose` and in method `handleRemoval`. As far as I understood you in personal communication, these two methods might be called concurrently. If this is really the case, this variable should be `volatile`.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/ZigBeeNetworkStateSerializerImpl.java
hsudbrock: Suggestion: Add javadoc comment to `delete` method (also to stay consistent with the remainder of the class).

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/ZigBeeNetworkStateSerializerImpl.java
```diff
@@ -180,4 +180,16 @@ public void deserialize(final ZigBeeNetworkManager networkState) {
...
+        final File file = new File(networkStateFilePath + "/" + networkStateFileName + networkId + ".xml");
```
hsudbrock: What do you think about extracting this file construction into a helper method? (As, with your change, it is now used at three places in the class.)

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/ZigBeeNetworkStateSerializerImpl.java
```diff
@@ -180,4 +180,16 @@ public void deserialize(final ZigBeeNetworkManager networkState) {
...
+        boolean networkStateExists = file.exists();
+        if (networkStateExists) {
+            if (file.delete()) {
```
hsudbrock: Citing from the Javadoc of `File.delete`:
> Note that the {@link java.nio.file.Files} class defines the {@link java.nio.file.Files#delete(Path) delete} method to throw an {@link IOException} when a file cannot be deleted. This is useful for error reporting and to diagnose why a file cannot be deleted.

In this light, what do you think about replacing this code with the following (will also, as a nice side effect, remove the need for the explicit `file.exists()` check:
```
try {
     Files.deleteIfExists(Paths.get(networkStateFilePath + "/" + networkStateFileName + networkId + ".xml"));
     logger.debug("Deleted network state file.");
} catch (IOException e) {
     logger.error("Error deleting network state file", e);
}
```

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeCoordinatorHandler.java
```diff
@@ -425,6 +434,13 @@ private void startZigBeeNetwork() {
         }, 1, TimeUnit.SECONDS);
     }
 
+    @Override
+    public void handleRemoval() {
+        // this flag is used in dispose() in order to do some special stuff when the bridge has been removed
```
hsudbrock: Suggestion: Be more explicit, and write "to remove the serialized network info" instead of "do some special stuff".

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeCoordinatorHandler.java
TomTravaglino: `handleRemoval` is called first and the first instruction there sets this variable. When `super.handleRemoval` is invoked then `dispose` will be executed in parallel. Not sure if this could be a potential issue but yes, `volatile`
 makes it more save.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeCoordinatorHandler.java
```diff
@@ -22,6 +22,7 @@
 import java.util.concurrent.TimeUnit;
 
 import org.eclipse.jdt.annotation.NonNull;
+import org.eclipse.jdt.annotation.Nullable;
```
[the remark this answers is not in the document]
TomTravaglino: Will remove if.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeCoordinatorHandler.java
TomTravaglino: OK. I think the code is clear enough.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeCoordinatorHandler.java
TomTravaglino: OK

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/ZigBeeNetworkStateSerializerImpl.java
TomTravaglino: See discussion in #222

TomTravaglino: Thx. Totally forgot to add the javadoc.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/ZigBeeNetworkStateSerializerImpl.java
TomTravaglino: Good idea.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/ZigBeeNetworkStateSerializerImpl.java
TomTravaglino: Can change it as suggested.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/ZigBeeNetworkStateSerializerImpl.java
TomTravaglino: Will change the messages to be similar to those in the other two methods. In addition I will change `deserialize` to log the exception on error level instead of debug level.

TomTravaglino:
@hsudbrock I Tried to write a unit test. Turned out that it is very difficult and time consuming to mock everything necessary to run the test.

hsudbrock:
> I Tried to write a unit test.
Nice :)

TomTravaglino:
I updated the PR. Please review it.

cdjackson:
LGTM

PR #224, opened 2018-06-22
Integrate ZigBee console into the ZigBee binding.

### LINKED ISSUE #219: Make ZigBee console commands available in the binding
The ZigBee framework used by this binding (cf. https://github.com/zsmartsystems/com.zsmartsystems.zigbee) features a console application. This console application permits to access a ZigBee dongle via simple console commands (like, e.g., enabling/disabling join mode, listing devices in the network, getting infos about a device's endpoints, and so forth).

I think that it would be helpful (e.g., for diagnostics and for development) to make the console commands of the core framework available in the binding, so that console commands can be used to control the ZigBee dongle that is represented by a ZigBee binding bridge.

Technically, Eclipse SmartHome permits to add console commands by adding a component implementing the interface `org.eclipse.smarthome.io.console.extensions.ConsoleCommandExtension`. These console commands are then available, e.g., in the openhab Karaf console. Assume for now that there is only one ZigBee bridge thing available. One could then implement such a console command extension by simply forwarding commands to an instance of the core ZigBee framework console, where that instance is configured to use the ZigBee dongle that is also used by the bridge handler. (If there are multiple bridge things available, the command could be prepended with the bridge ID to disambiguate between bridges.)

The `ZigBeeConsole` in the core framework can be configured by providing an instance of the `ZigBeeNetworkManager` and an instance of the `ZigBeeTransportTransmit`, which are then used to access the dongle when executing console commands. To use the console from the ZigBee binding, the console could hence be configured by using the instances of these classes that are already used by the `ZigBeeCoordinatorHandler`. To make this possible, the `ZigBeeCoordinatorHandler` could be extended with getter methods for these two objects to be used when configuring the core framework console instance.

My suggestion would be to add an additional bundle for such a console extension in the ZigBee binding, and to add an additional Karaf feature for that console extension to openhab (containing this additional ZigBee binding bundle, and the bundle(s) containing the core framework console implementation). By doing so, openhab users would be able to install the ZigBee binding either without console, or with console.

This is only a rough outline of such a console command extension, with the goal to get first feedback before starting work on a PR (which might require a PR with some rather small changes in the ZigBee core framework as well, for simplifying the integration).

I would be happy about any feedback!

hsudbrock:
This PR integrates the console commands from the ZigBee core framework into the binding. Resolves #219.

The integration of the core framework console into the binding consists of three new bundles: One general console bundle, and two bundles containing Ember-specific and Telegesis-specific console commands, respectively. Multiple bundles are used so that the general console bundle can be installed even if the binding bundles providing Ember and Telegesis support, respectively, are not available.

The general console bundle contains a console command extension that simply delegates console commands to ZigBee console commands from the core framework console. For executing the commands, the network manager from the bridge is used (a getter for it is added with this PR). To collect available commands, a hook for providing ZigBee console commands is provided. This hook is used by three ZigBee command providers: one providing ZigBeec console commands that do not require a specific dongle type, one providing Ember-specific commands, and one providing Telegesis-specific commands.

If there is a single ZigBee bridge available, the commands can be issued without specifying which ZigBee bridge to use. If  there are multiple ZigBee bridges available, there is one special command `setBridgeUid` which can be used to set the UID  of the bridge whose network manager shall be used for running the commands. This UID is simply stored in the console command extension, and will be lost on console bundle restart. This approach makes it - IMHO - simpler  to select among bridges than an approach where the bridge UID has to be specified as command parameter.

In #219, we had discussed to use reflection to determine all commands available in the core framework. As mentioned in the discussion, this approach was not so easy to implement.  I, hence, use  predefined sets of commands (one for general commands, one for ember commands, one for telegesis commands). I think that it would be sensible to move those sets to the core framework, and make them available via some method call there (as this is core framework-related logic, not binding-related logic). Then the core framework could be evolved in the future to some other approach for determining all available commands, if it turns out that using sets it too cumbersome to maintain. What do you think?

Note: To make this console extension usable within openhab, it is also necessary to add corresponding  Karaf features to the openhab distro; there's no PR yet, but  I will prepare one after this PR has been merged.


org.openhab.binding.zigbee.console/src/main/java/org/openhab/binding/zigbee/console/internal/ZigBeeConsoleCommandExtension.java
```diff
@@ -0,0 +1,189 @@
...
+        Stream<ZigBeeConsoleCommand> zigBeeConsoleCommands = commandProviders.stream().map(p -> p.getAllCommands())
+                .flatMap(l -> l.stream());
+
+        List<String> result = zigBeeConsoleCommands
+                .map(consoleCommand -> String.format("%s %s - %s", consoleCommand.getCommand(),
+                        consoleCommand.getSyntax(), consoleCommand.getDescription()))
+                .map(this::buildCommandUsage).collect(toList());
+
+        result.add(buildCommandUsage(
+                "setBridgeUid [bridgeUid] - sets the ID of the bridge to be used in subsequent commands; only needed if there is >= 1 bridge"));
```
MichaelRiess: > 1 bridge?

hsudbrock:
I have updated the PR now that the openHAB target platform has been updated (i.e., the zsmartsystems console modules are now available in the target platform, and zsmartsystem has been updated to version 1.0.14 in the target platform).

org.openhab.binding.zigbee.console.telegesis/src/main/java/org/openhab/binding/zigbee/console/telegesis/internal/TelegesisZigBeeConsoleCommandProvider.java
```diff
@@ -0,0 +1,55 @@
...
+
+    public static final List<ZigBeeConsoleCommand> TELEGESIS_COMMANDS = unmodifiableList(
+            asList(new TelegesisConsoleSecurityStateCommand()));
+
+    private Map<String, ZigBeeConsoleCommand> telegesisCommands = TELEGESIS_COMMANDS.stream()
+            .collect(toMap(ZigBeeConsoleCommand::getCommand, identity()));
+
+    @Override
+    public ZigBeeConsoleCommand getCommand(String commandName, ThingTypeUID thingTypeUID) {
+        if (THING_TYPE_TELEGESIS.equals(thingTypeUID)) {
```
amitjoy: Only a single return statement would suffice. We don't need any "if condition" since you are also returning `null` as well when the command is not found.

```java
public ZigBeeConsoleCommand getCommand(String commandName, ThingTypeUID thingTypeUID) {
      return telegesisCommands.get(commandName);
}
```

org.openhab.binding.zigbee.console/src/main/java/org/openhab/binding/zigbee/console/internal/ZigBeeConsoleCommandExtension.java
```diff
@@ -0,0 +1,190 @@
...
+            if (thing != null && BINDING_ID.equals(thing.getThingTypeUID().getBindingId()) && thing instanceof Bridge) {
+                return (Bridge) thing;
+            } else {
+                throw new CommandExecutionException("Could not find bridge with ID " + bridgeUID);
+            }
+        } else {
+            List<Thing> bridges = thingRegistry.stream().filter(
+                    thing -> BINDING_ID.equals(thing.getThingTypeUID().getBindingId()) && thing instanceof Bridge)
+                    .collect(toList());
+            if (bridges.size() == 0) {
```
amitjoy:
```java
bridges.isEmpty()
```

org.openhab.binding.zigbee.console.ember/src/main/java/org/openhab/binding/zigbee/console/ember/internal/EmberZigBeeConsoleCommandProvider.java
```diff
@@ -0,0 +1,66 @@
...
+            new EmberConsoleNcpConfigurationCommand(), new EmberConsoleNcpCountersCommand(),
+            new EmberConsoleNcpValueCommand(), new EmberConsoleNcpVersionCommand(), new EmberConsoleNcpStateCommand(),
+            new EmberConsoleSecurityStateCommand(), new EmberConsoleTransientKeyCommand()));
+
+    private Map<String, ZigBeeConsoleCommand> emberCommands = EMBER_COMMANDS.stream()
+            .collect(toMap(ZigBeeConsoleCommand::getCommand, identity()));
+
+    @Override
+    public ZigBeeConsoleCommand getCommand(String commandName, ThingTypeUID thingTypeUID) {
+        if (THING_TYPE_EMBER.equals(thingTypeUID)) {
```
amitjoy: Only a single return statement would suffice. We don't need any "if condition" since you are also returning `null` as well when the command is not found.

```java
public ZigBeeConsoleCommand getCommand(String commandName, ThingTypeUID thingTypeUID) {
      return emberCommands.get(commandName);
}
```

org.openhab.binding.zigbee.console/src/main/java/org/openhab/binding/zigbee/console/internal/ZigBeeConsoleCommandExtension.java
hsudbrock: thanks, will change

org.openhab.binding.zigbee.console.telegesis/src/main/java/org/openhab/binding/zigbee/console/telegesis/internal/TelegesisZigBeeConsoleCommandProvider.java
hsudbrock: With your change, the command might return a `ZigBeeConsoleCommand` even if it not applicable for the provided `thingType`, which must not happen.

Consider, e.g., a command that is available for both Telegesis and Ember chips, with chip-specific implementations. This `TelegesisZigBeeConsoleCommandProvider` should not return this command when being queried for a command for an Ember chip. With your change, the Telegesis command would be returned (instead of returning `null`).

Or do I miss something here?

amitjoy: Ah you are right. I have missed that point. Thanks for the clarification.

org.openhab.binding.zigbee.console.ember/src/main/java/org/openhab/binding/zigbee/console/ember/internal/EmberZigBeeConsoleCommandProvider.java
hsudbrock: see analogous comment above

PR #225, opened 2018-07-20
Add module to maven build that creates a p2 repository with built bundles

hsudbrock:
The p2 repository will be created in folder `releng/p2repo/target/repository` if the maven profile `p2repo` is active (e.g., when running `mvn clean install -Pp2repo`).

This is, imho, useful when working locally on extensions of this binding, where one can then reference this local p2 repository.

The maven module introduced with this PR is analogous to the one introduced in this [PR for the zsmartsystems ZigBee library](https://github.com/zsmartsystems/com.zsmartsystems.zigbee/pull/330) (and of course also includes the fix from this [follow-up PR](https://github.com/zsmartsystems/com.zsmartsystems.zigbee/pull/334)).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -19,6 +19,7 @@
 		<module>org.openhab.binding.zigbee.telegesis</module>
 		<module>org.openhab.binding.zigbee.cc2531</module>
 		<module>org.openhab.binding.zigbee.xbee</module>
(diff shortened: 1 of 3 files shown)


hsudbrock:
Thanks @cdjackson !

PR #229, opened 2018-08-03
Improve level control converter to update conditional on the on/off state

cdjackson:
This changes the level control converter to also use the on/off cluster. Level control is only updated if the onoff state is on. This should ensure that devices that report off, but with a non-zero level are correctly handled.

### DIFF
--- a/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterSwitchLevel.java
+++ b/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterSwitchLevel.java
@@ -9,6 +9,7 @@
 package org.openhab.binding.zigbee.internal.converter;
 
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.eclipse.jdt.annotation.NonNull;
 import org.eclipse.smarthome.config.core.Configuration;
@@ -27,9 +28,14 @@ import com.zsmartsystems.zigbee.ZigBeeEndpoint;
 import com.zsmartsystems.zigbee.zcl.ZclAttribute;
 import com.zsmartsystems.zigbee.zcl.ZclAttributeListener;
 import com.zsmartsystems.zigbee.zcl.clusters.ZclLevelControlCluster;
+import com.zsmartsystems.zigbee.zcl.clusters.ZclOnOffCluster;
 import com.zsmartsystems.zigbee.zcl.protocol.ZclClusterType;
 
 /**
+ * Level control converter uses both the {@link ZclLevelControlCluster} and the {@link ZclOnOffCluster}. If the
+ * {@link ZclOnOffCluster} has reported the device is OFF, then reports from {@link ZclLevelControlCluster} are ignored.
+ * This is required as devices can report via the {@link ZclLevelControlCluster} that they have a specified level, but
+ * still be OFF.
  *
  * @author Chris Jackson - Initial Contribution
  *
@@ -37,8 +43,11 @@ import com.zsmartsystems.zigbee.zcl.protocol.ZclClusterType;
 public class ZigBeeConverterSwitchLevel extends ZigBeeBaseChannelConverter implements ZclAttributeListener {
     private Logger logger = LoggerFactory.getLogger(ZigBeeConverterSwitchLevel.class);
 
+    private ZclOnOffCluster clusterOnOffServer;
     private ZclLevelControlCluster clusterLevelControl;
     private ZclLevelControlConfig configLevelControl;
+
+    private final AtomicBoolean currentState = new AtomicBoolean(true);
 
     @Override
     public boolean initializeConverter() {
@@ -51,7 +60,7 @@ public class ZigBeeConverterSwitchLevel extends ZigBeeBaseChannelConverter imple
         try {
             CommandResult bindResponse = clusterLevelControl.bind().get();
             if (bindResponse.isSuccess()) {
-                // Configure reporting - no faster than once per second - no slower than 10 minutes.
+                // Configure reporting
                 CommandResult reportingResponse = clusterLevelControl
                         .setCurrentLevelReporting(1, REPORTING_PERIOD_DEFAULT_MAX, 1).get();
                 if (reportingResponse.isError()) {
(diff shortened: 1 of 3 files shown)


org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterSwitchLevel.java
```diff
@@ -48,23 +57,33 @@ public boolean initializeConverter() {
             return false;
         }
 
-        try {
-            CommandResult bindResponse = clusterLevelControl.bind().get();
```
hsudbrock: What is the reason that reporting is no longer setup for the level control cluster?

cdjackson: Well spotted - that is indeed an error and that section should not have been removed.

cdjackson:
I've modified this so that it remembers the last level received. So, if we have a sequence like OFF, 30%, ON, the state update will be 30% and not ON. In fact, there will never be an ON state update with this modification - but it can be set to OFF.

weakfl:
> I've modified this so that it remembers the last level received. So, if we have a sequence like OFF, 30%, ON, the state update will be 30% and not ON.

Shouldn't the `OnLevel ` be used instead of the last value (if it is set)?

cdjackson:
What do you mean by “OnLevel”?

weakfl:
> What do you mean by “OnLevel”?

`OnLevel` is defined as attribute of the Level Control server cluster:

| Identifier | Name | Type | Range | Access | Default | M/O |
| --- | --- | --- | --- | --- | --- | --- |
| 0x0011 | OnLevel | uint8 | 0x01 – 0xff | Read Write | 0xff | O |

_Command: On
Temporarily store CurrentLevel.
Set CurrentLevel to the minimum level allowed for the device.
**Move CurrentLevel to OnLevel, or to the stored level if OnLevel is not defined**, over the
time period OnOffTransitionTime._

But I don't know if this can be configured already? Didn't see it in habmin...

cdjackson:
Ok, understood - I just wasn’t sure if you meant this attribute, or something else.

I guess we’re now in an uncertain area. If a device reports 30% then ON, do you use the 30% (because it just told you it was 30%) or do you use the default OnLevel attribute? Personally, I thought that the last reported level seems best, but it could depend on the device. We will not be downloading the OnLevel attribute regularly, so this would be a static setting.

weakfl:
> If a device reports 30% then ON, do you use the 30% (because it just told you it was 30%) or do you use the default OnLevel attribute? Personally, I thought that the last reported level seems best, but it could depend on the device.

I think if OnLevel is implemented and not 0xff you **should** always use it instead of the last level. Personally I prefer the last level to be restored as well. I just remembered that an OnLevel attribute was mentioned in the specification and thought I'd mention it while you're at it.

cdjackson:
So you suggest that when the ON command is received, we have to read the OnLevel attribute, and then set the level to this value? Maybe you are right, but this is more complex and will likely cause some delay. Or, maybe it doesn’t matter as (hopefully!) devices should report their level shortly after the device turns on anyway, in which case it will be updated to the actual level.

weakfl:
> So you suggest that when the ON command is received, we have to read the OnLevel attribute, and then set the level to this value?

That's how I understand the specs: _"The attribute OnLevel (see 3.10.2.3.4) determines whether commands of the On/Off cluster have a permanent effect on the CurrentLevel attribute or not."_
See 3.10.2.1.1 or details. I think **if** you really implement this, it might be enough to read the attribute at startup to avoid delays.

> Maybe you are right, but this is more complex

I totally agree and as I said I just wanted to point it out. I didn't mean to say you must implement it :) In the end it doesn't make much of a difference anyway imho.

cdjackson:
Thinking about it some more, I think my last point is probably correct - it shouldn’t matter as the reporting should update the value anyway. And if we’re worried about reporting not working, then we should implement polling after a state change so that the binding is updated in a timely way - this is what the ZWave binding does.

Otherwise, we would need to regularly poll the OnLevel attribute to make sure it doesn’t change, or as an absolute minimum, it would need to be re-read every time the device is powered off.

cdjackson:
> On 4 Aug 2018, at 15:09, weakfl <notifications@github.com> wrote:
> 
>  I think if you really implement this, it might be enough to read the attribute at startup to avoid delays.

I don’t think this will work. The OnLevel is dynamic - it’s not a fixed level that you set once and forget. Each time you turn the light on, it will change to the last level (as you already posted earlier). I think just polling after it turns on if reporting is not implemented is probably the easiest way, but I also suggest that we try the current implementation first to see how it works before adding complexity.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterSwitchLevel.java
```diff
@@ -61,10 +72,36 @@ public boolean initializeConverter() {
...
+                }
+            } catch (InterruptedException | ExecutionException e) {
+                logger.error("{}: Exception setting on off reporting ", endpoint.getIeeeAddress(), e);
+            }
+
+            // Set the currentState to ON. This will ensure that we only ignore levelControl reports AFTER we have
+            // really received an OFF report, thus confirming ON_OFF reporting is working
+            currentState.set(true);
+
+            // Add a listener, then request the status
```
hsudbrock: Is the second part of the comment outdated ("then request the status")?

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterSwitchLevel.java
```diff
@@ -125,11 +168,19 @@ public void attributeUpdated(ZclAttribute attribute) {
         logger.debug("{}: ZigBee attribute reports {}", endpoint.getIeeeAddress(), attribute);
         if (attribute.getCluster() == ZclClusterType.LEVEL_CONTROL
                 && attribute.getId() == ZclLevelControlCluster.ATTR_CURRENTLEVEL) {
-            Integer level = (Integer) attribute.getLastValue();
-            if (level != null) {
-                updateChannelState(levelToPercent(level));
+            lastLevel = levelToPercent((Integer) attribute.getLastValue());
```
hsudbrock: This change here removes the non-null check for the last level. Is this intended, or is such a check still needed to avoid NPEs, e.g., when calling `levelToPercent`?

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterSwitchLevel.java
cdjackson: Converters should request the state - it is done systematically elsewhere, so the comment is out of date.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterSwitchLevel.java
cdjackson: Yes - it was intended as I can't see any way that it should get to here and be null. Given the notification is for an update of this specific attribute, it should not be null and I considered this should be ok to remove.

hsudbrock:
> but I also suggest that we try the current implementation first to see how it works before adding complexity.

I agree - OK to merge @cdjackson ?

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterSwitchLevel.java
hsudbrock: OK, thanks!

cdjackson:
Thanks @hsudbrock - yes, ok to merge.

Issue #83, opened 2018-01-09
Color converter potential OnOff / LevelControl issue


cdjackson:
As seen in #81 use of OnOffCluster to turn a device off can result in a discrepancy between OnOff and LevelControl reports.

Checking the ZCL spec, it seems that the ```ZclCluster.xxxWithOnOff()``` methods should be used to ensure synchronisation between the two clusters.

@puzzle-star do you have any thoughts on this?

puzzle-star:
Hi @cdjackson. Indeed, for the color cluster changes I did to support XY color, I already did it like this. If the device supports Level Control and Switch Control, it uses the xxxWithOnOff commands even for OnIff. Also, it should be taking care of the reports either from Switch or Level clusters. I will have a look at it, and I could apply the same to the dimmer channel, but I won't be able to till the weekend

cdjackson:
What is seen in #81 is that a device is reporting both on/off, and level. So a device can seemingly report OFF, but level=254. If these reports coming in the wrong way around, then it's a potential problem...

Maybe it's not an issue if LevelWithOnOff is always used - I've not checked this myself, but #82 seems to indicate this issue -:

```
2018-01-09 14:27:35.906 [DEBUG] [gesis.internal.TelegesisFrameHandler] - TELEGESIS RX: Data 52 58 3A 37 39 36 43 2C 43 30 35 45 2C 30 31 2C 31 32 2C 30 30 30 36 2C 30 37 3A 18 ED 0A 00 00 10 00 2C 2D 34 32 2C 46 46
2018-01-09 14:27:35.909 [DEBUG] [gesis.internal.TelegesisFrameHandler] - Telegesis event received: TelegesisReceiveMessageEvent [ieeeAddress=null, networkAddress=31084, profileId=49246, destinationEp=1, sourceEp=18, clusterId=6, messageData=18 ED 0A 00 00 10 00, rssi=-66, lqi=255]
2018-01-09 14:27:35.914 [DEBUG] [ngle.telegesis.ZigBeeDongleTelegesis] - Telegesis RX: TelegesisReceiveMessageEvent [ieeeAddress=null, networkAddress=31084, profileId=49246, destinationEp=1, sourceEp=18, clusterId=6, messageData=18 ED 0A 00 00 10 00, rssi=-66, lqi=255]
[...]

PR #231, opened 2018-08-04
Color control converter enhancements, conditional updates if ON/OFF

cdjackson:
This adds conditional updates to the color converter as discussed in #83. Additionally it fixes a bug where the ON/OFF state update always resulted in the last brightness being used as the last state update.

The conditional updates mean that if there is an OnOffCluster OFF received, then state updates will cease until the ON is received for the same reason as discussed in #83. If no OnOffCluster reports are received, then this should have no impact.

@puzzle-star you might like to take a look at this?


puzzle-star:
Hi @cdjackson,

It makes total sense (and I had overlooked the bug ignoring off state when updating).

My only "concern" is whether some lights might not send both reports (onoff and level) if turned on via "moveToLevelWithOnOffCommand". I understand sending both should be the expected behaviour, but I will check with my IKEA lights.

puzzle-star:
I checked with my IKEA lights. At this moment the color ones, when receiving a moveToLevelWithOnOffCommand, are reporting a change in the OnOff cluster when changing from ON /OFF, but not reporting in the Level cluster, ever. I recall this used to work, so I will need to test further and sniff the network, which I will not be able to in a couple of weeks.

My understanding is, anyway, that as far as the OnOff state is reported (as it seems to be the case), it should be safe to use the fix from @cdjackson. Maybe at some moment till the next poll the brightness in the channel will be wrong, but this should not be worse than currently not being even updated at all.

cdjackson:
>  when receiving a moveToLevelWithOnOffCommand, are reporting a change in the OnOff cluster when changing from ON /OFF, but not reporting in the Level cluster, ever.

As the reporting configuration, and commands sent have not changed, I don't think this is something that will have changed in this PR...

> My understanding is, anyway, that as far as the OnOff state is reported (as it seems to be the case), it should be safe to use the fix from @cdjackson

Yes, given that On/Off is not used at all in the current version, this must be better now. However the level is also not used now if OnOff is reported and state is OFF. The logic here is exactly the same as used in the Level converter.

I will try and do a quick check myself given the testing by @puzzle-star is only with lights (I thought it was being used with some sort of controller) I can also test this.

puzzle-star:
Back from vacation, so I now can test (my previous test were limited). It seems to be behaving correctly now (turning on/off from both OpenHab and an IKEA controller).

I have found anyway an inconsistency not really related to the binding but to the fact the ZigBee and OpenHAB have different concept of the meaning of ON, and using MoveToLevelWithOnOffCommand to turn off:
- OpenHAB ON: Move Level to 100%
- ZigBee ON: Restore last known level (unless configured differently in bulbs supporting such configuration)
- Zigbee moveToLevelWithOnOffCluster(0): move level to minimum supported by the bulb and then turn off.

So consider this scenario:
1. Bulb is ON (100% or whatever other level)
2. I turn it OFF through OpenHAB (moveToLevelWithOnOffCommand(0))
3. I turn it ON with the remote / wall ZigBee switch (which send an ON command)

Use case:
 - I leave home.
 - Entrance light (all lights indeed) are turned off automatically when I leave the area.
 - I come back.
 - I turn on the entrance light (or any other light in the house) using the ZigBee wall switch.
 - Lights are turned to their minimum level instead of being fully turned on.

I think this must be a most common scenario for people using remotes. So now, as a user I would expect the bulb to either be fully turned on or at least come back to its previous state. But instead I get the bulbs to their minimum level of intensity, as this was stored as last level because of the moveToLevelWithOnOffCommand(0)

To get a more consistent behaviour, would you consider, for both Level and Color channel converters use this approach instead?
 - Turn ON / change LEVEL to other than 0 as of today: moveToLevelWithOnOffCommand(level)
 - Turn OFF / change LEVEL to 0: offCommand()

cdjackson:
Thanks for the feedback Pedro.

Let’s cover the PR first - if it’s working ok, then I suggest we merge this and cover the other issue separately so we don’t loose it when we one day want to remember why we changed this. If you’re happy with that, I suggest we merge this, and then you open another issue (just cut and paste the info below).

cdjackson:
I have created a new issue #244 to cover the request from @puzzle-star

hsudbrock:
thanks, will merge as discussed

PR #240, opened 2018-08-21
Update firmware failure string constants

cdjackson:
This tidies up some constants - don't merge right now though.

@hsudbrock I thought I saw an issue you raised on ESH about reading i18 constants from different bundles, but I can't find it now. Did I imagine that, or was it resolved (either way I can't find it).  I think it's still an issue here as this currently doesn't load (with or without the ```@text``` which I'm not sure is needed or not at the moment).

### DIFF
--- a/org.openhab.binding.zigbee.ember/src/main/java/org/openhab/binding/zigbee/ember/handler/EmberHandler.java
+++ b/org.openhab.binding.zigbee.ember/src/main/java/org/openhab/binding/zigbee/ember/handler/EmberHandler.java
@@ -116,7 +116,7 @@ public class EmberHandler extends ZigBeeCoordinatorHandler implements FirmwareUp
                         progressCallback.canceled();
                         break;
                     case FIRMWARE_UPDATE_FAILED:
-                        progressCallback.failed("zigbee.firmware.failed");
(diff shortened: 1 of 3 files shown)


hsudbrock:
Yes, I raised an issue at ESH, it was this issue here: https://github.com/eclipse/smarthome/issues/5420; this is the PR resolving that issue: https://github.com/eclipse/smarthome/pull/5429

The issue was about i18n for thing status information; here, it is i18n for firmware update progress information, which is done in the class `ProgressCallbackImpl`, so the PR does not help here, unfortunately. 

Nevertheless, the issue here looks similar to the one I had with the thing status info: The `ProgressCallbackImpl` looks up translations in the bundle of the firmware update handler, which is in the ember/telegesis bundle in this case, but the translation is in the binding's core bundle. One could probably address this in ESH like in my PR for the thing status info (maybe one could even find a more generic way, to resolve this at all relevant places in ESH...).

cdjackson:
Ok, thanks. I assumed it was a generic fix that addressed all translations that might be in different bundles - I’ll raise another issue.

cdjackson:
@hsudbrock I guess we may as well merge this. Either way, it tidies up the constants and eliminates the duplication.

hsudbrock:
That's right, we could already merge. I hope that I'll get time to look at the ESH issue in the next days...

Issue #244, opened 2018-08-31
Restore last level in dimmer


cdjackson:
From #231 from @puzzle-star 

I have found anyway an inconsistency not really related to the binding but to the fact the ZigBee and OpenHAB have different concept of the meaning of ON, together with using MoveToLevelWithOnOffCommand to turn OFF:

* OpenHAB ON: Move LEVEL to 100%
* ZigBee ON: Restore last known level (unless configured differently in bulbs supporting such configuration)
* Zigbee moveToLevelWithOnOffCommand(0): move LEVEL to the minimum supported by the bulb and then turn OFF.

So consider this scenario:

* Bulb is ON (100% or whatever other level)
* I turn it OFF through OpenHAB (moveToLevelWithOnOffCommand(0))
* I turn it ON with the remote / wall ZigBee switch (which send an ON command)

Use case:

* I leave home.
* Entrance light (all lights indeed) are turned off automatically when I leave the area.
* I come back.
* I turn on the entrance light (or any other light in the house) using the ZigBee wall switch.
* Lights are turned to their minimum level instead of being fully turned on.

I think this must be a most common scenario for people using remotes. So now, as a user I would expect the bulb to either be fully turned on or at least come back to its previous state. But instead I get the bulbs to their minimum level of intensity, as this was stored as last level because of the moveToLevelWithOnOffCommand(0)

To get a more consistent behaviour, would you consider, for both Level and Color channel converters use this approach instead?

* Turn ON / change LEVEL to other than 0 as of today: moveToLevelWithOnOffCommand(level)
* Turn OFF / change LEVEL to 0: offCommand()

puzzle-star:
Hi @cdjackson 

Thanks for creating the issue. I have sent a pull-request with the changes (pull #246)

PR #246, opened 2018-09-02
Restore last level in dimmer (Issue #244)

puzzle-star:
As per Issue #244 (Restore last level in dimmer)

As ZigBee and OpenHAB have a different concept of the meaning of ON, using MoveToLevelWithOnOffCommand to turn OFF may produce some undesired side effects:

- OpenHAB ON: Move LEVEL to 100%
- ZigBee ON: Restore last known level (unless configured differently in bulbs supporting such configuration)
- Zigbee moveToLevelWithOnOffCommand(0): move LEVEL to the minimum supported by the bulb and then turn OFF.

So consider this scenario:

- Bulb is ON (100% or whatever other level)
- I turn it OFF through OpenHAB (moveToLevelWithOnOffCommand(0))
- I turn it ON with the remote / wall ZigBee switch (which send an ON command)

Use case:

- I leave home.
- Entrance light (all lights indeed) are turned off automatically when I leave the area.
- I come back.
- I turn on the entrance light (or any other light in the house) using the ZigBee wall switch.
- Lights are turned to their minimum level instead of being fully turned on.

I think this must be a most common scenario for people using remotes. So now, as a user I would expect the bulb to either be fully turned on or at least come back to its previous state. But instead I get the bulbs to their minimum level of intensity, as this was stored as last level because of the moveToLevelWithOnOffCommand(0)

To get a more consistent behaviour, would you consider, for both Level and Color channel converters use this approach instead?

- Turn ON / change LEVEL to other than 0 as of today: moveToLevelWithOnOffCommand(level)
- Turn OFF / change LEVEL to 0: offCommand()


cdjackson:
Thanks @puzzle-star I will try and test this over the next few days, but please can you run the formatter on the PR so we have a consistent format that won't change the next time someone touches the file.

cdjackson:
@puzzle-star can you write a few words as to why you propose this - ie why use the Off command for off rather than the original ```moveToLevelWithOnOffCommand```. I just want to have some record so that if we have issues in future, we can reference back to this.

Thanks.

puzzle-star:
Sure, but just a question: you mean in the PR itself, commit, or on this thread / post? 

Also, I did not miss your request to use the formatter: just figuring out how to do it without installing the Eclipse IDE, as I am not using it normally (I am a command-line guy...)

puzzle-star:
Hi @cdjackson,

Updated the PR description, and passed the code through the formatter.

Hope it is OK now (please let me know if not)

PR #252, opened 2018-09-05
Provide option in Ember to use boost mode (and make default) and set tx power

### LINKED ISSUE #243: Feature Request: Radio Tx Power - HUSBZB-1 range is terrible
I was using a HUSBZB-1 with another software program and had serviceable results but noticed after forming a network with openhab that the range deteriorated to a ridiculous amount (can't hit a node 20 feet away). I found that if I formed the network on the stick with the alternate program before adding it to openhab it worked better. Turns out that the other software was forming the network with a configured txpower of 8 (dBm) and the openhab stack just leaves it at 0.

I can of course initialize the stick outside of openhab, but it would be nice if I could at least configure the txpower. Obviously it would be nicer if the stack picked a model specific reasonable default.

cdjackson:
Provides two options in Ember -:

* Boost most, which improves the receive performance at the expense of power consumption. Given the coordinator will nearly always be plugged into a mains powered system, this is the default.
* TX Power - allows the transmit power to be set up to 8dBm to improve transmit performance.

Closes #243 

### DIFF
--- a/org.openhab.binding.zigbee.ember/ESH-INF/thing/controller_ember.xml
+++ b/org.openhab.binding.zigbee.ember/ESH-INF/thing/controller_ember.xml
@@ -44,6 +44,26 @@
                 </options>
             </parameter>
 
+            <parameter name="zigbee_powermode" type="integer">
+                <label>Power Mode</label>
+                <description>Sets the Ember power mode. Boost mode will improve receive and transmit performance.</description>
+                <default>1</default>
+                <options>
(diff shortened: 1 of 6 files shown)


cdjackson:
No - it's just waiting on someone to review it!

cdjackson:
Unfortunately you can't approve - I still get merging blocked.

cdjackson:
Thanks @hsudbrock - that gives me a green light so I'll merge.

PR #254, opened 2018-09-06
Add door lock state channel

cdjackson:
Adds a door lock channel (untested)


cdjackson:
It is worth noting that this has been tested now.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/ZigBeeBindingConstants.java
```diff
@@ -127,6 +127,10 @@
     public static final ChannelTypeUID CHANNEL_ELECTRICAL_RMSCURRENT = new ChannelTypeUID(
             "zigbee:electrical_rmscurrent");
 
+    public static final String CHANNEL_NAME_DOORLOCK_STATE = "doorlockstate";
+    public static final String CHANNEL_LABEL_DOORLOCK_STATE = "Door Lock State";
+    public static final ChannelTypeUID CHANNEL_DOORLOCK_STATE = new ChannelTypeUID("zigbee:door_state");
```
hsudbrock: As far as I see, for the other `ChannelTypeUID` constants, the channel type UID is `zigbee:<type-id-from-channels.xml>`. Here, this is different (`zigbee:door_state` instead of `zigbee:door_lock`, where `door_lock` is the channel type ID from `channels.xml`).

Should this be aligned, or is this on purpose?

cdjackson:
> This PR does not add support for doorlock PINs

Correct - this only adds support for the door itself. Clearly there is a lot more that needs to be added, but this is what was requested.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/ZigBeeBindingConstants.java
cdjackson: Sorry - I'm not sure what you mean - I think this is aligned with the XML isn't it? Both the XML and the constants are the same I think? (ie ```zigbee:door_state```).  Maybe I misunderstand something?

hsudbrock: What I mean is:
* In `channels.xml`, line 35 is ` <channel-type id="door_lock">`
* In `ZigBeeBindingConstants.java`, line 132, it says `new ChannelTypeUID("zigbee:door_state");`.

It looked to me as if it should be either `door_lock` or `door_state` in both, and not different strings as is now. But maybe I am also mistaken here?

cdjackson:
Hmmm - on my local machine, this is correct. It looks like I’ve not pushed the latest updates following testing.  I’ll update.

PR #255, opened 2018-09-07
Initial code additions for setting join key, and support for Ember coordinator

cdjackson:
Provides a mechanism to set ZigBee Install Codes, and implements this in the Ember coordinator.


triller-telekom:
I have a question on the way you implemented the discovery for zigbee 3 devices. In the README file you say

> ZigBee 3 requires that devices use an install code to securely join the network. This must be added
to the binding before the discovery starts.

So I am wondering whether this "must before" is due to the zigbee specification or if it is because of the way you implemented it.

Is it possible to discover a zigbee 3 device by just listening to the zigbee radio traffic? I.e. am I able to "see" the device with its zigbee network address but just cannot pair it without its new key?

Or does the zigbee specification enforce to set the new key before I start a scan and then only find the device with such a key?

Background to my question is: Having a temporary key as a parameter on a thing doesn't look like a nice solution to me. If it is possible tos ee the device I'd rather put a discovery result without the key in the inbox and add the necessary key when approving the thing from the inbox. Otherwise we should think about a way to handle this in a generic manner in ESH.

cdjackson:
> Having a temporary key as a parameter on a thing doesn't look like a nice solution to me

Should have commented on this. This is NOT a temporary key on the thing. It is not added to the thing (it's on the bridge) and it is not saved at all in the thing configuration. It is passed directly to the dongle only and it will normally only be stored in the dongle for a short time as well (nominally 5 minutes).

triller-telekom:
> If the device doesn't revert to the well known key, and you don't enter a join code, then the device will never be detected as it will not know the key and will not be allowed to join the network by the trust centre.

So are you saying that without the key, the coordinator does not even "see" that the device exists?

What does the coordinator see from such a new zigbee 3 device if there is no join code set? And I guess if it is set we see (and can request) everything that we know about current zigbee 1 and 2 devices, right?

> This is NOT a temporary key on the thing. It is not added to the thing (it's on the bridge)

Still, I would argue that it is a hack to enable someone to pass this join key to the zigbee chip through the bridge configuration, which should only contain static values to be able to talk to the bridge-device.

cdjackson:
> So are you saying that without the key, the coordinator does not even "see" that the device exists?

Basically, yes. It's the same thing as if the device doesn't have the link key in a standard ZigBee network - it can't join without the key, and the binding will receive no notification that the device even exists.

> What does the coordinator see from such a new zigbee 3 device if there is no join code set?

The coordinator will see some low level packets, but they won't be decoded as they will fail decryption, and the binding will see nothing.

> And I guess if it is set we see (and can request) everything that we know about current zigbee 1 and 2 devices, right?

Yes - it's the same - this is only for joining, and this is basically to provide a completely secure system which is one of the main features of ZigBee 3.

> Still, I would argue that it is a hack to enable someone to pass this join key to the zigbee chip through the bridge configuration, which should only contain static values to be able to talk to the bridge-device.

Maybe, but what do you suggest? This is probably best to use the system for passing device configuration information, but that's been under discussion for the past few years and doesn't yet exist. I see no other option, and I can't leave this feature out until there is a better solution. If a better solution comes along one day, then we can migrate it.

hsudbrock:
@cdjackson I just saw that you requested a review - as discussed last week this was the first of the open PRs I had a look at :) I was waiting with commenting due to the discussion initiated by @triller-telekom. 

The reason for my waiting is that "injecting" (in lack for a better word) the install code for a ZigBee device into the trust center via a config parameter of the bridge that is immediately removed after the config update did not entirely match my intuition of a config parameter for the bridge. Isn't the device-specific install key rather something that should be configured on the thing? But then, if one wants to use discovery there is no thing yet on which the install code could be configured. I.e., I could not come up with something better that wouldn't look "odd" in some other way; hence, I was happy about your question 'Maybe, but what do you suggest?' above.

I suggest to wait until @triller-telekom has news.

cdjackson:
Hmmm - I’m a bit confused. How can it be added to a thing if the thing does not exist yet? Do you want to discuss this (I wasn’t aware that @triller_telecom was doing anything to be honest - is there anything I can help with to get this moving?).

I agree that maybe this isn’t the best fit, but there’s nothing better and we’ve been trying to get this implemented for about 6 months and I’m really not keen to wait another couple of months (or more) while something else gets implemented. Can you give me a timescale?

cdjackson:
@hsudbrock please understand what this is doing. This is telling the trust centre that it is allowed to let a device join the network. The trust centre is the coordinator, so either way, this information needs to be sent to the coordinator, and not to a new thing which doesn't exist. As stated above, it is even impossible to link this to a thing, as the thing doesn't exist. Until the TC knows this key, it can't even find out about the device, so it can't be linked to a thing.

As I said to @triller-telekom a better fit for this is the new device configuration - but that doesn't yet exist, and it's been on the todo list for a couple of years. I don't want to wait for this, or some other feature to be added before we can include this feature.

Feel free to give me a call if you want to discuss this as I feel that we could spend a lot of time doing so on here. If @triller-telekom is working on something again, maybe I can help, but I'd like to get this moving please.

hsudbrock:
@cdjackson I think I do understand what this is doing - in particular, I understand that the install key ultimately has to be set in the trust center, and not on the device.

My comment above had the purpose to (a) ping @triller-telekom as he had not replied to your question yet, and I thought that he might perhaps have some valuable comment, and (b) to show you that this is not forgotten on my side and that I also want to push this forward. So be assured, I also like to have this feature soon and not wait and discuss the perfect solution for a long time (in particular, as someone can probably still improve this to a better solution - if there is any - after a merge).

cdjackson:
Thanks Henning. I’m happy to discuss if you like, otherwise I’ll leave it with you.

triller-telekom:
@cdjackson Sorry for my late reply. I had various offline discussions with @htreu from the Eclipse Smarthome team and also with @hsudbrock.

What @hsudbrock writes here:
> via a config parameter of the bridge that is immediately removed after the config update did not entirely match my intuition of a config parameter for the bridge

is also what bothers me the most.

So instead of doing that we came to another possible solution, which we think would be nice for the openHAB users.

So the proposal would be to create a service (lets call it `ZigbeeKeyService`) which is configurable and which takes a list of possible `installKeys`. The `ZigbeeHandlerFactory` would create and start this service once it creates a bridge and passes the `BridgeHandler` to that service (via the constructor for example).

Each time the `ZigbeeKeyService` is modified or activated it writes its `installKey`s via the bridgehandler to the coordinator.

With this solution the openHAB users, for example, can buy 10 bulbs and enter all their `installKey`s under the `service` tab of paperUI into the `ZigbeeKeyService`. So they have one single place for all their keys.

As a plus, the bridge would not have this "odd" configuration parameter which is only used to pass the value through to the coordinator.

@cdjackson What do you think about this solution?

Disclaimer: Of course the final goal should be to have something like a config flow API where openHAB users are guided through certain steps when they want to pair a device, but its a long way to have that, so I agree with you that we should find a good intermediate solution which is good for the openHAB users.

cdjackson:
To me, this seems a bit strange. Install keys are transient - they only have a life of a few minutes, so your example of adding 10 bulbs seems a bit strange and I can't quite understand it (might just be me though). In fact, I don't expect that it's even possible to add 10 bulbs like this since the coordinators can't hold that many keys if I remember correctly.

At the end of the day though, I don't mind if someone wants to create another service to support installation configuration etc. However, I would still suggest to merge this for now so we can use this while discussion, design, and implementation of an alternative takes place.

Is that ok?

hsudbrock:
I, personally, like the suggestion from @triller-telekom and @htreu - the main reason being that it nicely integrates with paperUI, where openHAB users could configure the ZigBee binding with one or more install keys. This is - but this might be personal opinion - a nicer UI than having to configure a bridge config parameter that then disappears magically.

That said, I agree with @cdjackson that this should not prevent merging of this PR. I think that @triller-telekom 's suggestion would be an improvement of the solution proposed in this PR that does not necessarily needs to be included in this PR before merging it - but I personally would like such an improvement in another PR if someone wants to contribute this.

hsudbrock:
@cdjackson Let me have a short look at the low-level details of this PR before I approve.

triller-telekom:
Sounds good to me.

@cdjackson I am not aware of how many keys can be stored within a coordinator, so my example with the 10 bulbs was just an assumption that I made. Does the specification tell how many installKeys can be stored?

I know that we only need those keys for the first time pairing, but in case the users rests his device, he could easily re-trigger a discovery if the keyservice has the installKey for it...

cdjackson:
No - the specification doesn't state how many install keys can be stored - it's implementation dependant, and is configurable depending on memory. I'm not sure I really see why you would add more than 1 at a time to be honest. As above, they are short duration and will expire after a few minutes.

And please don't get me wrong - I'm not against having such a system as described, and as I already said, I completely agree that the way I've implemented this isn't the nicest. However, given the current interfaces, I didn't see any other option, and I don't want to drag new interfaces into this PR which is likely to take some time... I'm happy to discuss this further though.

Thanks Henning - I think the merge was done with the "Update Branch" button in GH - I'll take a look at that shortly.

hsudbrock:
@cdjackson There is one duplicate left (the parameter `zigbee_initialise` is added to the file `controller_ember.xml` by this PR, and is hence twice in the file); this should be fixed before merging.

One thought that was triggered when I found this small leftover issue: In principle, I could quickly fix such small and clear issues in one of your PRs by adding a commit with a fix to the PR (at least GitHub offers me that option - I have not tried this and I won't unless I have your OK). Do you think that it would be helpful if I would quickly do such small fixes to speed things up? If you like, we can discuss in person next time we talk (or just leave as is as we work right now - of course also fine for me).

cdjackson:
Thanks Henning. I’ve removed the other duplicate - hopefully that’s all now.

Let’s discuss the issue next week.

PR #257, opened 2018-09-10
Set the default link key

cdjackson:
Recent Ember dongle driver does not set the default link key, so at the moment, the link key won't be set during initialisation.  Other dongles should also be updated in future so this ensures the standard link key is set.

### DIFF
--- a/org.openhab.binding.zigbee/README.md
+++ b/org.openhab.binding.zigbee/README.md
@@ -10,6 +10,23 @@ A ZigBee Coordinator is the network controller, and is therefore the heart of th
 
 Coordinators need to be installed manually and the serial port and baud rate must be set. These are set to match the configuration that the dongle is in. Should you wish to use a different baud rate than the default speed of the device, you must change the configuration of the dongle using some other, and then configure the binding to match your change. If in doubt, you should leave the settings at their default values which should work in most cases.
 
+#### Coordinator Configuration
+
+##### Link Key (zigbee_linkkey)
+
+The key is defined as 16 hexadecimal values. If not defined, this will default to the well known ZigBee HA link key.
+
+If defined with the word ```INSTALLCODE:``` before the key, this will create a link key from an install code which may be shorter than 16 bytes.
+
+eg ```5A 69 67 42 65 65 41 6C 6C 69 61 6E 63 65 30 39```
+eg ```INSTALLCODE:00 11 22 33 44 55 66 77```
+
+##### Network Key (zigbee_networkkey)
+
+The key is defined as 16 hexadecimal values. If not defined, a random key will be created.
+
+#### Supported Coordinators
+
 The following coordinators are known to be supported.
 
 | Name and Link              | Coordinator | Comment                       |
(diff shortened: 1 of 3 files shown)


org.openhab.binding.zigbee/README.md
```diff
@@ -10,6 +10,23 @@ A ZigBee Coordinator is the network controller, and is therefore the heart of th
 
 Coordinators need to be installed manually and the serial port and baud rate must be set. These are set to match the configuration that the dongle is in. Should you wish to use a different baud rate than the default speed of the device, you must change the configuration of the dongle using some other, and then configure the binding to match your change. If in doubt, you should leave the settings at their default values which should work in most cases.
 
+#### Coordinator Configuration
+
+##### Link Key (zigbee_linkkey)
+
+The key is defined as 16 hexadecimal values. If not defined, this will default to the well known ZigBee HA link key.
+
+If defined with the work ```INSTALLCODE:``` before the key, then this will create a link key from an install code which may be shorter than 16 bytes.
```
hsudbrock: Could these be typos?
* `with the work` -> `with the word`
* `before the key, then this will` -> `before the key, this will`

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeCoordinatorHandler.java
```diff
@@ -237,10 +249,20 @@ public void initialize() {
...
         }
 
-        logger.debug("Key final array {}", networkKey);
+        logger.debug("Network key final array {}", networkKey);
+
+        // Process the link key
+        try {
+            logger.debug("Link Key String {}", linkKeyString);
+            linkKey = new ZigBeeKey(linkKeyString);
+        } catch (IllegalArgumentException e) {
```
hsudbrock: Minor suggestion: Add a log statement within the `catch` block, so that it gets visible in the log that (and why) the system falls back to the well-known key.

That the system fell back to the well-known key is also visible (though somewhat implicit) from the log statement with the final link key array, the additional log statement would make this more explicit.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeCoordinatorHandler.java
```diff
@@ -237,10 +249,21 @@ public void initialize() {
...
+            linkKey = KEY_ZIGBEE_ALLIANCE_O9;
+            logger.debug("Link Key String has invalid format. Revert to default key.", linkKeyString);
```
hsudbrock: I believe the `{}` placeholder for the `linkKeyString` parameter is missing in the log string.

hsudbrock:
Thanks for the update, looks good to me.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeCoordinatorHandler.java
cdjackson: Thanks - cut and paste error ;)

PR #258, opened 2018-09-11
Add ASH protocol statistics channels

cdjackson:
This adds a set of ASH protocol level statistics counters to allow evaluation of the protocol performance.

### DIFF
--- a/org.openhab.binding.zigbee.ember/ESH-INF/thing/controller_ember.xml
+++ b/org.openhab.binding.zigbee.ember/ESH-INF/thing/controller_ember.xml
@@ -8,6 +8,15 @@
 		<label>Ember EM35x Coordinator</label>
 		<description>Ember EM35x NCP</description>
 
+		<channels>
+			<channel id="rx_dat" typeId="ember_ash_rx_dat" />
+			<channel id="tx_dat" typeId="ember_ash_tx_dat" />
+			<channel id="rx_nak" typeId="ember_ash_rx_nak" />
+			<channel id="tx_nak" typeId="ember_ash_tx_nak" />
+			<channel id="rx_ack" typeId="ember_ash_rx_ack" />
+			<channel id="tx_ack" typeId="ember_ash_tx_ack" />
+		</channels>
+
 		<config-description>
 			<parameter-group name="network">
 				<context></context>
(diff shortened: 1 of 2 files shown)


org.openhab.binding.zigbee.ember/src/main/java/org/openhab/binding/zigbee/ember/handler/EmberHandler.java
```diff
@@ -43,6 +48,22 @@
 public class EmberHandler extends ZigBeeCoordinatorHandler implements FirmwareUpdateHandler {
     private final Logger logger = LoggerFactory.getLogger(EmberHandler.class);
 
+    private ScheduledFuture<?> pollingJob = null;
```
hsudbrock: Very minor comment, just in case you want to change: No need to assign `null` as this is the default anyways; I, personally, prefer to omit assigning `null` to fields as it always looks a little odd at first read (probably personal preference).

org.openhab.binding.zigbee.ember/ESH-INF/thing/controller_ember.xml
```diff
@@ -8,100 +8,160 @@
...
+		<label>Data frames sent</label>
+		<description>Provides a count of DATA frames sent by the host.</description>
+		<state readOnly="true">
+		</state>
+	</channel-type>
+
+	<channel-type id="ember_ash_rx_nak">
+		<item-type>Number</item-type>
+		<label>NAK frames received</label>
+		<description>Provides a count of DATA frame with an error or that were discarded by the NCP due to lack of memory.
```
hsudbrock: Minor comment: I think that here, and in the channel types below, it should be `DATA frames` (and not `DATA frame`).

org.openhab.binding.zigbee.ember/ESH-INF/thing/controller_ember.xml
```diff
@@ -8,100 +8,160 @@
...
+		</description>
+		<state readOnly="true">
+		</state>
+	</channel-type>
+
+	<channel-type id="ember_ash_tx_nak">
+		<item-type>Number</item-type>
+		<label>NAK frames sent</label>
+		<description>Provides a count of DATA frame with an error or that were discarded by the host due to lack of memory.
```
hsudbrock: Question: Why is this "due to lack of memory", and not also due to other issues? Background of the question: I had a look into the `AshFrameHandler`, which will (as far as I see) also send a NAK if it receives a bad package. Or am I thinking in the wrong direction here?

hsudbrock:
I just see that the master branch cannot automatically be merged into the feature branch from this PR - do you want to resolve the conflicts? (It looks like it conflicts with already merged changes in the ember coordinator thing type...)

org.openhab.binding.zigbee.ember/ESH-INF/thing/controller_ember.xml
cdjackson: > Why is this "due to lack of memory"

That's just what the doc states. It might also be other reasons but the Silabs doc states the following -:

> • Purpose: Indicates receipt of a DATA frame with an error or that was discarded due to lack of memory.

cdjackson:
I've made the changes, but I've not rebased as there will be more conflicts after we merge #271 so we might as well address that first.

org.openhab.binding.zigbee.ember/ESH-INF/thing/controller_ember.xml
hsudbrock: Doesn't the comment that you are are citing refer to the `NAK`s sent by the NCP to the host, i.e., to those `NAK`s counted in the channel `ember_ash_rx_nak`?

I don't understand yet why this also applies to the `NAK`s sent by the host to the NCP - as far as I see, the host sending a `NAK` to the NCP is unrelated to lack-of-memory issues, but rather to [bad backets](https://github.com/zsmartsystems/com.zsmartsystems.zigbee/blob/e94d9d907b9c10b4e8c7ca4ea0cdab195241606e/com.zsmartsystems.zigbee.dongle.ember/src/main/java/com/zsmartsystems/zigbee/dongle/ember/internal/ash/AshFrameHandler.java#L169) and [out-of-sequence transmissions](https://github.com/zsmartsystems/com.zsmartsystems.zigbee/blob/e94d9d907b9c10b4e8c7ca4ea0cdab195241606e/com.zsmartsystems.zigbee.dongle.ember/src/main/java/com/zsmartsystems/zigbee/dongle/ember/internal/ash/AshFrameHandler.java#L211).

cdjackson:
Sorry - I missed that this was on the host side and this was a cut and paste issue between NCP and Host.

hsudbrock:
Thanks Chris for the info. Sorry when I seem picky here - this is just me trying to understand things where you have so much more knowledge...

OK for you to merge now?

cdjackson:
No problem at all - it's best to be correct.

I think it's ok for me to merge - thanks Henning.

PR #271, opened 2018-10-07
Minor updates for Ember config and Bridge config updates

cdjackson:
This tidies up a few minor issues in the Ember dongle and coordinator handler -:

* Regroups the configuration parameters, and moves them into a more logical order so they appear more meaningfully in the UI.
* Moves the processing of the keys before the initialisation code so that keys can be updated if they are modified by the init code. This ensures the UI is consistent with the configuration.
* Fixes a bug in the install code processing as Paper UI is passing NULL if it is not set.
* Ignores configuration parameters that have not changed. This prevents the bridge being reinitialised if there are really no changes to the config since PaperUI passes all configuration even if there are no changes.

### DIFF
--- a/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeCoordinatorHandler.java
+++ b/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeCoordinatorHandler.java
@@ -17,6 +17,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
@@ -82,8 +83,8 @@ public abstract class ZigBeeCoordinatorHandler extends BaseBridgeHandler
         implements Identifiable<ThingUID>, ZigBeeNetworkStateListener, ZigBeeNetworkNodeListener {
     private final Logger logger = LoggerFactory.getLogger(ZigBeeCoordinatorHandler.class);
 
-    protected int panId;
-    protected int channelId;
+    protected Integer panId;
+    protected Integer channelId;
     protected ExtendedPanId extendedPanId;
 
     private IeeeAddress nodeIeeeAddress = null;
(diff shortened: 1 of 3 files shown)


org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeCoordinatorHandler.java
```diff
@@ -180,6 +181,31 @@ public void initialize() {
             initializeNetwork = true;
         }
 
+        // Process the network key
+        try {
+            logger.debug("Network Key String {}", networkKeyString);
+            networkKey = new ZigBeeKey(networkKeyString);
+        } catch (IllegalArgumentException e) {
```
hsudbrock: May I suggest that we add a logging statement in case of `IllegalArgumentException` as well? I think that this is useful in case that a user provides a custom network key via the bridge configuration, which is invalid.

I see that one can deduce that this was the case from the log (because in this case one will see the log message "Network key initialised" below) - but this is somewhat implicit imho.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeCoordinatorHandler.java
```diff
@@ -230,40 +256,17 @@ public void initialize() {
...
             try {
-                // Reset the initialization flag
+                // Reset the initialization flag and save the keys
                 Configuration configuration = editConfiguration();
+                configuration.put(CONFIGURATION_LINKKEY, linkKey.toString());
+                configuration.put(CONFIGURATION_NETWORKKEY, networkKey.toString());
                 configuration.put(CONFIGURATION_INITIALIZE, false);
                 updateConfiguration(configuration);
             } catch (IllegalStateException e) {
                 logger.error("Error updating configuration: Unable to reset initialize flag. ", e);
```
hsudbrock: Should this error log message be updated, because now the error could be caused by either resetting the initialize flag, or by updating the keys?

Suggestion: `logger.error("Error updating configuration with link key, network key, and initialize flag. ", e);`

cdjackson:
Thanks Henning - I agree, it makes sense and I'll do the updates later today.

cdjackson:
@hsudbrock I've updated as suggested - thanks.

Issue #233, opened 2018-08-09
ZigBee bridge configuration change results in non-working things


hsudbrock:
*Context*
After certain configuration updates of a coordinator handler, the handler is reinitialized (update of baud rate, flow control, or port, see `ZigBeeCoordinatorHandler#handleConfigurationUpdate`). Reinitialization is done via a call to `dispose` followed by a call to `initialize`.

*Issue*
After the reinitialization, previously paired things are no longer usable, in the sense that commands sent to a device via the channel converters do not reach the device.

*First analysis*
The `dispose` call to the coordinator handler will do a shutdown of the `ZigBeeNetworkManager`; the `initialize` call will then create a new `ZigBeeNetworkManager`.

The converters, however, still use classes using the old network manager (e.g., a `ZclOnOffCluster` instance obtained from the old network manager). This results in exceptions like the following when using the converters:
```
00124B00028A1B7D: Exception sending command to channel zigbee:device:0138143B:00124b00028a1b7d:00124B00028A1B7D_1_switch_onoff
	at java.util.concurrent.ThreadPoolExecutor$AbortPolicy.rejectedExecution(ThreadPoolExecutor.java:2047)
	at java.util.concurrent.ThreadPoolExecutor.reject(ThreadPoolExecutor.java:823)
	at java.util.concurrent.ScheduledThreadPoolExecutor.delayedExecute(ScheduledThreadPoolExecutor.java:326)
	at java.util.concurrent.ScheduledThreadPoolExecutor.schedule(ScheduledThreadPoolExecutor.java:533)
	at com.zsmartsystems.zigbee.ZigBeeNetworkManager.scheduleTask(ZigBeeNetworkManager.java:499)
	at com.zsmartsystems.zigbee.transaction.ZigBeeTransaction.sendTransaction(ZigBeeTransaction.java:69)
	at com.zsmartsystems.zigbee.ZigBeeNetworkManager.unicast(ZigBeeNetworkManager.java:902)
	at com.zsmartsystems.zigbee.zcl.ZclCluster.send(ZclCluster.java:159)
	at com.zsmartsystems.zigbee.zcl.clusters.ZclOnOffCluster.offCommand(ZclOnOffCluster.java:344)
	at org.openhab.binding.zigbee.internal.converter.ZigBeeConverterSwitchOnoff.handleCommand(ZigBeeConverterSwitchOnoff.java:143)
[...]

PR #273, opened 2018-10-08
Reinitialise the thing and all converters if the bridge goes offline/online

cdjackson:
This is a proposed fix for #233 to ensure that the node is always initialised with the current instant of the ```ZigBeeNetworkManager```.

If the bridge goes offline, then we reset the ```nodeInitialised``` flag and stop polling (this is not really related, but I wanted to ensure it is stopped when we reinitialise). When the bridge comes back online, the ```doNodeInitialisation()``` method will be called, and the node will be initialised again.

@hsudbrock I've not really tested this - if you have the opportunity to test this it would be good. I don't see any downside of this, but happy to discuss...


hsudbrock:
Thanks @cdjackson - this fix looks simpler than what I had in mind, and I like that :)

I will try to test it soon.

hsudbrock:
First test result: I cannot see the stack traces that I described in #233 anymore - but I cannot send commands to paired devices for some other reason; I will investigate what that other reason could be.

PR #275, opened 2018-10-14
Add stack compliance level to the device properties

cdjackson:
This adds the Stack Compliance Level from the node descriptor into the device properties in the binding. @hsudbrock this will give you what you wanted wrt knowing if the device is Z3 or not. R21 is the Z3 compliance level IIRC.


hsudbrock:
Thanks Chris!

PR #277, opened 2018-10-19
Do not synchronize on 'this' when closing the serial port

### LINKED ISSUE #276: ZigBeeSerialPort can cause deadlock when closing
The `ZigBeeSerialPort` can cause a deadlock when its `close()` method is called, when using version 2.1.7r2 of the RXTX library.

The deadlock happens as follows:
1. The implementation of the `SerialPort` wrapped by the `ZigBeeSerialPort` calls the event listener callback `ZigBeeSerialPort#serialEvent(event)` in thread A.
2. Before this callback has finished processing the event, the `ZigBeeSerialPort.close()` method is called in another thread B.
3. Thread B acquires the monitor of the `ZigBeeSerialPort` object in the `close` method.
4. Thread A blocks in `serialEvent`, waiting for the monitor of the `ZigBeeSerialPort` 
5. Thread B calls `serialPort#removeEventListener()`. This method in `RXTXPort` sets a flag for stopping the native event loop, and then waits until the thread running the native event loop terminates.
6. The thread running the internal event loop, however, is thread A that currently blocks waiting for the monitor of `ZigBeeSerialPort`, and the system is in deadlock.

I believe that the simple fix should be to not remove the event listener in a block synchronized on the monitor of the `ZigBeeSerialPort` object, and I will add a PR for this.

I had a brief look at the implemenation of `gnu.io` used in openHAB, the nrjavaserial library. This implementation will not run into this deadlock, as it will stop waiting for the termination of the event loop thread after three seconds. There might still be an issue when disposing the ZigBee bridge in openHAB on slow systems, as the method from ESH that calls the bridge's `dispose()` method will stop the `dispose()` call after a timeout of 5 seconds - but this should be fixed by the deadlock patch as well.

hsudbrock:
Fixes #276 

The deadlock described in #276 happens because the call to `removeEventListener` is within a `synchronized(this)` block. This PR pushes all statements for closing the port out of this synchronized block, except the final call to `notify` (as one must own the object's monitor to notify).

@cdjackson I am wondering whether I break something by moving more than the call to `removeEventListener` out of the `synchronized` block. Do you recall the reason that the other commands for closing the port were inside the `synchronized` block?

### DIFF
--- a/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeSerialPort.java
+++ b/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeSerialPort.java
@@ -178,20 +178,21 @@ public class ZigBeeSerialPort implements ZigBeePort, SerialPortEventListener {
     public void close() {
         try {
             if (serialPort != null) {
+                serialPort.removeEventListener();
+                serialPort.enableReceiveTimeout(1);
+
+                outputStream.flush();
+
+                inputStream.close();
+                outputStream.close();
+
+                serialPort.close();
+
+                serialPort = null;
+                inputStream = null;
+                outputStream = null;
+
                 synchronized (this) {
-                    serialPort.removeEventListener();
-                    serialPort.enableReceiveTimeout(1);
-
-                    outputStream.flush();
-
-                    inputStream.close();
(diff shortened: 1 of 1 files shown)


cdjackson:
Thanks Henning. I don't see any issues with this and can't see any reason for this to be synchronised other than the ```notify```.

PR #278, opened 2018-10-19
README: fix switched numbers

t-8ch:
Not actually tested, but the other coordinator type seems more reasonable.

### DIFF
--- a/org.openhab.binding.zigbee/README.md
+++ b/org.openhab.binding.zigbee/README.md
@@ -41,7 +41,7 @@ The following coordinators are known to be supported.
 
 #### TI2531 Coordinator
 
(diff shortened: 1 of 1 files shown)


t-8ch:
Or is it called `coordinator_cc2531` the only place the `ti` prefixed is used is the README

cdjackson:
Yes - it should be cc2531 - I missed that you had changed this as well - I thought you'd just changed the numbers.

t-8ch:
I only changed the numbers, I did not get further into the code :-)

cdjackson:
There should not be need to change the code. Just the readme is wrong.

t-8ch:
Sure, I wanted to express that I did not yet know enough to recognize, that the prefix should have been changed too. Want me send another PR?

cdjackson:
Yes please.

PR #283, opened 2018-10-27
Fix removal of configuration parameter when configuration has been changed

hanneshofmann:
In this PR, we fixed the unintentional removal of configuration parameters that do not belong to an attribute (`ZigBeeDeviceConfigHandler#CONFIG_TYPE_ATTRIBUTE`). 

For instance, when a thing consists of two configuration parameters `zigbee_macaddress` and `attribute_02_0000_0033_18`, a configuration change will result in the removal of the `zigbee_macaddress` parameter value. Since this parameter is mandatory, the respective thing handler will remain `UNINITIALIZED` after the configuration has been changed.


hanneshofmann:
One note: I just had a look into the open [PR](https://github.com/openhab/org.openhab.binding.zigbee/pull/281) regarding the introduction of the `ZclClusterConfigFactory` and I assume that the issue that configuration parameters are removed would be also resolved with that PR. Because, as part of the open PR,  only changed parameters are updated and the entire configuration will be no longer replaced.

@cdjackson Would you mind that I review the open [PR](https://github.com/openhab/org.openhab.binding.zigbee/pull/281)? Consequently, we could reject this PR.

cdjackson:
> @cdjackson Would you mind that I review the open PR? Consequently, we could reject this PR.

This is fine, and I see you have raised comments there (thanks) so I will close this.

PR #287, opened 2018-10-31
Fix and improve readability of Mired/Kelvin conversion in color converter

### LINKED ISSUE #267: Color temperature converter mixes up max and min
When controlling Osram Lightify bulbs, 100% is high temp and the lower bound (which is unfortunately not 0) is low temp. With Hue bulbs the scale is reversed (100% is 2000K), and the Hue bulb scale direction is probably the one that wrong

The bug is that the converter code queries ColorTemperatureMin cluster attribute for the lower kelvin bound, but this is actually the minimum in mireds (reports as 153 (~6500K) for Hue bulbs). Since the Hue bulbs support the queries the scale gets inverted.

The Osram bulbs do not support querying the Min and Max temp so the hardcoded default of 2000 and 6500 is used.  Unfortunate side effect is that nothing below 16% is supported as a command because the bulbs actually only support down to 2700K.

On a side note, I sometimes wish that the color temperature was exposed as a number rather than dimmer either in mireds or K.

hsudbrock:
ZigBee devices return the color temperature measured in Mired, while the color temperature converter internally uses Kelvin.

This PR introduces some methods for converting between Mired, Kelvin, and percentages to make more explicit where the corresponding conversions happen in the converter.

Moreover, this PR fixes #267 by computing the maximal Kelvin value based on the minimal Mired value and vice versa (as small Mired values correspond to large Kelvin values and vice versa).


hsudbrock:
Hi Chris; can this be merged from your side? If yes, could you approve via the review once more, so that github will let us merge?

PR #297, opened 2018-11-09
Verbose debug logging and status handling of the CoordinatorHandler

triller-telekom:
Without this PR the coordinator reacts immediately on changes from the
underlying zss library. However this leads to correct but problematic status
changes, for example: ONLINE -> OFFLINE -> ONLINE

This happens during initialization where the status from the zss library
reports: UNINITIALISED -> ONLINE -> INILIALISING -> OFFLINE -> ONLINE -> ONLINE

The problem with these changes are that the children of such a bridge handler
react immediately with `bridgeStatusChanged()` on whatever state the bridge
has.

Inside the `bridgeStatusChanged()` a child `ThingHandler` might also want
to initialize the network inside the zss library and this will cause
unexpected results.

### DIFF
--- a/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeCoordinatorHandler.java
+++ b/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeCoordinatorHandler.java
@@ -745,10 +745,7 @@ public abstract class ZigBeeCoordinatorHandler extends BaseBridgeHandler
             case INITIALISING:
                 break;
             case ONLINE:
-                // make sure to not confuse child things
-                if (getThing().getStatus() != ThingStatus.ONLINE) {
-                    updateStatus(ThingStatus.ONLINE);
-                }
+                updateStatus(ThingStatus.ONLINE);
                 break;
             case OFFLINE:
                 updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.COMMUNICATION_ERROR);
--- a/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeCoordinatorHandler.java
+++ b/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeCoordinatorHandler.java
@@ -173,12 +173,15 @@ public abstract class ZigBeeCoordinatorHandler extends BaseBridgeHandler
 
         if (getConfig().get(CONFIGURATION_INITIALIZE) != null) {
             initializeNetwork = (Boolean) getConfig().get(CONFIGURATION_INITIALIZE);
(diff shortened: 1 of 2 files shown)


cdjackson:
I've updated the framework to better handle transitions. It restricts some transitions, and hold off state transitions until the dongle initialisation is complete. It ought to resolve this at source, thus negating the need for this PR.

https://github.com/zsmartsystems/com.zsmartsystems.zigbee/pull/432

If this resolves the issue I'd suggest to revise this to remove the coordinator transitions, and keep the other fixes/debug logging etc.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeCoordinatorHandler.java
```diff
@@ -731,13 +738,17 @@ public ZigBeeEndpoint getEndpoint(IeeeAddress address, int endpointId) {
...
     public void networkStateUpdated(final ZigBeeTransportState state) {
+        logger.debug("{}: networkStateUpdated called with state={}", nodeIeeeAddress, state);
         switch (state) {
             case UNINITIALISED:
                 break;
             case INITIALISING:
                 break;
             case ONLINE:
-                updateStatus(ThingStatus.ONLINE);
+                // make sure to not confuse child things
```
cdjackson: As a matter of interest, why does this "confuse child things"?  Do children get the ```bridgeStatusChanged``` callback? If so, I just wonder if this transition change code shouldn't be put into the framework so that children are only notified of bridge changes?

Or am I missing something else?

triller-telekom: > Do children get the bridgeStatusChanged callback?

Yes I have observed this. My bridge (ember coordinator) was ONLINE and ZSS reported ONLINE again and this has triggered the `bridgeStatusChanged` of the child things.

But I think its fine for ZSS to report ONLINE if it already is ONLINE, because then the user of the ZSS library (in this case the bridgehandler) can decide what to do with this information. I.e. the bridge might have missed a transition to ONLINE before and only get the second call with ONLINE. In such a case it can decide on its own if it accepts the second ONLINE call.

WDYT?

triller-telekom: You are right, this check already exists:

https://github.com/eclipse/smarthome/blob/master/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManagerImpl.java#L231

So, sorry about the confusion. What I had observed was probably related to the oscillation of the bridge status, which you have fixed in ZSS now.

I have removed the check from this PR.

cdjackson: Thanks.

PR #298, opened 2018-11-13
Add a converter for battery alarms from the power configuration profile

### LINKED ISSUE #295: Low battery channel based on Power Configuration cluster
Currently, the channel converter for "low battery" can only be used for devices implementing the IAS Zone cluster. It would be nice to get a low battery channel also for devices implementing the power configuration cluster.

There are already two channel converters using the power configuration cluster, providing a battery-level and a battery-voltage channel, respectively. Those channels are created if the device provides the optional attribute `BatteryPercentageRemaining` or `BatteryVoltage`, respectively.

The converter for "low battery" would rely on other optional attributes of the power configuration cluster, viz the `BatteryAlarmState`, which reports an alarm when the battery level or the battery voltage decreases below one of four thresholds, respectively (`BatteryVoltageMinThreshold`, `BatteryVoltageThreshold{1-3}`, `BatteryPercentageMinThreshold`, and `BatteryPercentageMinThreshold{1-3}`).

hsudbrock:
Resolves #295 by adding a channel converter that provides a battery alarm channel based on the power configuration cluster. This channel takes one out of five values: "No battery alarm was reported" or "The battery level dropped below one of four thresholds defined on the device".

Here's an example how this looks like in Paper UI (channel `Battery Alarm`). (The device also provides the IAS Zone cluster, resulting in the Low Battery channel, and it also provides the battery voltage via the power configuration cluster.)

image

I would be happy about any feedback.

### DIFF
--- a/org.openhab.binding.zigbee/ESH-INF/thing/channels.xml
+++ b/org.openhab.binding.zigbee/ESH-INF/thing/channels.xml
@@ -20,11 +20,11 @@
         <category>Energy</category>
         <state readOnly="true">
           <options>
-            <option value="minThresholdReached">Battery level dropped below minimal threshold</option>
-            <option value="threshold1Reached">Battery level dropped below threshold 1</option>
-            <option value="threshold2Reached">Battery level dropped below threshold 2</option>
-            <option value="threshold3Reached">Battery level dropped below threshold 3</option>
-            <option value="noThresholdReached">No battery alarm</option>
+            <option value="minThreshold">Below min threshold</option>
+            <option value="threshold1">Below threshold 1</option>
+            <option value="threshold2">Below threshold 2</option>
+            <option value="threshold3">Below threshold 3</option>
+            <option value="noThreshold">No battery alarm</option>
           </options>
         </state>
     </channel-type>
--- a/org.openhab.binding.zigbee/ESH-INF/thing/channels.xml
+++ b/org.openhab.binding.zigbee/ESH-INF/thing/channels.xml
@@ -11,6 +11,23 @@
         <category>Energy</category>
         <state pattern="%.1f" readOnly="true" />
     </channel-type>
+    
+    <!-- Battery Alarm -->
+    <channel-type id="battery_alarm">
+        <item-type>String</item-type>
+        <label>Battery Alarm</label>
+        <description>The battery alarm state</description>
+        <category>Energy</category>
+        <state readOnly="true">
+          <options>
+            <option value="minThresholdReached">Battery level dropped below minimal threshold</option>
+            <option value="threshold1Reached">Battery level dropped below threshold 1</option>
+            <option value="threshold2Reached">Battery level dropped below threshold 2</option>
+            <option value="threshold3Reached">Battery level dropped below threshold 3</option>
+            <option value="noThresholdReached">No battery alarm</option>
+          </options>
+        </state>
+    </channel-type>
 
     <!-- Color Channel -->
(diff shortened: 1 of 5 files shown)


org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterBatteryAlarm.java
```diff
@@ -0,0 +1,166 @@
...
+        if (cluster == null) {
+            logger.error("{}: Error opening power configuration cluster", endpoint.getIeeeAddress());
+            return false;
+        }
+
+        try {
+            CommandResult bindResponse = cluster.bind().get();
+            if (bindResponse.isSuccess()) {
+                CommandResult reportingResponse = cluster.setReporting(cluster.getAttribute(ATTR_BATTERYALARMSTATE),
+                        (int) ofMinutes(10).getSeconds(), (int) ofHours(2).getSeconds()).get();
```
triller-telekom: Shouldn't we put these magic numbers, "10" and "2" into constants?

hsudbrock: fine for me, will update

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/ZigBeeBindingConstants.java
```diff
@@ -193,6 +198,13 @@
 
     public static final String FIRMWARE_FAILED = "@text/zigbee.firmware.failed";
 
+    // List of channel state constants
+    public static final String STATE_OPTION_BATTERY_MIN_THRESHOLD_REACHED = "minThresholdReached";
+    public static final String STATE_OPTION_BATTERY_THRESHOLD_1_REACHED = "threshold1Reached";
```
cdjackson: Same comment as above - do these constants needs to be ```STATE_OPTION_BATTERY_MIN_THRESHOLD_REACHED ``` or can they just be ```STATE_OPTION_BATTERY_MIN_THRESHOLD```?

org.openhab.binding.zigbee/ESH-INF/thing/channels.xml
```diff
@@ -11,6 +11,23 @@
...
+    
+    <!-- Battery Alarm -->
+    <channel-type id="battery_alarm">
+        <item-type>String</item-type>
+        <label>Battery Alarm</label>
+        <description>The battery alarm state</description>
+        <category>Energy</category>
+        <state readOnly="true">
+          <options>
+            <option value="minThresholdReached">Battery level dropped below minimal threshold</option>
```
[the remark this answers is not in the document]
hsudbrock: I see your point, Chris. Shortening the option values sounds sensible, I will push a commit changing this. I will also propose a shorter variant of the Strings displayed for the different states.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/ZigBeeBindingConstants.java
hsudbrock: Fine for me, I will change the constant names.

hsudbrock:
Thanks for your quick feedback, Chris! I have pushed a commit addressing your suggestions.

PR #302, opened 2018-11-19
EmberHandler: Check if counters from dongle are available before updating the channels

### LINKED ISSUE #303: NPE during EmberHandler initialization
If openHAB is started without a stick attached, the following call returns an empty map, leading to an NPE afterwards:

```
Map<String, Long> counters = dongle.getCounters();
```

I will provide a PR to check if the map is not empty before updating the channels.

```
2018-11-19 11:24:52.510 [WARN ] [rappedScheduledExecutorService:67   ] - Scheduled runnable ended with an exception: java.lang.NullPointerException: null
        at org.openhab.binding.zigbee.ember.handler.EmberHandler$1.run(EmberHandler.java:107)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
        at java.util.concurrent.FutureTask.runAndReset(FutureTask.java:308)
        at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$301(ScheduledThreadPoolExecutor.java:180)
        at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:294)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        at java.lang.Thread.run(Thread.java:748)
```

### LINKED ISSUE #437: Discovery results from deleted Things reappear in the Inbox
If I delete a Thing which is currently not reachable (maybe because it is battery powered and the battery is currently unplugged), the Thing is removed. But the Device itself did not receive the network leave request and thus is still part of the network.

If one turns on the device again, a node int he `ZigbeeNetworkManager` is created, which leads to a new `DiscoveryResult` put into the Inbox.

I will provide a PR to only create new Inbox results if a `startScan()` is triggered.

triller-telekom:
Fixes #303

### DIFF
--- a/org.openhab.binding.zigbee.ember/src/main/java/org/openhab/binding/zigbee/ember/handler/EmberHandler.java
+++ b/org.openhab.binding.zigbee.ember/src/main/java/org/openhab/binding/zigbee/ember/handler/EmberHandler.java
@@ -103,18 +103,20 @@ public class EmberHandler extends ZigBeeCoordinatorHandler implements FirmwareUp
             public void run() {
                 Map<String, Long> counters = dongle.getCounters();
 
-                updateState(new ChannelUID(getThing().getUID(), UID_ASH_RX_DAT),
-                        new DecimalType(counters.get(ASH_RX_DAT)));
-                updateState(new ChannelUID(getThing().getUID(), UID_ASH_TX_DAT),
-                        new DecimalType(counters.get(ASH_TX_DAT)));
-                updateState(new ChannelUID(getThing().getUID(), UID_ASH_RX_ACK),
-                        new DecimalType(counters.get(ASH_RX_ACK)));
-                updateState(new ChannelUID(getThing().getUID(), UID_ASH_TX_ACK),
-                        new DecimalType(counters.get(ASH_TX_ACK)));
-                updateState(new ChannelUID(getThing().getUID(), UID_ASH_RX_NAK),
-                        new DecimalType(counters.get(ASH_RX_NAK)));
-                updateState(new ChannelUID(getThing().getUID(), UID_ASH_TX_NAK),
-                        new DecimalType(counters.get(ASH_TX_NAK)));
+                if (!counters.isEmpty()) {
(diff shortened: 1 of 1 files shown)


triller-telekom:
Thank you for pointing me to the mistake that I did with using the wrong repo and thus the wrong issue number, I have fixed it now.

cdjackson:
While adding this protection is probably ok, maybe it's worth not starting the thread that gathers the data if the dongle is not opened correctly? ie there's no point in gathering statistics if the port isn't open.

triller-telekom:
@cdjackson I agree. But I guess this needs a larger PR to address this, because the port opening is done asynchronously by the `startZigBee` call. But we should put it on the TODO list. Shall I create a new issue for that?

Also the scheduling should be done via `scheduler.scheduleWithFixedDelay`, see http://www.eclipse.org/smarthome/documentation/development/guidelines.html#d-runtime-behavior

cdjackson:
Ok, I'll merge this then, and let's create a separate issue to cover the other two points. Thanks.

cdjackson:
@kaikreuzer it seems that Jenkins is still being called (or is still a required check) that blocks merging. If I understood our conversation on this yesterday, you would expect that this should not be the case for newly submitted PRs?

kaikreuzer:
You should mention that before you merge the PR, now I cannot see this check anymore.

Looking now at the details, it only shows me the one expected successful check: 

screenshot 2018-11-19 at 14 09 59

cdjackson:
Ok, sorry, it was still showing Jenkins as waiting.

cdjackson:
It's the same as we saw yesterday, and the same as you can still see on the other open PRs on this repo.

kaikreuzer:
Ah, got it! It was a setting in the project configuration about merge checks - I've not set the correct check in there:

screenshot 2018-11-19 at 14 14 21

cdjackson:
Thanks.

PR #311, opened 2018-11-27
Channel converter for the IAS tamper attribute

### LINKED ISSUE #309: New feature: Channel converter for the IAS tamper attribute
We want to introduce a channel which reports that a device has bee tampered.

For the implementation we would
- add a new channel type with the id "ias_tampering_switch"
- create a channel converter which extends the ZigBeeConverterIas

TomTravaglino:
Resolves #309


cdjackson:
@TomTravaglino this needs to be rebased against the latest master - otherwise looks fine to me.

org.openhab.binding.zigbee/ESH-INF/thing/channels.xml
```diff
@@ -146,6 +146,15 @@
         <state readOnly="true" />
     </channel-type>
 
+    <!-- IAS Tampering Switch Channel -->
```
hsudbrock: Minor comment: To align with other comments, write `IAS Tamper Channel` instead of `IAS Tampering Switch Channel`.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterIas.java
```diff
@@ -109,6 +103,21 @@ protected boolean supportsIasChannel(ZigBeeEndpoint endpoint, ZoneTypeEnum requi
         return zoneType == requiredZoneType;
     }
 
+    protected boolean hasIasZoneInputCluster(ZigBeeEndpoint endpoint) {
+        if (endpoint.getInputCluster(ZclIasZoneCluster.CLUSTER_ID) == null) {
```
hsudbrock: I don't quite see the difference between the two `if` statements in this method - don't they just test the same? Or do I overlook something here?

(I know that this code was copied from the battery converter - but nevertheless think that this is a valid question :) )

TomTravaglino: I took a closer look to it and I agree that both statements do the same.
I can delete the second statement and then have this method body
``` 
        if (endpoint.getInputCluster(ZclIasZoneCluster.CLUSTER_ID) == null) {
            logger.trace("{}: IAS zone cluster not found", endpoint.getIeeeAddress());
            return false;
        }

        return true;
```

hsudbrock: Sounds good :+1:

TomTravaglino:
I addressed  the comments and suggestions and I rebased the branch against master. Please take a look.

org.openhab.binding.zigbee/ESH-INF/thing/channels.xml
```diff
@@ -146,6 +146,15 @@
         <state readOnly="true" />
     </channel-type>
 
+    <!-- IAS Tamper Channel -->
+    <channel-type id="ias_tamper">
+        <item-type>Switch</item-type>
+        <label>Tamper</label>
+        <description></description>
```
hsudbrock: Would it be sensible to add a description, like for most other channel types in this file?

org.openhab.binding.zigbee/ESH-INF/thing/channels.xml
```diff
@@ -146,6 +146,15 @@
...
+        <category></category>
```
hsudbrock: Would `Alarm` be a suitable category here? (From the bottom of https://www.eclipse.org/smarthome/documentation/concepts/categories.html).

Do you have any opinion here, @cdjackson ?

org.openhab.binding.zigbee/ESH-INF/thing/channels.xml
TomTravaglino: Could add `Indicates if a device is tampered`

hsudbrock: Small improvement: `Indicates if a device is tampered with`

TomTravaglino:
I updated the PR. Now the category of channel type `ias_tamper` is `Alarm`. And there is a small test for the converter.

TomTravaglino:
Could someone please do the review? ;-)

cdjackson:
@TomTravaglino please can you provide an update to the README for this channel?

Issue #316, opened 2018-11-30
Wrong command send to level_control channel


Hambootje:
I want a switch and a dimmer for manual control of my lightbulb. If I link the level_control channel to both a switch and a dimmer item, the bulb turn on to 100% level by the switch.
The wanted behaviour is that the bulb is switched on to the last level before is was switched off.

I have defined 2 items as shown below, both linked to the Levelcontrol channel of a tradfri lightbulb.
The switch sends ON/OFF, the dimmer 0-100.
The ON command from the switch sets the lightbulb to 100% level. However, this is not represented by the dimmer item.

`Switch Werkkamer_Lamp      "Werkkamer Lamp [%s]" [ "Lighting" ]`    
`Dimmer Werkkamer_Lamp_dim    "Werkkamer Dim [%d %%]" [ "Lighting" ]`

According Chris: 

> This should not be the case. If I remember correctly, when the OnOff command is sent to a level_control channel, it should send the same command as when it’s sent to the switch channel.

> I checked, and you are right that different commands are used. Please raise an issue for this and I’ll have a look at it when I get a chance.

### Configuration
Coordinator | Ember EM35x or CC2531EMK
openHAB version | 2.4.0.SNAPSHOT
Hardware | Raspberry Pi 3+
Java version | openjdk version "1.8.0_152"
Devices | zigbee:device:013C1ADE:000b57fffe909e96:000B57FFFE909E96_1_dimmer

cdjackson:
This should now be resolved. If there's any issues, please reopen or open another issue.

PR #318, opened 2018-12-01
Handle OnOff command in level converter same as switch converter

cdjackson:
As per #316 when we receive an OnOffType command in the ```SwitchLevel``` Converter, we should perform the same function as if it were received in the ```SwitchOnoff``` Converter. The current implementation always sets the level to 100% which is not correct in many instances.

@puzzle-star any comments.


puzzle-star:
Hi @cdjackson,

I have a first question here.

Is the channel consolidation still taking place? I mean, if the Dimmer channel is detected, is the Switch channel deleted? I did not see where this went out, but I now see the two channels for my bulbs. The reason why I ask is because if channel consolidation is the expected behaviour, we should not have both channels detected, but if both are to be seen, the behaviour requested in https://github.com/openhab/org.openhab.binding.zigbee/issues/316 would be achieved by using the Switch channel instead of linking two items of different type to the same channel.

Anyway, the topic here is whether sending ON to a dimmer should mean:
1. Turn on to your previous state
2. Turn on to full state

In general, I tend to think sending OnOffType.ON to a Dimmer in ESH should mean the same as sending PercentType.HUNDRED, as this is the conversion naturally present in ESH.

So when I send ON to my "Everything" group composed of lights and dimmers, I would expect everything to go to ON at full brightness.

But this is exactly the opposite use case as the one requested in the issue this PR tries to address.

Maybe this behaviour should be made configurable? (unless channel consolidation does not take place, in which case this is just achieved by using the Switch channel directly)


Pedro

cdjackson:
> I have a first question here.
> 
> 

Channel consolidation should still be occurring unless it was removed in the channel factory change. We discussed this previously, and it should still be there (I have not checked).  It is “required” for ESH concepts not to have both channels.

> Anyway, the topic here is whether sending ON to a dimmer should mean:
> 
> Turn on to your previous state
> Turn on to full state
> In general, I tend to think sending OnOffType.ON to a Dimmer in ESH should mean the same as sending PercentType.HUNDRED, as this is the conversion naturally present in ESH.
> 
> 

I’m not sure I would agree - dimmer.100% should be 100%, but sending OnOff I think should send the OnOff - the same as the OnOff channel (IMHO). This gives you both functions - the same as we would have with separate channels, so the user can decide what they want.
> So when I send ON to my "Everything" group composed of lights and dimmers, I would expect everything to go to ON at full brightness.
> 
> 

But you can still achieve this - just send the 100% to the group instead.

> Maybe this behaviour should be made configurable? (unless channel consolidation does not take place, in which case this is just achieved by using the Switch channel directly)
> 
This is possible, although I would probably prefer to avoid making it configurable if we can avoid it. Why can’t you just send the 100% command to your group if you want it to be on at 100%?

puzzle-star:
> Channel consolidation should still be occurring unless it was removed in the channel factory change. We discussed this previously, and it should still be there (I have not checked). It is “required” for ESH concepts not to have both channels.

Well, all my lights have all detected channels available on my OH (i.e. color bulbs have color, dimmer, and switch). Maybe if these were detected at some moment they will keep on appearing? But anyway I went to a clean installation of M5 a couple of weeks ago on a new device and all channels for totally new things still appeared.

> I’m not sure I would agree - dimmer.100% should be 100%, but sending OnOff I think should send the OnOff - the same as the OnOff channel (IMHO). This gives you both functions - the same as we would have with separate channels, so the user can decide what they want.

I wonder whether ESH concepts wouldn't expect an ON to a Dimmer to mean 100% in general. Is there some place where this could be checked?

> But you can still achieve this - just send the 100% to the group instead.

Nothing materially wrong here :-)  I was thinking of the same approach while answering. I just wonder what is the expected behaviour in ESH.

> This is possible, although I would probably prefer to avoid making it configurable if we can avoid it. Why can’t you just send the 100% command to your group if you want it to be on at 100%?

If I recall correctly, the original implementation of the Dimmer channel, was to send a "moveToLevelWithOnOffCommand" with a 100% for ON commands, so I undertand it has been the expected (or at least historical) behaviour by everybody.

But as you say, this can be also achieved by sending 100% instead of ON. Just a lots of rules and sitemap related things to be changed in my case, and wondering how many people would be on the same situation.

As far as there is a common approach, and it is documented, I am sure both possibilities are OK. Otherwise, 50% of the users thinking the "natural way" is one, will not expect the "other way" happening. :-) and I would then think that making this configurable might make sense.

Anyway, looking at the PR, and if you decide the new behaviour is the right one, I think that in its current implementation it sends commands twice. The first "if" branch decided what to send in terms of Level and the following "if" branch decided what to send in terms of command, but in the PR the first branch is already sending commands.

Pedro

cdjackson:
> Well, all my lights have all detected channels available on my OH

Ok, so it sounds like we also lost the channel consolidation in the latest update :( 

> I wonder whether ESH concepts wouldn't expect an ON to a Dimmer to mean 100% in general.

Good question - I'm not sure if this is defined. My guess is this won't be defined, although I personally would like to see this sort of thing defined so that all bindings try to provide a consistent interface. 

> Anyway, looking at the PR, and if you decide the new behaviour is the right one, I think that in its current implementation it sends commands twice.

Thanks - I did have a return in there, but it seems to have been lost at some point.

@tomhoefer, @hsudbrock any preference from your side? I personally don't mind too much either way, although I think that this PR provides better consistency between the channels, and provides more functionality in that it is at least possible to restore the light to its previous value which is not currently possible.

cdjackson:
> Anyway, looking at the PR, and if you decide the new behaviour is the right one, I think that in its current implementation it sends commands twice. The first "if" branch decided what to send in terms of Level and the following "if" branch decided what to send in terms of command, but in the PR the first branch is already sending commands.

Maybe I misunderstand your point - as above, I thought there was a return in the first ```if``` branch, and there is on line 148 -:

```
            if (clusterOnOffServer != null) {
                // If we support the OnOff cluster, and this is an OnOffType,
                // then we should perform the same function as the SwitchOnoffConverter
                if (cmdOnOff == OnOffType.ON) {
                    clusterOnOffServer.onCommand();
                } else {
                    clusterOnOffServer.offCommand();
                }
                return;
            } else {
                if (cmdOnOff == OnOffType.ON) {
                    percent = PercentType.HUNDRED;
                } else {
                    percent = PercentType.ZERO;
                }
            }
```

Is this what you meant by the first branch, or am I missing something?

puzzle-star:
> Thanks - I did have a return in there, but it seems to have been lost at some point.

My mistake, the "return" is there... :-)

> Ok, so it sounds like we also lost the channel consolidation in the latest update :(

I reverted the commit for the chanel factory in my installation to make it work, and the channels are still there, so there must be something else. I will check...

Just if the PR goes through, I would suggest the new behaviour being documented somewhere, as it changes the previous one, and then also apply a similar approach fro the color channel converter

jfrank1845:
> 
> 
> @hsudbrock any thoughts on this? I would like to get this merged in the coming days as there will soon be a release of OH2.4.

If it matters any what users think - I'd like to second this since we are losing our switch channel in favor of the consolidation.  Its a lot easier to just say "100" to force a light to full brightness rather than try to keep track of previous percentages to restore previous state.

cdjackson:
> If it matters any what users think

Sure, it matters - thanks for posting.

hsudbrock:
Sorry, Chris, I missed this (was incapacitated earlier this week :/). I'll take a look.

hsudbrock:
I had a look, Chris, and I think that this is a sensible change.

cdjackson:
Thanks Henning.

PR #319, opened 2018-12-01
Improved handling of OTA initialisation

### LINKED ISSUE #312: Device FW update fails due to incorrect determination of OTA server, cluster and endpoint
I just tried to update the device firmware of a thing and the firmware update failed repeatedly using the following devices:
- Bitron Video AV2010/22A [V14 -> V15]
- CentraLite 3328-D [V18005310 -> V1F0B5310]
- BitronHome 902010/25 [V30 -> V40]

When checking the logs, I see that the firmware image transmission has not been even started and the binding claims that the OTA server could not be created for the respective device:
> [symbolicName=org.openhab.binding.zigbee, className=org.openhab.binding.zigbee.handler.ZigBeeThingHandler] - 000D6F000BBCA4C7: Can't create OTA server

I believe there is a simple bug in the  `ZigBeeThingHandler#updateFirmware` how the OTA server, endpoint and cluster are determined since it relies on a specific ordering of the endpoints (e.g. if the first endpoint provides the OTA server, the OTA cluster and the OTA endpoint will never be initialized). 

@cdjackson What do you think? Did I miss something? If you want, I can contribute a fix for that.

cdjackson:
@hanneshofmann this ought to resolve the issue with the OTA initialisation. I've not tested it as it will fail compilation at the moment until the libraries are updated.
Closes #312 


cdjackson:
Thanks. The timeouts _may_ be device issues (some devices don't respond to the final message in the transfer), but as mentioned, raise an issue and provide the logs and I'll take a look.

cdjackson:
Merging as already reviewed above by @hanneshofmann

PR #324, opened 2018-12-02
Fix null factory being passed to new ZigBeeThingHandler (issue #320)

puzzle-star:
Create a default ZigbeeChannelConverterFactory if one is not set before creating a ZigBeeThingHandler


cdjackson:
@puzzle-star I think it's been agreed that this is not the correct way to implement this as the factory should be injected through the handler, and OSGi. I suggest to close if you agree?

puzzle-star:
Agreed

Issue #322, opened 2018-12-02
Network state file gets easily lost when a power fail or crash happens


puzzle-star:
Network state file is not written atomically, and is fully rewritten quite often, so if a power fail or a crash occurs, the network state is easily lost, forcing a full network rediscovery.

The related PR is "Atomic network state file update (issue #322) (PR #323)"

puzzle-star:
Hopefully fixed the PR. Only the needed commit on it

puzzle-star:
Hi @cdjackson,

Any chance to merge this before M8? It should add some stability in case of crash, as rediscovery when the file is lost is painful, specially on big networks.


Best Regards, 
Pedro

cdjackson:
I don’t think there will be an M8, but I will try and get it merged ASAP - I’ll look at it in the next couple of days.

cdjackson:
I should add that really, if you are having issues with power, you should solve that problem. Any system that has power removed frequently, as you seem to indicate is happening on your system will remain unreliable.

puzzle-star:
Agreed, but RasbPi are quite "sensitive" some times, and systems with low memory may end up also in killing the Java process at the worst moment...

So this is an enhancement that would provide some safety against this, hopefully at a very low cost.

The file is quite big and i.e. in my network I see it written several times in a raw, so anything going wrong at that time will garble it. Writing to a temp file and renaming (the best way and most atomically way possible) is more a fail safe operation in this case

puzzle-star:
Hi @cdjackson,

Well, I have done some tests under Linux and Windows, and there seems not to be a portable way to do the rename / move atomically in a safe way without a bit of coding, considering the target file might be open (i.e. by a backup).

java.nio.Files.move(orig, dest, ATOMIC_MOVE) will do it correctly (and atomically!) both under Linux and Windows even if the target file exists, but will fail under Windows (not Linux) when the target file is open (curiously enough, it does not fail when the original file is open instead...).

So the only way I found to do this in a portable way, considering the above, is making sure the target file of a rename does not exist if we want the operation to work even if it is open (i.e., by renaming the target file itself to a new unique name beforehand).

Would if be OK for you if I implement it like this?

File save:
- Write the new config to a temporary file ([_name_].xml.new)
- Rename the old config file to a backup unique file name ([_name_].xml -> [_name_].xml.[_timestamp_])
[...]

PR #327, opened 2018-12-03
Ability to pause writing/reading of network state file for backups

triller-telekom:

### DIFF
--- a/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeCoordinatorHandler.java
+++ b/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeCoordinatorHandler.java
@@ -890,4 +890,19 @@ public abstract class ZigBeeCoordinatorHandler extends BaseBridgeHandler
     public @NonNull ThingUID getUID() {
         return getThing().getUID();
     }
+
+    /**
+     * Pauses serialization so the network state xml file can be backed up and restores safely
+     */
+    public void pauseSerialization() {
+        networkStateSerializer.pauseSerialization();
+    }
+
+    /**
+     * Continues serialization of the network state xml file after it has been paused before
+     */
+    public void continueSerialization() {
+        networkStateSerializer.continueSerialization();
+    }
+
 }
--- a/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/ZigBeeNetworkStateSerializerImpl.java
+++ b/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/ZigBeeNetworkStateSerializerImpl.java
@@ -68,6 +68,8 @@ public class ZigBeeNetworkStateSerializerImpl implements ZigBeeNetworkStateSeria
      */
     private final String networkStateFilePath;
 
+    private boolean serializationPaused = false;
+
     public ZigBeeNetworkStateSerializerImpl(String networkId) {
         this.networkId = networkId;
         networkStateFilePath = ConfigConstants.getUserDataFolder() + "/" + ZigBeeBindingConstants.BINDING_ID;
@@ -109,6 +111,10 @@ public class ZigBeeNetworkStateSerializerImpl implements ZigBeeNetworkStateSeria
      */
     @Override
     public synchronized void serialize(final ZigBeeNetworkManager networkManager) {
+        if (serializationPaused) {
+            return;
+        }
+
         XStream stream = openStream();
 
         logger.debug("Saving ZigBee network state: Start.");
@@ -141,6 +147,9 @@ public class ZigBeeNetworkStateSerializerImpl implements ZigBeeNetworkStateSeria
      */
     @Override
     public void deserialize(final ZigBeeNetworkManager networkState) {
+        if (serializationPaused) {
+            return;
+        }
         logger.debug("Loading ZigBee network state: Start.");
 
         final File file = getNetworkStateFile();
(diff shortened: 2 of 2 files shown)


triller-telekom:
@cdjackson I would like to hear your opinion on this proposal.

This fix would make it possible to backup or restore the xml files while the coordinatorhandler keeps running, i..e. we do not need to dispose and re-initialize it.

Since this is only in the binding and only in the implementation and not in the interface, I think it doesn't harm anything.

cdjackson:
I don't have any real problem with this. My only question is if it might be worth setting a flag if ```serialize``` is called when serialisation is paused, and then perform the serialisation when ```continueSerialization``` is called? It probably doesn't matter too much since the network is probably serialised regularly enough, but this might still be a good idea to ensure everything is consistent...

triller-telekom:
> It probably doesn't matter too much since the network is probably serialised regularly enough

That was my assumption too.

But nevertheless I added that flag now. Since the `serialize` method needs the `Networkmanger` as a parameter, I am using this `Networkmanger` as a flag, i.e. if it is set, there was a call while paused and if it is `null` there was none.

If there was a call to `serialize` while paused, I am executing it once on resume.

Is this what you imagined?

puzzle-star:
Hi @cdjackson and @triller-telekom,

If the purpose is to be able to backup the file, and the issue is that it cannot be done while it is being written because its contents would be partial, then the fix for issue #322 will solve this in a more straightaway fashion, as with it the file is atomically written, so it should be always on a backup-able state.

Pedro

cdjackson:
@puzzle-star I don't think that #322 will resolve the issue here.  Let's assume for a second that the rename in #322 was an atomic operation (we can discuss that question in the other PR!) - since the copy (for the backup) required here is not, there will still be a problem.

Thread 1 - opens the file for backup and starts to read
Thread 2 - tries to rename the file

One of these two operations will fail - probably the copy/rename for the serialisation since I suspect that this will not work if the file is already open.

I know we are talking rare occasions here, but given that the power failure case you discuss in #322 should already be very rare, we trade one unlikely situation for another, but it doesn't actually solve anything.

cdjackson:
@triller-telekom rethinking this - are you sure this is what you want to do? I thought from @hsudbrock that you were going to write a backup provider that was not simply copying this file, but actually interacted with the ```ZigBeeNetworkManager```. You should note that it is not possible to fully back / restore up the network using this file, and that is not its purpose (although it could probably be changed).

puzzle-star:
Hi @cdjackson,

I tend to think of Linux/*nix...

Thanks case you describe above would work OK on Linux, as the rename is truly atomic, and removing a file is perfectly ok even if open (just the directory entry is removed, but the open fds still point to the right inode, which will persist till the file is closed by the las process)

But I will check on Windows

Also Java.nio.Files should provide platform independent way to for atomic moves / renamed (can it be usedcin the binding, or only Java.io.File?)

cdjackson:
In some documentation I find on the web, I see 

> The renameTo() method returns boolean (true or false), indicating whether the renaming was successful. Renaming of moving a file may fail for various reasons, like the file being open, wrong file permissions etc.

I'm just wary that we probably can't guarantee this works on every system, with every JVM implementation.

puzzle-star:
That's why I added the check, that deleted the file if the rename fails and then retries.

On OSes implementations where the the rename would fail, the reason is normally, even if I cannot imagine all scenarios, because if the target tfuke existing or being open (discarding permission issues, as the it would not work even now)

If java.nio.Files is an option, it should provide a platform independent way to do it. Will check later.

Keeping a trail of backups, as you say, would also work. Just take into account then that the file is updated quite often (I have seen > 30 updates in a couple of minutes, when the network is rediscovered).

cdjackson:
> Keeping a trail of backups, as you say, would also work. Just take into account then that the file is updated quite often (I have seen > 30 updates in a couple of minutes, when the network is rediscovered).

Why does this matter?

puzzle-star:
Because if we keep a limited number of backups (i.e. the last 5 ones), all of them will be in the last minute normally. If the purpose is to keep some history of backups, this won't do it. If not, just keeping the last one will do.

cdjackson:
The purpose is not to keep a history - the purpose is to ensure reliability of data.

Can I suggest that we don't discuss this further here - it's not the right place. Thanks.

puzzle-star:
Ok, then

puzzle-star:
PS: which is the right place? :)

cdjackson:
> PS: which is the right place? :)

Your PR - not this one. Your PR is looking at the strategy for serialisation which is what we are discussing, but it's not the right PR here.

triller-telekom:
Alright, back to the topic :)

For sure i could serialize the network state within a backupprovider. But for this I need access to the coordinator to access its NetworkManager.

Yes, I need that access anyway for the trust center keys from the stick, but I would like to avoid it whereever possible. Also, for now its easier to just copy and restore the network file in case of a plain backup/restore without changing the usb stick.

So my question now is: Does it do any harm to the codebase if we allow for pausing and resuming the serialization? If not, I would like to see this PR merged.

What's your opinion on this?

cdjackson:
No - I don't generally mind merging this, but on the other hand I would prefer not to add code if it doesn't actually do what is needed ;)

I'm in two minds though - you are having develop a backup provider, so why not just get access to the coordinator thing, which I think is not too hard (?? - I think Henning does it in the console already) and read out the state periodically? This is clearly a better solution as it allows you to generate all the data that you need - otherwise you will need to generate two backup solutions which seems a bit strange. I'm pretty sure that is what Henning was going to do when I discussed this with him a while back.

So - why not do that instead?

Again, I don't really mind merging this, but as I think this is only a Qivicon request as there is no plan in ESH or OH to have such a backup provider, I would prefer not to add code that no-one uses.

cdjackson:
@triller-telekom and @hsudbrock do we still need this PR, or can Qivicon do the backup through other means? Henning - I recall we discussed this some time ago and you were going to directly access the ```ZigBeeNetworkManager``` to perform the backup/restore. Given that it is not possible to get all information you need from the XML, and will therefore have to implement another approach, doesn't it make sense to just have a single backup/restore rather than two different mechanisms?

hsudbrock:
@cdjackson I am not sure about @triller-telekom 's status here - let's wait until he's back, then I can discuss with him.

triller-telekom:
We do not need this PR anymore. I therefore close it.

PR #329, opened 2018-12-05
Fix color temperature from/to percent conversion in color temperature converter

### LINKED ISSUE #313: Color tempterature converter uses cold for high values, warm for low values (but should be vice versa)
In ESH, for the color temperature channel '0% should be the coldest setting (highest Kelvin value)' (see (cf. https://www.eclipse.org/smarthome/documentation/development/bindings/thing-definition.html)).

In the ZigBee binding, this is reversed (0% is the lowest Kelvin value, 100% is the highest Kelvin value).

Maybe this was introduced by my PR https://github.com/openhab/org.openhab.binding.zigbee/pull/287 (for some bulbs - as the goal of that PR was to align how the converter behaves for bulbs that provide min/max Mired values, and for bulbs that do not provide min/max Mired values).

I plan to prepare a fix for this soon. If I overlook a reason why this should be the other way round for this binding, I would be happy about a ping, otherwise I will prepare a (probably small) PR to change this.

hsudbrock:
Fixes #313 by adapting the conversion from Kelvin to percent and vice versa, such that warm color temperatures correspond to high percent values, and cold color temperatures correspond to low percent values.

Also replaces the binding-specific channel type for color temperature with the system channel type, and adds endpoint ids to some log messages.

### DIFF
--- a/org.openhab.binding.zigbee/ESH-INF/thing/channels.xml
+++ b/org.openhab.binding.zigbee/ESH-INF/thing/channels.xml
@@ -39,15 +39,6 @@
         <category>ColorLight</category>
     </channel-type>
 
-    <!-- Color Temperature Channel -->
-    <channel-type id="color_temperature">
-        <item-type>Dimmer</item-type>
-        <label>Color Temperature</label>
-        <description>The color temperature channel allows to set the color
-            temperature of a light from 0 (cold) to 100 (warm).</description>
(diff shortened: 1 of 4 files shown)


hsudbrock:
@cdjackson As proposed by you, I also switched to the system channel for color temperature with this PR.

PR #332, opened 2018-12-08
Update binding to latest ZSmart Systems library

cdjackson:
This PR is in preparation of an update to the ZSS libraries. There are two main changes - the ```bind()``` method was removed from the ```ZclCluster``` class, and the mesh updater was changed to be a standard extension and needs to be initialised differently.

This will fail checks and will be merged immediately after the libraries are updated in the TP.


kaikreuzer:
> will be merged immediately after the libraries are updated in the TP.

I.e. right now?

cdjackson:
I was just doing a rebuild - I wasn’t sure if everything has flowed through yet?

PR #336, opened 2018-12-13
Recognize color mode of color bulb

### LINKED ISSUE #326: Recognize color mode of color bulb
This issue is related to the PR https://github.com/eclipse/smarthome/pull/5762

We want to recognize if the color of the bulb has been set via color scheme or via color temperature scheme. In the Hue binding this has been solved by setting the color temperature to `UnDefType.NULL` when the color mode is not 'color temperature' (see linked PR). We would like to do the same in the ZigBee binding.

Here the idea: In the method `ZigBeeConverterColorTemperature.attributeUpdated(ZclAttribute attribute)` when `attribute` is `ColorMode` and its value is not '0x02' (ColorTemperatureMireds) then we would set the channel status to `UnDefType.NULL`. The additional code could look like this 
```
        if (attribute.getCluster() == ZclClusterType.COLOR_CONTROL
                && attribute.getId() == ZclColorControlCluster.ATTR_COLORMODE) {
            
            Integer colorMode = (Integer) attribute.getLastValue();
            if (colorMode != ColorModeEnum.COLORTEMPERATURE.getKey()) {
                updateChannelState(UnDefType.NULL);
            }
        }
```        
But I faced an issue when I investigated this approach. When you switch from color to color temperature and vice versa then `ColorMode` is not updated automatically (I checked it in the console) and so the listener is not called. I tried then to read `ColorMode` with a console command and in that moment its value was updated and the listener was called.
How can we solve this issue? Is it an option to (force-)update `ColorMode` by calling `com.zsmartsystems.zigbee.zcl.ZclCluster.read(int)` when either HSB or the color temperature changes?

TomTravaglino:
Resolves #326

This is an approach to recognize if the color has been set via color scheme or via color temperature scheme. It corresponds to the solution realized in https://github.com/eclipse/smarthome/pull/5762 which sets the color temperature to UnDefType.NULL when the color mode is not "color temperature".

Your feedback is welcome.

### DIFF
--- a/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterColorColor.java
+++ b/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterColorColor.java
@@ -26,6 +26,7 @@ import org.eclipse.smarthome.core.thing.Channel;
 import org.eclipse.smarthome.core.thing.ThingUID;
 import org.eclipse.smarthome.core.thing.binding.builder.ChannelBuilder;
 import org.eclipse.smarthome.core.types.Command;
+import org.eclipse.smarthome.core.types.UnDefType;
 import org.openhab.binding.zigbee.ZigBeeBindingConstants;
 import org.openhab.binding.zigbee.converter.ZigBeeBaseChannelConverter;
 import org.openhab.binding.zigbee.internal.converter.config.ZclLevelControlConfig;
@@ -40,6 +41,7 @@ import com.zsmartsystems.zigbee.zcl.clusters.ZclColorControlCluster;
 import com.zsmartsystems.zigbee.zcl.clusters.ZclLevelControlCluster;
 import com.zsmartsystems.zigbee.zcl.clusters.ZclOnOffCluster;
 import com.zsmartsystems.zigbee.zcl.clusters.colorcontrol.ColorCapabilitiesEnum;
+import com.zsmartsystems.zigbee.zcl.clusters.colorcontrol.ColorModeEnum;
 
 /**
  * Converter for color control. Uses the {@link ZclColorControlCluster}, and may also use the
@@ -183,6 +185,9 @@ public class ZigBeeConverterColorColor extends ZigBeeBaseChannelConverter implem
                 logger.debug("{}: Exception configuring on/off reporting", endpoint.getIeeeAddress(), e);
             }
         }
+
+        ZclAttribute colorModeAttribute = clusterColorControl.getAttribute(ZclColorControlCluster.ATTR_COLORMODE);
+        clusterColorControl.setReporting(colorModeAttribute, 1, REPORTING_PERIOD_DEFAULT_MAX, 1);
 
         // Create a configuration handler and get the available options
         configLevelControl = new ZclLevelControlConfig();
@@ -422,7 +427,7 @@ public class ZigBeeConverterColorColor extends ZigBeeBaseChannelConverter implem
         HSBType newHSB = new HSBType(oldHSB.getHue(), oldHSB.getSaturation(), brightness);
         currentHSB = newHSB;
         lastBrightness = brightness;
-        if (currentState.get()) {
+        if (currentState.get() && !isCurrentColorModeTemperature()) {
             updateChannelState(newHSB);
         }
     }
@@ -432,7 +437,7 @@ public class ZigBeeConverterColorColor extends ZigBeeBaseChannelConverter implem
         HSBType oldHSB = currentHSB;
         HSBType newHSB = new HSBType(hue, saturation, oldHSB.getBrightness());
         currentHSB = newHSB;
-        if (currentState.get()) {
+        if (currentState.get() && !isCurrentColorModeTemperature()) {
             updateChannelState(newHSB);
         }
     }
@@ -458,6 +463,12 @@ public class ZigBeeConverterColorColor extends ZigBeeBaseChannelConverter implem
         updateColorXY(x, y);
         xChanged = false;
         yChanged = false;
+    }
+
+    private boolean isCurrentColorModeTemperature() {
+        ZclAttribute colorModeAttribute = clusterColorControl.getAttribute(ZclColorControlCluster.ATTR_COLORMODE);
+        Integer colorMode = (Integer) colorModeAttribute.getLastValue();
+        return ColorModeEnum.COLORTEMPERATURE == ColorModeEnum.getByValue(colorMode);
     }
 
     @Override
@@ -511,6 +522,11 @@ public class ZigBeeConverterColorColor extends ZigBeeBaseChannelConverter implem
                         if (Math.abs(y - lastY) < .0000001) {
                             lastY = y;
                             yChanged = true;
+                        }
+                    } else if (attribute.getId() == ZclColorControlCluster.ATTR_COLORMODE) {
+                        Integer colorMode = (Integer) attribute.getLastValue();
+                        if (ColorModeEnum.getByValue(colorMode) == ColorModeEnum.COLORTEMPERATURE) {
+                            updateChannelState(UnDefType.UNDEF);
                         }
                     }
                 }
--- a/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterColorColor.java
+++ b/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterColorColor.java
@@ -77,6 +77,8 @@ public class ZigBeeConverterColorColor extends ZigBeeBaseChannelConverter implem
     private boolean xChanged = false;
     private boolean yChanged = false;
 
+    private ColorModeEnum lastColorMode;
+
     private final AtomicBoolean currentState = new AtomicBoolean(true);
 
     private ZclLevelControlConfig configLevelControl;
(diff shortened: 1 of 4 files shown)


TomTravaglino:
> One question. This PR sets the color temperature channel to UNDEF if the color mode is not in color temp. Why doesn't it set the color channel to undef if the mode IS color temp? If we're saying they are exclusive, then shouldn't both channels be consistent?

I've implemented it to behave like the Hue binding. [Here](https://github.com/eclipse/smarthome/blob/303ba1c1c22f4cd8bdb9b068735373c29c43de06/extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/handler/HueLightHandler.java#L420) you can see that there only the color temperature is set to UNDEF but never the color.

cdjackson:
Ok, but why shouldn’t it be the other way around as well?  Just because one binding does something, doesn’t (necessarily) mean that it’s right?  Maybe there’s a good reason that it’s only done on the color temperature. Maybe it’s considered that color is still valid even when in color temp mode? Or maybe it was just an oversight when the Hue change was made and it should also be done for color?

TomTravaglino:
@kaikreuzer and @triller-telekom did the review of the [Hue PR](https://github.com/eclipse/smarthome/pull/5762). Maybe they know why in the Hue binding the color temperature is set to UNDEF when the color mode is not in color temp but not the other way around.

triller-telekom:
> Maybe it’s considered that color is still valid even when in color temp mode?

That's exactly the reason behind it: See https://github.com/eclipse/smarthome/pull/5762#issuecomment-401268937 where Kai says:

> I don't think that this [setting the color_channel to NULL] makes sense, because the color channel can always have a valid state and you always want to be able to display it in a color picker (e.g. if the user switches from slider to color picker).

So the way @TomTravaglino implemented it is correct.

Does that make sense to you @cdjackson ?

cdjackson:
To be honest, it doesn’t.

With at least one of my bulbs here, I can read back a color that is in no way related to color temperature. So while it might be correct that you _could_ display color temperature in a color box, it does not follow that the color you are displaying will in any way be related to the color temperature if the bulb reports something different when it is in a different color mode. Based on what I see here in at least one bulb I just tested, this will in fact be the case.

Maybe in the Hue bridge this is accounted for to align color and color temperature, but that is not necessarily correct when reading the bulbs directly.

triller-telekom:
I had to read your answer quite a few times, so let me sum up what I understand from it:

You are saying that if you set a colour**temperature** and keep the colour (i.e. HSB or x/y) value, then the zigbee bulb that you have reports **two** colours: One in a colour**temperature** (the newly set one) and the colour one (old value which was previously set).

Then you have the problem that you have to update the OH channel and don't know which value to pick. Is that correct?

In this case I would agree with you that in the zigbee binding we should set the colour value to `NULL` once we set a colour**temperature**, in addition to setting the temperature `NULL` while setting a colour value.

cdjackson:
Sorry for the bad explanation.

Yes - basically you're right with your understanding. The color temperature is not handled as a color - it's a different mode. The color and colortemperature are independent (at least in this bulb - I've not tried others yet) - and the mode is what differentiates them.

It's the same (or similar) with the way that Level control works (at least in some devices) - a device can report OFF, but report the Level to be 50% - that doesn't mean it is on - just that if you change the On/Off mode to On, it will come on at 50%.

I hope that makes sense?

I also had a look through the ZCL docs, and there is nothing in there to link color and color-temperature. The two seem to be treated as separate modes. The docs talk about switching modes when a color temperature command is received, but it doesn't say that is should update the current X/Y or Hue attributes, so I would assume they should remain set as they were.

TomTravaglino:
I changed ZigBeeConverterColorColor to set the channel state to UNDEF when the color mode turns to color temp.
At the first glance it worked but then I observed that attribute changes are periodically reported to the converters which then overwrite the state. This happens although the values of the attributes don't change.

Is this an expected behavior?

I tested with a Osram Smart+ Classic E27 Multicolor and a Osram Lightify CLA60 RGBW.
I attached a log file which contains the item events and the log entries produced by the converters.

[colormode-attribute-changes.txt](https://github.com/openhab/org.openhab.binding.zigbee/files/2727087/colormode-attribute-changes.txt)

cdjackson:
I've not looked at the log, but I suspect what you are seeing is what I eluded to above? That is that attribute reports will be sent to report color, even when the bulb is in the CT mode, and unless you use the state you will update the color when the device is in CT mode?

Is that what you see? I will take a look at the log later, but if I understand your question, then yes, this is expected.

TomTravaglino:
> That is that attribute reports will be sent to report color, even when the bulb is in the CT mode

And report happens although nothing changed.

> and unless you use the state you will update the color when the device is in CT mode

I update the state of the color channel although the device is in CT mode.

cdjackson:
> And report happens although nothing changed.

Yes, that is correct. In the reporting, we have to set a minimum and maximum report time. So if there are changes, the reports will come in at the minimum report time, and if there is no change the reports will come in at the maximum report time.

> I update the state of the color channel although the device is in CT mode.

I would suggest that you need to not update color if it is in CT mode - ie you need to remember the color mode, and only update color if CM is in color mode, and only update CT if CM is in CT mode. Otherwise the system is going to propagate incorrect state to the user.

TomTravaglino:
@cdjackson I updated the PR and had to merge master into my branch. Unfortunately the checks fail with this message

`14:45:12 [ERROR] Failed to execute goal org.openhab.tools.sat:sat-plugin:0.5.0:checkstyle (default) on project org.openhab.binding.zigbee: Unable to execute mojo: Execution null of goal org.apache.maven.plugins:maven-checkstyle-plugin:2.17:checkstyle failed: Plugin org.apache.maven.plugins:maven-checkstyle-plugin:2.17 or one of its dependencies could not be resolved: Could not find artifact org.eclipse.platform:org.eclipse.equinox.p2.touchpoint.eclipse:jar:2.2.200 -> [Help 1]`

Maybe a temporary problem. Is there a way to restart the checks?
Btw, on my local machine the build completes without any issues

TomTravaglino:
The checks have passed now. Apparently it was a temporary problem.

TomTravaglino:
Gentle reminder 😉

cdjackson:
Sorry - I will look at this once CI is working again.

AfromD:
I am watching this with interest, as I had a similar request some time ago.
https://github.com/openhab/org.openhab.binding.zigbee/issues/39

I have solved this privately by programming a dummy CT channel as a rule, and using that to convert CT commands to HSB commands, thus always keeping my bulbs in HSB mode whilst being able to give color temperature commands. I kind of made the same conclusions as you all did above, that the color channel is always valid and useful, that is why I am translating CT into HSB. This way the state of the bulb is always refelected correctly in the HSB channel.
Now when actually using the CT for control, it would be useful to have a mechanism to reflect that back into the HSB channel. (saves you from checking CT for !NULL and then applying custom logic). I had a look at Tom's commits above, and didn't find it there.
In my rule I use the following HSB values for any CT, which seem right for philips lights: Hue fixed at 45, saturation= CT as PercentType * 0,6, and brightness as is. 
Just as an idea, what to update the HSB channel with whenever CT is commanded.

jfrank1845:
Joining the conversation here vs the one on community.openhab - Anyway Chris your last post mentioned you guys might implement a dimmer channel back again so we have CT/Dimmer/Color?? That does seem to make a lot of sense to me as that would then allow us to always see a brightness state.  One other thing I just thought of regarding this current UNDEF is that if we are in CT mode and want to go back to color mode keeping the previous state we currently have no way to do that as the color state would have been switch to UNDEF.  It would almost be best to have a switch that shows color mode on/off and then leave the CT and Color items alone... that way you can send the command to the color mode to switch it.  All in that would then leave us with Color HSB , Brightness Dimmer, CT Dimmer and a color mode switch.  I know the brightness is redundant when using HSB but obviously we want to keep color as an HSB type.  I think that channel layout would allow for the most robust management and operation of the bulbs.   What do you guys think?

SuperSchelli:
I got loads of 0210 Bulbs from Philips and i am using the deconz bridge. Using the latest Snapshot build.

Atm, you can read the Brightness value by just assigning a Dimmer to the color channel, but as stated you have the problem, that if you just change one channel, the others are  not updated. If you then touch another channel there can be a huge jump in bulb visibility, since this channel has some random old value... 
I saw ways to derive an approximate ColorTemperature Value from the color Value and vice versa posted by CWeitkamp some place.

I was about to program my OH2 functionality with rules as following :
As jfrank1845 posted above. 4 Channels.

1 Switch between ColorTemperature and Color Mode (so i know which mode is active)
2 Dimmer Brightness
3 Dimmer CT
4 Color Channel HSB

When you adjust any of the 3 Channels 2-4, the other 2 Channels are updated with the respective values or the ones that are closest to the changed Channel.
If you move 2 - the mode remains whatever it is, just brightness changes
If you move 3 - you switch to CT mode but adjust the color channel along with it
if you move 4 - you switch to color mode but adjust the CT channel and brightness with it

cdjackson:
> i am using the deconz bridge

This binding doesn't support the deconz bridge so I suspect you're using the deconz binding?

SuperSchelli:
Yep, true, but i am having the exact same trouble with the hue binding emulated by deconz atm. have to take a look there what this is doing.  
sorry for the screwup, but the discussion is interesting nonetheless
i have ordered a cc2531 USB Stick and wanted to try around with this one as well, that should use this binding then if i am not mistaken.

cdjackson:
If you want to get something implemented, then it will need to be in the other binding… However, I think this might raise cross binding questions on how to handle this sort of thing, so feel free to comment ;)

SuperSchelli:
Well. I wanted to add, that there are also bulbs, that can only handle a color channel and no CT channel.

Now I have rooms which have a mixture of bulbs in them, but now I just want it darker, I don't want to adjust all bulbs on their own. so u want 1 brightness slider for the whole room, but depending on which bulb I control this command needs To be translated into the right channel for each bulb, depending on its capabilities. 
Now next imagine I want to change colors with only one slider, but some bulbs don't have a color channel. still if that color is a lot darker, it would be cool for the other bulbs to at least dim down a bit, or go to a warmer CT mode to blend in more with the selected color. this "problem" is indeed not a binding related thing (or you could also say, faces every binding)., so it could be more like a feature for Openhab, since I guess more people want this feature sooner or later... maybe a function to translate values or just have the ability to stick all the bulbs in one group and use one slider...

PR #337, opened 2018-12-14
Fixes log statements with format parameters and exceptions

hsudbrock:
At several places, there are log statements of the form `logger.info("some format string with one format parameter", formatParameter, exception)`. This results in not logging the exception (as it is considered as a format parameter, which is not used in the format). This commit circumvents this by using `String.format` in those cases.

### DIFF
--- a/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/discovery/ZigBeeDiscoveryService.java
+++ b/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/discovery/ZigBeeDiscoveryService.java
@@ -222,7 +222,9 @@ public class ZigBeeDiscoveryService extends AbstractDiscoveryService implements 
                             thingDiscovered(discoveryResult);
                         }
                     } catch (Exception e) {
-                        logger.error("Participant '{}' threw an exception", participant.getClass().getName(), e);
+                        logger.error(
+                                String.format("Participant '%s' threw an exception", participant.getClass().getName()),
+                                e);
                     }
                 }
 
--- a/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeCoordinatorHandler.java
+++ b/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeCoordinatorHandler.java
@@ -165,7 +165,7 @@ public abstract class ZigBeeCoordinatorHandler extends BaseBridgeHandler
                 linkKeyString = (String) param;
             }
         } catch (ClassCastException | NumberFormatException e) {
-            logger.error("{}: ZigBee initialisation exception ", thing.getUID(), e);
+            logger.error(String.format("%s: ZigBee initialisation exception", thing.getUID()), e);
             updateStatus(ThingStatus.OFFLINE);
             return;
         }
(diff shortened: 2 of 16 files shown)


cdjackson:
Are you really sure about this @hsudbrock ?

I am sure I've tested this, and also pretty sure that the logger methods allow an exception to be on the end of the format string, without the ```{}```. The exception will then be appended.

cdjackson:
From the [SLF4J FAQ](https://www.slf4j.org/faq.html#paramException) -:

image


I also just ran a test and it works here. 

```
        Logger logger = LoggerFactory.getLogger(ZigBeeConsoleMain.class);

        Exception e = new NullPointerException();
        logger.debug("This is {} and exception", "1", e);
```

Resulted in -:

```
09:55:25.158  DEBUG  This is 1 and exception
java.lang.NullPointerException
	at com.zsmartsystems.zigbee.console.main.ZigBeeConsoleMain.main(ZigBeeConsoleMain.java:116)
```

Maybe there's a version issue and you are using an old logging framework as I note that the FAQ above states this was added in SLF4J  ```1.6.0```? openHAB seems to be using ```1.7.2``` at the moment, so it should be no problem to use this formatting.

I also did a search in ESH, and while there is a large number of different logging systems used for exceptions, this doesn't seem to be uncommon.

https://github.com/eclipse/smarthome/blob/7176f8961a4caba501fe38bab52a44c8705011c3/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/service/WatchQueueReader.java#L151

Personally I would prefer to keep the cleaner form of logging without adding all the format statements.

Or do I somehow misunderstand something with your PR and there's something that is different about these particular log statements?

hsudbrock:
No, I am not 100% sure, I might have used an outdated slf4j. Let me check, and if you are right (which I believe you are) I will reject this PR and add a PR replacing such format statements I added in past PRs due to my lack of knowledge...



Am 15. Dezember 2018 11:18:10 MEZ schrieb Chris Jackson <notifications@github.com>:
>From the [SLF4J FAQ](https://www.slf4j.org/faq.html#paramException) -:
>
>image
>
>
>I also just ran a test and it works here. 
>
>```
>      Logger logger = LoggerFactory.getLogger(ZigBeeConsoleMain.class);
>
>        Exception e = new NullPointerException();
>        logger.debug("This is {} and exception", "1", e);
>```
>
>Resulted in -:
>
>```
>09:55:25.158  DEBUG  This is 1 and exception
>java.lang.NullPointerException
>	at
>com.zsmartsystems.zigbee.console.main.ZigBeeConsoleMain.main(ZigBeeConsoleMain.java:116)
>```
>
>Maybe there's a version issue and you are using an old logging
>framework as I note that the FAQ above states this was added in SLF4J 
>```1.6.0```? openHAB seems to be using ```1.7.2``` at the moment, so it
>should be no problem to use this formatting.
>
>I also did a search in ESH, and while there is a large number of
>different logging systems used for exceptions, this doesn't seem to be
>uncommon.
>
>https://github.com/eclipse/smarthome/blob/7176f8961a4caba501fe38bab52a44c8705011c3/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/service/WatchQueueReader.java#L151
>
>Personally I would prefer to keep the cleaner form of logging without
>adding all the format statements.
>
>Or do I somehow misunderstand something with your PR and there's
>something that is different about these particular log statements?
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/openhab/org.openhab.binding.zigbee/pull/337#issuecomment-447557026

-- 
Diese Nachricht wurde von meinem Android-Gerät mit K-9 Mail gesendet.

hsudbrock:
Thanks for your feedback, Chris - you are absolutely correct, with the more recent slf4j implementations it is fine as is, and I will close this PR.

PR #341, opened 2018-12-20
Reduce logging of bind error to debug level

cdjackson:
We are a little inconsistent with the logging in different converters, but in any case I don't think this should be logged as ERROR level since the binding will just revert to polling and the user won't see much difference. This seems to be more consistent with what I've done elsewhere (where anything is logged at all!).


cdjackson:
@hsudbrook - FYI there does seem to be some sort of bug in the library that sometimes doesn’t correlate the bind responses - the bind is working ok, and the response received, but the transaction times out and we get this error which is causing concern in OH at the moment.

I will look at the bind issue separately.

PR #344, opened 2018-12-23
Add ReportingConfiguration handler

### LINKED ISSUE #305: [enhancement] Make polling/reporting interval configurable
Add an option to configure the polling/reporting interval per device.

As discussed in #49, this would allow to work around issues with devices which are not properly reporting state changes, resulting in a disparity of the device state in OH and the actual device state.

Two such devices are the Busch-Jaeger 6711 relay insert and 6715 dimmer insert. Both devices only report their state at the default interval of 2 hours but not if a state change is triggered by the wall-mounted switch.

cdjackson:
This adds reporting and polling configuration handler. It is incomplete as it requires #342 and possibly a subsequent PR to refactor the LevelControlHandler before it can be completed.

Closes #305 

### DIFF
--- a/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeThingHandler.java
+++ b/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeThingHandler.java
@@ -59,6 +59,7 @@ import org.openhab.binding.zigbee.internal.ZigBeeDeviceConfigHandler;
 import org.openhab.binding.zigbee.internal.converter.ZigBeeChannelConverterFactory;
 import org.openhab.binding.zigbee.internal.converter.config.ZclClusterConfigFactory;
 import org.openhab.binding.zigbee.internal.converter.config.ZclClusterConfigHandler;
+import org.openhab.binding.zigbee.internal.converter.config.ZclReportingConfig;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -142,7 +143,7 @@ public class ZigBeeThingHandler extends BaseThingHandler implements ZigBeeNetwor
     /**
      * Creates a ZigBee thing.
      *
-     * @param zigbeeDevice the {@link Thing}
+     * @param zigbeeDevice   the {@link Thing}
      * @param channelFactory the {@link ZigBeeChannelConverterFactory} to be used to create the channels
      */
     public ZigBeeThingHandler(Thing zigbeeDevice, ZigBeeChannelConverterFactory channelFactory) {
@@ -368,6 +369,10 @@ public class ZigBeeThingHandler extends BaseThingHandler implements ZigBeeNetwor
                     continue;
                 }
 
+                if (channel.getConfiguration().get(ZclReportingConfig.CONFIG_POLLING) == null) {
+                    channel.getConfiguration().put(ZclReportingConfig.CONFIG_POLLING, handler.getPollingPeriod());
+                }
+
                 handler.handleRefresh();
 
                 // TODO: Update the channel configuration from the device if method available
@@ -418,7 +423,7 @@ public class ZigBeeThingHandler extends BaseThingHandler implements ZigBeeNetwor
      * Process a static cluster list and add it to the existing list
      *
      * @param initialClusters a collection of existing clusters
-     * @param newClusters a string containing a comma separated list of clusters
+     * @param newClusters     a string containing a comma separated list of clusters
      * @return a list of clusters if the list is updated, or an empty list if it has not changed
      */
     private List<Integer> processClusterList(Collection<Integer> initialClusters, String newClusters) {
@@ -642,7 +647,7 @@ public class ZigBeeThingHandler extends BaseThingHandler implements ZigBeeNetwor
      * changes.
      *
      * @param channel the {@link ChannelUID} to be updated
-     * @param state the new {link State}
+     * @param state   the new {link State}
      */
     public void setChannelState(ChannelUID channel, State state) {
         if (firmwareUpdateInProgress) {
@@ -660,7 +665,7 @@ public class ZigBeeThingHandler extends BaseThingHandler implements ZigBeeNetwor
      * received.
      *
      * @param channel the {@link ChannelUID} to be triggered
-     * @param event the event to be emitted
+     * @param event   the event to be emitted
      */
     @Override
     public void triggerChannel(ChannelUID channel, String event) {
--- a/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/config/ZclReportingConfig.java
+++ b/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/config/ZclReportingConfig.java
@@ -1,0 +1,188 @@
+/**
+ * Copyright (c) 2010-2018 by the respective copyright holders.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.openhab.binding.zigbee.internal.converter.config;
+
+import java.math.BigDecimal;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Objects;
+
+import org.eclipse.jdt.annotation.NonNull;
+import org.eclipse.smarthome.config.core.ConfigDescriptionParameter;
+import org.eclipse.smarthome.config.core.ConfigDescriptionParameter.Type;
+import org.eclipse.smarthome.config.core.ConfigDescriptionParameterBuilder;
+import org.eclipse.smarthome.config.core.Configuration;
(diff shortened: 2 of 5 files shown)


hsudbrock:
Hi Chris; quick question on this PR: Is it already sensible to review from your side, or will you still change the PR? (Asking as you write that the PR is incomplete in the PR description, but maybe you already completed it but simply forgot to update the PR description.)

cdjackson:
I think it is complete as #342 was merged some time back now.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/config/ZclReportingConfig.java
```diff
@@ -0,0 +1,179 @@
...
+import org.eclipse.smarthome.config.core.ConfigDescriptionParameter.Type;
+import org.eclipse.smarthome.config.core.ConfigDescriptionParameterBuilder;
+import org.eclipse.smarthome.config.core.Configuration;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.zsmartsystems.zigbee.zcl.ZclCluster;
+
+/**
+ * Configuration handler for reporting of digital attributes. This should be applied for each channel - ie if a channel
```
hsudbrock: What is the reason for fixing the same reporting configuration for all attributes? Is it simply for having a simpler structure of the configuration? Or is it so rare that it is sensible to configure different reporting options for different attributes used by one channel?

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeThingHandler.java
```diff
@@ -368,6 +369,10 @@ private synchronized void doNodeInitialisation() {
                     continue;
                 }
 
+                if (channel.getConfiguration().get(ZclReportingConfig.CONFIG_POLLING) == null) {
+                    channel.getConfiguration().put(ZclReportingConfig.CONFIG_POLLING, handler.getPollingPeriod());
```
hsudbrock: I believe that by setting the polling period configuration for every channel, users will think that it is also configurable for every channel; but, as far as I see it, the configuration is currently only applied for switch level and switch on off converters.

Shouldn't the polling period config value only be set for the channels where it can actually be configured?

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/config/ZclReportingConfig.java
```diff
@@ -0,0 +1,179 @@
...
+    public static final String CONFIG_POLLING = CONFIG_ID + "polling";
+
+    private int reportingTimeMin = REPORTING_DEFAULT_MIN;
+    private int reportingTimeMax = REPORTING_DEFAULT_MAX;
+    private int reportingChange = 1;
+    private int pollingPeriod = POLLING_DEFAULT;
+
+    // Reporting change configuration. These will need to be set for each channel.
+    private boolean isAnalogue = false;
+    private BigDecimal defChange;
```
hsudbrock: Small suggestion: Call variable `defaultChange`, to make a little more clear what the variable is for.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/config/ZclReportingConfig.java
```diff
@@ -0,0 +1,179 @@
...
+        parameters.add(ConfigDescriptionParameterBuilder.create(CONFIG_REPORTINGMAX, Type.INTEGER)
+                .withLabel("Maximum Reporting Period")
+                .withDescription("The maximum time period in seconds between device state updates")
+                .withDefault(REPORTING_DEFAULT_MAX.toString()).withMinimum(new BigDecimal(REPORTING_RANGE_MIN))
+                .withMaximum(new BigDecimal(REPORTING_RANGE_MAX)).build());
+
+        if (isAnalogue) {
+            parameters.add(ConfigDescriptionParameterBuilder.create(CONFIG_REPORTINGCHANGE, Type.INTEGER)
+                    .withLabel("Report On Change")
+                    .withDescription("The maximum time period between device state updates")
```
hsudbrock: Should the description rather be something like `The minimum change of the attribute value needed to trigger a device state update`?

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeThingHandler.java
```diff
@@ -368,6 +369,10 @@ private synchronized void doNodeInitialisation() {
                     continue;
                 }
 
+                if (channel.getConfiguration().get(ZclReportingConfig.CONFIG_POLLING) == null) {
```
hsudbrock: I am wondering why this is only done for the polling period here, but not for the other configuration parameters (reportingmax, reportingmin, reportingchange). Is this just because, currently, the reporting configuration is only used for the polling interval (and this needs, hence, to be extended as soon as some converter also makes use of the other reporting configuration options)?

cdjackson: This is the feature we discussed elsewhere - if the reporting configuration fails to be set, then we have to update the polling period. Polling period gets set to different values following the initialisation - the reporting is not changed.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/config/ZclReportingConfig.java
cdjackson: I could change I guess - to me it was quite clear, and in line with the reduced names on minimum and maximum to min and max. Do you really think this name needs to be extended?

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeThingHandler.java
hsudbrock: OK, thanks for the info!

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/config/ZclReportingConfig.java
cdjackson: Attribute reporting is the same for all attributes. Well, to be more specific, there are two types - digital and analogue, and this handler will handle both types, so it's not exactly the same for all attributes.

Maybe I misunderstand your point somehow?

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeThingHandler.java
cdjackson: Sorry - i missed this comment.

All converters return a polling period, and most should work shouldn't they? Do you mean that this config handler is only (currently) applied to level and on/off channels? If that's what you mean, then this config will currently only show for these channels so I'm not sure what you mean by this comment - sorry.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/config/ZclReportingConfig.java
hsudbrock: My question relates to the JavaDoc added here, which says:

> This should be applied for each channel - ie if a channel uses multiple attributes, then they should all be configured the same.

Why is it required to configure all attributes used within a channel the same? Assume a channel uses two digital attributes `A` and `B` - why not permit to configure reporting for those two attributes differently?

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeThingHandler.java
hsudbrock: I think I see your point now; the config is set, but not used, but this will not hurt anybody... Thanks for your response on this.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/config/ZclReportingConfig.java
cdjackson: I'm not 100% sure why I wrote this, but I think this was referring to a channel where you might get the data from multiple attributes (ie it's not just a case of reading a single attribute).

There are some examples (eg pressure sensor) but I think maybe it's confusing (especially as I don't recall why I wrote this!).

hsudbrock: ok, thanks Chris!

cdjackson:
Sorry Henning - what was the question? GH doesn't seem to show any questions/comments any more - I'm not sure if they are all hidden/removed when the PR is merged...

hsudbrock:
You already answered it after I commented above, and I resolved the corresponding discussion, so all good here!

PR #346, opened 2018-12-23
Generic buttons defined by attribute changes

t-8ch:
This PR extends the generic button functionality. It allows the definition of buttons based on attribute changed events.
An example configuration for a Xiaomi Aqara wallswitch is provided.

The logic and checks to validate the configuration should be cleaned up.
Also tests are need.
If there is interest in this feature I will provide them.

### DIFF
--- a/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterGenericButton.java
+++ b/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterGenericButton.java
@@ -9,18 +9,17 @@
 package org.openhab.binding.zigbee.internal.converter;
 
 import static java.lang.Integer.*;
-import static java.lang.String.format;
 
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.Map;
+import java.util.*;
 import java.util.Map.Entry;
-import java.util.Objects;
-import java.util.Set;
 import java.util.concurrent.ExecutionException;
+import java.util.function.Consumer;
+import java.util.function.Function;
+import java.util.function.Predicate;
 
+import com.zsmartsystems.zigbee.zcl.*;
 import org.eclipse.smarthome.core.thing.Channel;
 import org.eclipse.smarthome.core.thing.CommonTriggerEvents;
 import org.eclipse.smarthome.core.thing.ThingUID;
@@ -30,9 +29,6 @@ import org.slf4j.LoggerFactory;
 
 import com.zsmartsystems.zigbee.CommandResult;
 import com.zsmartsystems.zigbee.ZigBeeEndpoint;
-import com.zsmartsystems.zigbee.zcl.ZclCluster;
-import com.zsmartsystems.zigbee.zcl.ZclCommand;
-import com.zsmartsystems.zigbee.zcl.ZclCommandListener;
 
 /**
  * Generic converter for buttons (e.g., from remote controls).
@@ -43,8 +39,9 @@ import com.zsmartsystems.zigbee.zcl.ZclCommandListener;
  * As the configuration is done via channel properties, this converter is usable via static thing types only.
  *
  * @author Henning Sudbrock - initial contribution
+ * @author Thomas Weißschuh - support for attribute-based buttons
  */
-public class ZigBeeConverterGenericButton extends ZigBeeBaseChannelConverter implements ZclCommandListener {
+public class ZigBeeConverterGenericButton extends ZigBeeBaseChannelConverter implements ZclCommandListener, ZclAttributeListener {
 
     private Logger logger = LoggerFactory.getLogger(this.getClass());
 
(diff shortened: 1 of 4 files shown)


org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterGenericButton.java
```diff
@@ -74,30 +75,59 @@ public synchronized boolean initializeConverter() {
...
+                }
+
+                try {
+                    CommandResult bindResponse = bind(serverCluster).get();
+                    if (!bindResponse.isSuccess()) {
+                        logger.error("{}: Error 0x{} setting server binding for cluster {}", endpoint.getIeeeAddress(),
+                                toHexString(bindResponse.getStatusCode()), clusterId);
+                    }
+                } catch (InterruptedException | ExecutionException e) {
+                    logger.error(endpoint.getIeeeAddress() + ": Exception setting server binding to cluster " + clusterId, e);
```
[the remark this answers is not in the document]
t-8ch: Fixed

t-8ch:
Will do!

t-8ch:
Currently the value of the attribute is compared via via its string representation.
This works nicely around the fact, that the configuration only allows string values, while the actual attribute may have a wider range of types.
However this will make it impossible to specify integer values as hex, like the other values.
Do we want to:
* Force users to specify the exact string versions
* Special-case integer attribute values
* Only allow integer attribute values
?

t-8ch:
How should the tests be grouped/ordered in relation to the existing command tests?
Should they by similar functionality or group by command/attribute?

t-8ch:
It seems the channels can't be linked from the UI, as no profile can be selected.
Is this intentional/known or a bug in thing definition?

t-8ch:
@cdjackson One question about the configuration application:
Currently after pairing the switch from the inbox it will take some time until it is fully discovered. After the device is fully discovered a second Thing, with the properties and channels configured via the XML files, appears in the Inbox.
Adding this Thing will again trigger a full discovery process.
In the end there are two Things configured, one generic and one properly usable.
Is this intentional or is my configuration somehow broken?

cdjackson:
No - this is not the intention. There should only be a single thing discovered - initially the generic device, and finally the specific device. I am reasonably sure that it used to work like this, but I also know there have been some changes in ESH recently to remove some inbox functionality, so maybe this is now broken.

t-8ch:
Ok. Maybe someone with one of the other configured devices can try to reproduce the issue.
Otherwise I will have to do some digging.

hsudbrock:
Hi @t-8ch, thanks for contributing!

I had a first look at the PR, and before I start a detail-level review, I have a general question. The reason for the question is that, at first glance, it looks a little odd to me to use trigger channels in combination with ZigBee attribute reporting - but maybe this is only due to how the Xiaomi Aqara wall switch (which I have not used so far) is implemented.

You have added three trigger channels for the wall switch, `left`, `right`, and `range-check`; consider `left` for now, assuming that this models button presses for the left button of the wall switch.

The channel triggers when the device reports a new value for the attribute `85` of cluster `0x12` (value `1` for single press, `0` for long press, and `2` for double press).

The question: Will the value of the attribute always indicate the last press that was performed on the left button of the device? I.e., imaging doing a double press, then waiting for 10 seconds, and then reading the attribute from the device. Will the attribute value still be `2`? If not, which attribute value will the device show when reading the attribute after 10 seconds?

Background of the question: It seems strange to me that an attribute is used to model an action like a button press, and I wonder why this is modeled like this for the wall switch. Attributes smell like some state that can be retrieved from the device - which could, maybe, be better modeled with a state channel and not a trigger channel in openHAB.

t-8ch:
To be honest, I only ever used the attribute changed commands emitted by the switch presses.
(That always and promptly are emitted on each button press, even when the same button was already pressed last time).
I will have to double check on the actual value of the attribute.

hsudbrock:
Thanks for the feedback, @t-8ch. From what you write, it sounds like the wall switch uses attributes in a - for me - strange way (which might be completely normal when looking at it from another angle though...) From what you write, it sounds reasonable to cover this with the `GenericButtonConverter`.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterGenericButton.java
```diff
@@ -74,41 +75,65 @@ public synchronized boolean initializeConverter() {
...
-            } catch (InterruptedException | ExecutionException e) {
-                logger.error(endpoint.getIeeeAddress() + ": Exception setting binding to cluster " + clusterId, e);
-            }
+        return true;
+    }
 
-            clientCluster.addCommandListener(this);
-            clientClusters.add(clientCluster);
+    private boolean bindCluster(String clusterType, Collection<ZclCluster> existingClusters, int clusterId,
+                                Function<Integer, ZclCluster> clusterFunction,
```
hsudbrock: small suggestion: rename `clusterFunction` to `getClusterByIdFunction` (or some other name telling what the function does). WDYT?

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterGenericButton.java
```diff
@@ -43,23 +40,27 @@
...
-    private Map<ButtonPressType, CommandSpec> handledCommands = new HashMap<>();
+    private Map<ButtonPressType, CommandSpec> handledCommands = new EnumMap<>(ButtonPressType.class);
     private Set<ZclCluster> clientClusters = new HashSet<>();
+    private Set<ZclCluster> serverClusters = new HashSet<>();
 
     @Override
     public synchronized boolean initializeConverter() {
         for (ButtonPressType buttonPressType : ButtonPressType.values()) {
-            CommandSpec commandSpec = parseCommandSpec(buttonPressType);
+            CommandSpec commandSpec = parseCommandSpec(channel.getProperties(), endpoint.getIeeeAddress(), buttonPressType);
```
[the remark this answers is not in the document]
t-8ch: Because `parseCommandSpec` was made `static` and won't have access to the endpoint directly. I will remove the `static`.

t-8ch:
Thanks for the feedback. I will try to implement it soon.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterGenericButton.java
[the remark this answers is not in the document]
cdjackson: Why does it need to be static? It doesn't seem to be used statically from my quick look (other than in tests, and that's generally not a good reason to make it static). Or do you mean you will remove the static from the ```parseCommandSpec``` method?

t-8ch:
@hsudbrock I update the PR to incorporate your suggestions.
Could you have another look?

FYI: Currently I don't have access to the switch, so this version of the PR is not actually tested with a physical device.

t-8ch:
The CI seems to be happy now.

hsudbrock:
@t-8ch Thanks for the update. Sorry, I did not get to have a look again yet, will do this soon!

Just one hint, as I see that you force-pushed to your branch: For the reviewer, it is often easier if you add new commits instead of changing commits, so that one can see what has changed in addition. (When merging the PR github will squash the commits anyway...)

t-8ch:
If you prefer it this way sure.
I tried to create logical commits, as the PR does multiple things. Would it be ok, if I add more commits and just before the merge I reorganize them to be 3 (or 2) commits again?

hsudbrock:
> If you prefer it this way sure.

Thanks; no need to change anything now for the review though, just as hint for next time :)

> Would it be ok, if I add more commits and just before the merge I reorganize them to be 3 (or 2) commits again?

Fine for me; as far as I recall, github will squash the commits though when merging, creating a commit that has the PR title as commit message, so the reorganization might not be visible in this project's commit history anymore. Not sure 100% though...

t-8ch:
You should be able to choose when performing the merge.

hsudbrock:
OK, thanks for the hint!

t-8ch:
Hm, I will also perform some cosmetic cleanups. (Rewording, things that should not be changed, etc)

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterGenericButton.java
```diff
@@ -43,25 +39,29 @@
  * As the configuration is done via channel properties, this converter is usable via static thing types only.
  *
  * @author Henning Sudbrock - initial contribution
+ * @author Thomas Weißschuh - support for attribute-based buttons
  */
-public class ZigBeeConverterGenericButton extends ZigBeeBaseChannelConverter implements ZclCommandListener {
+public class ZigBeeConverterGenericButton extends ZigBeeBaseChannelConverter implements ZclCommandListener, ZclAttributeListener {
 
-    private Logger logger = LoggerFactory.getLogger(this.getClass());
+    private Logger logger = LoggerFactory.getLogger(ZigBeeConverterGenericButton.class);
```
hsudbrock: I suggest to leave the logger definition as it was (i.e., using `this.getClass()`); This will yield the correct logger names when using subclassing. Although subclassing is not used here, using `this.getClass()` frees the reader from having to think about why the class object is used instead of the more dynamic `this.getClass()`.

hsudbrock: I would like to learn here: What is the advantage of not using `this.getClass()`, but rather using `ZigBeeConverterGenericButton.class`?

t-8ch: When looking for the source (code) of an issue one knows where to find it.
Also, it would be more predictable which logger to configure when the logger name does not depend on a (potentially unknown) subclass.

hsudbrock: Thanks for your feedback!

hsudbrock:
Hi @t-8ch; could you please rebase the PR on the `master` branch, so that it can be merged?

I just had a look at the merge conflict, I think it is only two things to do:
* For the merge conflict in `discovery.txt`, simply use the line added by this PR and the line that was added by #366 
* For the merge conflict in `ZigBeeConverterGenericButton`, add the improvement from #366 into the method `CommandSpec#matches` as follows:

```@Override
        boolean matches(ZclCommand command) {
            if (commandId == null) {
                return false;
            }
            boolean commandIdMatches = command.getCommandId().intValue() == commandId;
            return commandIdMatches
                    && (commandParameterName == null || commandParameterValue == null || matchesParameter(command));
        }
```

WDYT?

t-8ch:
@hsudbrock Done

hsudbrock:
@t-8ch Thanks!

PR #347, opened 2018-12-24
Add option for Ember child aging

cdjackson:
This adds support to change the way that Ember coordinators age out devices in their child tables.

ZigBee routers (and the coordinator) only have room to allow a certain number of devices to join the network via each router - once the child table in a router is full, devices will need to join via another router (assuming the child can communicate via another router). To avoid the child table becoming full of devices that no longer exist, routers will age out children that do not contact them within a specified period of time. 

Once a child is removed from the child table of a router, it will be asked to rejoin if it tries to communicate with the parent again. Setting this time too large may mean that the router fills its tables with devices that no longer exist, while setting it too small can mean devices unnecessarily rejoining the network.

Note that ZigBee compliant devices should rejoin the network seamlessly, however some non-compliant devices may not rejoin which may leave them unusable without a manual rejoin.

### DIFF
--- a/org.openhab.binding.zigbee.ember/ESH-INF/thing/controller_ember.xml
+++ b/org.openhab.binding.zigbee.ember/ESH-INF/thing/controller_ember.xml
@@ -84,6 +84,25 @@
 					<option value="0">Normal</option>
 				</options>
 			</parameter>
+			
+            <parameter name="zigbee_childtimeout" type="integer" groupName="ember">
+                <label>Child Aging Timeout</label>
+                <description>Sets the period over which the coordinator will age children. Children who have not checked in within this period will be removed from the child table may be asked to rejoin the network. Note that setting this value too high may prevent new devices joining the network.</description>
+                <default>86400</default>
+                <options>
+                    <option value="320">5 Minutes</option>
+                    <option value="1800">30 Minutes</option>
+                    <option value="7200">2 Hours</option>
+                    <option value="43200">12 Hours</option>
+                    <option value="86400">1 Day</option>
+                    <option value="172800">2 Days</option>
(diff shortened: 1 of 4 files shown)


cdjackson:
Minor updates following testing.

@hsudbrock I plan to merge this shortly as I prefer not to wait until you are all back in the office ;) . If you have any comments I'm happy to address them in another PR, but I'd like to get this merged to see if it improves usability.

cdjackson:
Rebased...

hsudbrock:
Looks good, Chris, thanks. One small question (issue?): The config option looks like the number of seconds corresponding to the configured time range. This fits for all time ranges, except for five minutes (where the config option is `320`, instead of the expected `300`).

cdjackson:
Yes - the reason for this is 320 is the default in the NCP, but I didn't want to write 5 minutes 20 seconds. I think many of these are not 100% correct since the equation to set them isn't in seconds - it's a square term and multiplier.

I don't think it's important to the user if it's 5 minutes, or 5 minutes 20 as they won't know exactly what the timing is anyway as they won't know exactly when messages are received in the coordinator.

If you feel strongly, I can change this to say ```5 minutes 20 seconds``` though?

hsudbrock:
I agree, for the user it is not important at all. I just stumbled over the slight inconsistency that - for all option values other than "5 minutes" - the number configured in the option corresponds exactly to the number of seconds indicated by the option description, but may differ slightly from the number actually configured on the NCP in `EmberHandler` via `pollTimeoutValue`and `pollTimeoutValue`. Only for the "5 minutes" it is the other way round (here the number from the option does not differ from the number configured in `EmberHandler`, but differs from 300 seconds).

But no strong feelings - maybe not worth a PR for changing.

PR #349, opened 2018-12-25
Provide configuration option for mesh update period

cdjackson:
This adds a configuration option to set the period used for the mesh updates...

The binding is able to search the network to get a list of what devices can communicate with other devices. This is a useful diagnostic feature as it allows users to see the links between devices, and the quality of these links. However, this can generate considerable traffic, and some battery devices may not poll their parents often enough to provide these updates, and users may consider that it is better to reduce the period, or disable this feature.

### DIFF
--- a/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeCoordinatorHandler.java
+++ b/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeCoordinatorHandler.java
@@ -109,7 +109,7 @@ public abstract class ZigBeeCoordinatorHandler extends BaseBridgeHandler
 
     private boolean macAddressSet = false;
 
-    private final int MESH_UPDATE_TIME = 300;
+    private final int MESH_UPDATE_PERIOD = 86400;
 
     /**
      * Set to true on startup if we want to reinitialize the network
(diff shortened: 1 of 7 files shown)


org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeCoordinatorHandler.java
```diff
@@ -373,9 +373,15 @@ private void initialiseZigBee() {
...
+        int meshUpdateTime = MESH_UPDATE_TIME;
+        if (getConfig().get(CONFIGURATION_MESHUPDATEPERIOD) != null) {
+            logger.debug("Mesh Update Period {}", getConfig().get(CONFIGURATION_MESHUPDATEPERIOD));
+            meshUpdateTime = ((BigDecimal) getConfig().get(CONFIGURATION_MESHUPDATEPERIOD)).intValue();
+        }
+
         // Add the extensions to the network
         ZigBeeDiscoveryExtension discoveryExtension = new ZigBeeDiscoveryExtension();
-        discoveryExtension.setUpdatePeriod(MESH_UPDATE_TIME);
+        discoveryExtension.setUpdatePeriod(meshUpdateTime);
```
hsudbrock: The configuration is named `mesh update period`, whereas the variable storing the config value here is named `mesh update time`, which seems like a (minor) inconsistency. If you consider this sensible as well, you could adapt the name of the variable here to match.

cdjackson: Yep - I agree - thanks.

PR #368, opened 2019-01-19
Skip parallel writing attempts of the network state

puzzle-star:
Following OH community [discussion](https://community.openhab.org/t/zigbee-binding/15763/1892?u=puzzle-star), provide a way for the binding to skip parallel writes of the network state file.

On a large network, the network state is serialized several times concurrently. As the serialization is  synchronized, there is no functional isses, but the network state is serialized several times (this is a costly operation), while only the first and "last" are really necessary.

This will definetely improve after changing serialization to write state per node instead of the full network, but meanwhile it might still be beneficious to make sure full network save events are skipped for parallel requests

### DIFF
--- a/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/ZigBeeNetworkStateSerializerImpl.java
+++ b/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/ZigBeeNetworkStateSerializerImpl.java
@@ -86,6 +86,9 @@ public class ZigBeeNetworkStateSerializerImpl implements ZigBeeNetworkStateSeria
      */
     private final String networkStateFilePath;
 
+    private final Object concurrentWriteCountSync = new Object();
+    private int concurrentWriteCount = 0;
+
     public ZigBeeNetworkStateSerializerImpl(String networkId) {
         this.networkId = networkId;
         networkStateFilePath = ConfigConstants.getUserDataFolder() + "/" + ZigBeeBindingConstants.BINDING_ID;
@@ -125,8 +128,7 @@ public class ZigBeeNetworkStateSerializerImpl implements ZigBeeNetworkStateSeria
      *
      * @param networkManager the network state
      */
-    @Override
-    public synchronized void serialize(final ZigBeeNetworkManager networkManager) {
+    private synchronized void doSerialize(final ZigBeeNetworkManager networkManager) {
         XStream stream = openStream();
 
         logger.debug("Saving ZigBee network state: Start.");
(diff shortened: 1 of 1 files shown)


cdjackson:
@puzzle-star why not use an ```AtomicBoolean```? That would negate the need for the synchronisation and counter and simplify this significantly I think?

Otherwise I'm happy to add this as a temporary measure, but as discussed will look to resolve this through a different serialisation API as per https://github.com/zsmartsystems/com.zsmartsystems.zigbee/issues/513.

t-8ch:
Or `Lock.tryLock()`

puzzle-star:
Hi @cdjackson  and @t-8ch,

I tried both ReentrantLocks and AtomicBooleans/AtomicIntegers.

At some moment we need to "decouple" the waiting thread from the synchronization of the counter, as we want still the counter to be checked and modified to return immediately while a thread waits and another is running.

The issue I found with any implementations I could think (and have tried) of ReentrantLocks and AtomicBooleans/AtomicIntegers is that there is a small fraction of time after the doSerialize finishes and before we decrease the counter when a new thread, could enter, and get (incorrectly) discarded as the the serialization has finished but the counter has not been updated yet.

I tried to fix it by adding extra locks, and the result was even more complex and deadlock-prone. The only solution that left no window of opportunity for a race condition was synchonization with wait() / notify(). This was not my first prefferred option.

Happy to revisit if you find a better alternative with AtomicBooleans/Integers or ReentrantLocks. Just please be aware that it is not that I did not consider these, but that indeed I considered them and discarded the option due to a race condition.

Of course, there might be other issues in my implemeentation, and it is worth reviewing, but if the issue is complexity, it is the simplest way I found without race conditions or deadlocks (that I could see anyway).

cdjackson:
> At some moment we need to "decouple" the waiting thread from the synchronization of the counter, 

Why do you need the counter? It's only counting to 1 as it's ensuring exclusive access.

There should not be a small amount of time - that is the purpose of the concurrency classes...

I don't really see why you can't simply do something like -:

```
if(!atomicBoolean.compareAndSet(false, true)) {
    return;
}
serialize();
atomicBoolean.set(false);
```

This should always return if the atomic boolean is already true, thus ensuring sole access to ```serialize();```

Other concurrent solutions should also work - there shouldn't be a need to do what you're doing with counters.

puzzle-star:
Hi @cdjackson,

> This should always return if the atomic boolean is already true, thus ensuring sole access to `serialize();`

True, and this would be a solution for a standard "latch" (maybe better anyway with trylock() for this): one thread passes and the others are discarded.

But we want:
- The first to pass, the second to wait, and the next ones to return immediately.
- If the first finishes, make sure we do not miss a "third" one that could come before we account for the situation

The first requirement could be fulfilled with one AtomicInteger (so we need a counter, or two booleans)

The second one, though, requires synchronization between two different parts of the code to wake up the waiting thread only after we accounted for the first one having finished. This requires synchronization, and AtomicBooleans do not include the capacity to "notify" a waiting thread atomically.

So the implementation for a latch is simple for a 1 thread latch, but not trivial (as far as I know) for a 2 thread latch fulfilling the former requirements.

Of course, I might have missed something...

cdjackson:
> The first thread to pass, the second one to wait, and the next ones to return immediately.

I didn't realise you want this - why not just keep it simple and only allow 1 - this should be fine and much simpler. This is only a temporary fix so I would keep it simple - otherwise maybe we are better to wait a few weeks and do it in the framework as discussed.

cdjackson:
> If we only allow one and skip the others, we will lose several network state saves.

Do you need to save every state change? I don't think so.

What is important is that you save the state before the system closes - that will always be done anyway - and that you save the state periodically in case the system crashes. In fact, this is not normally too important assuming you save the state right at the beginning after the interrogation is complete.

> I agree this solution is not as simple as a 1-thread latch, but anyway it is not really that complex once you look into it, and it seems to be working correctly at the moment

True, but we also need to remove this in a few weeks, so maybe it's easier not to add it ;)

PR #377, opened 2019-01-30
Update device/node config on thing handler initialization

### LINKED ISSUE #376: Update device/node config on thing handler initialization
We want the config properties of a thing to be set on the device when the thing handler is initialized. Therefore I will change the method `org.openhab.binding.zigbee.handler.ZigBeeThingHandler.doNodeInitialisation()`  to call `handleConfigurationUpdate(...)`. `handleConfigurationUpdate` and `org.openhab.binding.zigbee.internal.ZigBeeDeviceConfigHandler.updateConfiguration(...)` will also need little changes to process unchanged properties too.

TomTravaglino:
Fixes #376 

In ZigBeeThingHandler the method 'handleConfigurationUpdate'
(inherited by BaseThingHandler) has been overloaded by a new one which
also accepts a boolean to process also unchanged properties. This is
needed because the 'new' properties are compared with those ones stored
in the thing which are identical during the thing handler
initialization. Hence the 'new' properties would then be ignored.

In ZigBeeThingHandler.doNodeInitialisation() now the method
'handleConfigurationUpdate' is called passing the config properties
stored in the thing. These will overwrite the config properties defined
in zsmartsystems which are sent to the device by default when the ZigBee
network is formed.

In ZigBeeDeviceConfigHandler the method 'updateConfiguration' has been
overloaded by a new one which also accepts a boolean to process also
unchanged properties. This is needed because the 'new' properties are
compared with those ones stored in the thing which are identical during
the thing handler initialization. Hence the 'new' properties would then
be ignored.

### DIFF
--- a/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeThingHandler.java
+++ b/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeThingHandler.java
@@ -402,6 +402,12 @@ public class ZigBeeThingHandler extends BaseThingHandler implements ZigBeeNetwor
             logger.error("{}: Exception getting binding table ", nodeIeeeAddress, e);
         }
 
+        // Update the configuration of the device
+        Map<String, Object> configProperties = thing.getConfiguration().getProperties();
+        if (!configProperties.isEmpty()) {
+            handleConfigurationUpdate(configProperties, false);
+        }
+
         updateStatus(ThingStatus.ONLINE);
 
         startPolling();
@@ -555,12 +561,17 @@ public class ZigBeeThingHandler extends BaseThingHandler implements ZigBeeNetwor
 
     @Override
     public void handleConfigurationUpdate(Map<String, Object> configurationParameters) {
+        handleConfigurationUpdate(configurationParameters, true);
+    }
+
+    private void handleConfigurationUpdate(Map<String, Object> configurationParameters, boolean ignoreUnchangedParameters) {
         logger.debug("{}: Configuration received: {}", nodeIeeeAddress, configurationParameters);
 
         Configuration configuration = editConfiguration();
         for (Entry<String, Object> configurationParameter : configurationParameters.entrySet()) {
-            // Ignore any configuration parameters that have not changed
-            if (Objects.equals(configurationParameter.getValue(), configuration.get(configurationParameter.getKey()))) {
+            // If required, ignore any configuration parameters that have not changed
+            if (ignoreUnchangedParameters && Objects.equals(configurationParameter.getValue(),
+                    configuration.get(configurationParameter.getKey()))) {
                 logger.debug("{}: Configuration update: Ignored {} as no change", nodeIeeeAddress,
                         configurationParameter.getKey());
                 continue;
@@ -584,7 +595,7 @@ public class ZigBeeThingHandler extends BaseThingHandler implements ZigBeeNetwor
 
         ZigBeeNode node = coordinatorHandler.getNode(nodeIeeeAddress);
         ZigBeeDeviceConfigHandler deviceConfigHandler = new ZigBeeDeviceConfigHandler(node);
-        deviceConfigHandler.updateConfiguration(configuration, configurationParameters);
+        deviceConfigHandler.updateConfiguration(configuration, configurationParameters, ignoreUnchangedParameters);
 
         // Persist changes
         updateConfiguration(configuration);
--- a/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/ZigBeeDeviceConfigHandler.java
+++ b/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/ZigBeeDeviceConfigHandler.java
@@ -53,10 +53,24 @@ public class ZigBeeDeviceConfigHandler {
      */
     public void updateConfiguration(@NonNull Configuration currentConfiguration,
             Map<String, Object> updatedParameters) {
+        updateConfiguration(currentConfiguration, updatedParameters, true);
+    }
+
+    /**
+     * Processes the updated configuration. As required, the method shall process each known configuration parameter and
+     * set a local variable for local parameters, and update the remote device for remote parameters.
+     * The currentConfiguration shall be updated.
+     *
+     * @param currentConfiguration the current {@link Configuration}
+     * @param updatedParameters a map containing the updated configuration parameters to be set
+     * @param ignoreUnchangedParameters ignore parameters that have not changed
+     */
+    public void updateConfiguration(@NonNull Configuration currentConfiguration, Map<String, Object> updatedParameters,
+            boolean ignoreUnchangedParameters) {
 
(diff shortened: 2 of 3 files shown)


cdjackson:
Maybe I'm missing something here, but isn't ```ZigBeeDeviceConfigHandler.updateConfiguration``` always called with ```ignoreUnchangedParameters``` set to true?

Another option is to just remove this check - I don't really like having it there, but since PaperUI always transfers ALL config parameters, I think it is needed. My preference is for the UI to only pass through parameters that need to be set - this is what HABmin does, and I originally didn't have this check, but due to PaperUI, I added it...

TomTravaglino:
Is there any critical impact if we simply remove the check and the apps/UIs pass whatever they want?

cdjackson:
I wouldn't say "critical" - the issue is that UIs can send a lot of data, and this _might_ end up causing a lot of messages to be sent to devices each time you save the thing information. For battery devices especially, this _might_ cause problems.

Personally, I would prefer not to have to have this check, and for the UI to better manage the configuration, but this is an argument I lost a long time ago. At the moment, the statement from ESH is that the current configuration system is NOT to be used for configuring devices - only the thing handler - and therefore it's not really optimised for device config.

TomTravaglino:
> Maybe I'm missing something here, but isn't `ZigBeeDeviceConfigHandler.updateConfiguration` always called with `ignoreUnchangedParameters` set to true?

I added an invocation `handleConfigurationUpdate(configProperties, false)` in the method `doNodeInitialisation`. 

Some words to the background of this flag:
I want to create a thing with a specific configuration defined in the config description of the thing type. I observed that on thing creation such a config is not sent to the device. Instead only the default config defined in zsmartsystems (e.g. in `ZclOccupancySensingCluster`) is sent to the device. Therefore I had the idea to call `handleConfigurationUpdate` on thing handler initialization and pass the thing config to it. But in `handleConfigurationUpdate` these parameters (no matter where they come from) are compared with those ones stored in the thing and will be then ignored. So I introduced the flag `ignoreUnchangedParameters` to skip the comparison step.

cdjackson:
> Instead only the default config defined in zsmartsystems (e.g. in `ZclOccupancySensingCluster`) is sent to the device.

I'm not sure what you are referring to here? The ZSS libraries don't send any default configuration to the device so I think I'm misunderstanding what you are referring to? Can you point me at the code in the ZSS libraries that sends configuration?

In general, if you set an attribute, then it should be sent to the device, and I don't think the ZSS libs should change this?

TomTravaglino:
> I'm not sure what you are referring to here? The ZSS libraries don't send any default configuration

You're right, ZSS doesn't send a default config to the device. I (heavily) misunderstood the method `ZclOccupancySensingCluster.initializeAttributes()`. Guess the values in the device are it's own default values.

Back to the approach with the `ignoreUnchangedParameters` flag. What do you think of it?

cdjackson:
I still don't understand the use case, and I'm really not keen with this implementation. This will send a lot more configuration data each time the binding starts - why? I think this will cause a lot of problems.

hsudbrock:
I agree with Chris - I don't like that all configuration properties are sent to the device every time the thing handler starts up.

I believe that the use case is "restore a backup of a ZigBee thing, including restoring device configuration contained in the thing configuration to the ZigBee device" (please correct me if I am wrong, @TomTravaglino).

Maybe we can find a better solution with this use case in mind?

cdjackson:
Since this use case is really only a Qivicon function rather than openHAB, I'm inclined to suggest that it should be solved you your backup app...

It should (I think) be possible to remove (ie save) the configuration before the backup, then to set it again after the backup using the ```handleConfiguration``` method. This will also resend all configuration.

TomTravaglino:
We have two use cases.
- Pair a device: When the thing is created the config provided by the config description should be sent to device. This doesn't happen so far.
- Restore a backup: As you described, the config stored in the (restored) thing config should be sent to the device.

I started to investigate if it is possible to listen for thing creation events (e.g. `ThingAddedEvent`) and so send the config only once.

cdjackson:
> * When the thing is created the config provided by the config description should be sent to device. This doesn't happen so far.

Why not? If there is NO configuration, which should be the case when there is a new device, then the configuration should be sent. Please can you debug WHY this doesn't work?

TomTravaglino:
> Why not? If there is NO configuration, which should be the case when there is a new device, then the configuration should be sent. Please can you debug WHY this doesn't work?

Will check it again. I already tested with a device for which we have a config description and I'm sure the values of the cluster attributes were different from those in the config description. However, I will give feedback asap.

cdjackson:
This doesn't make sense though? Your PR was to (conditionally) remove the check that tests if the config is the same as the config that is set. From what you're saying, the configuration was different, but it was still not set - in this case I don't really see how this PR will help?

TomTravaglino:
Looks like confusion is coming up. When I wrote 
> I'm sure the values of the cluster attributes were different from those in the config description

I meant that this was what I observed before my PR. To ensure that my observation was correct I did a test with a version from `master`. I used a Centralite 3305-Geu (motion sensor) and this is its config-description (I have omitted the unimportant parts)
```
<?xml version="1.0"?>
<config-description uri="motion-sensor">
...
    <parameter name="attribute" endpoint="2" clusterId="0406" direction="in" attributeId="0012" dataType="20" type="integer" min="1" max="254" step="1">
      <default>3</default>
    </parameter>
    <parameter name="attribute" endpoint="2" clusterId="0406" direction="in" attributeId="0011" dataType="21" type="integer" min="0" max="65534" step="1">
      <default>6</default>
    </parameter>
    <parameter name="attribute" endpoint="2" clusterId="0406" direction="in" attributeId="0010" dataType="21" type="integer" min="0" max="65534" step="1">
      <default>9</default>
    </parameter>
</config-description>
```
These are the thing config parameters after the device was paired. It contains the values from the config description.
```
Configuration parameters:
	attribute_2_in_0406_0010_21 : 9
	attribute_2_in_0406_0011_21 : 6
	attribute_2_in_0406_0012_20 : 3
```
These are the cluster attribute values after the device was paired.
```
read 64743/2 1030 18
Reading server cluster Occupancy sensing (0406), Attribute 18
Cluster 0406, Attribute 18, type UNSIGNED_8_BIT_INTEGER, value: 1

read 64743/2 1030 17
Reading server cluster Occupancy sensing (0406), PIRUnoccupiedToOccupiedDelay
Cluster 0406, Attribute 17, type UNSIGNED_16_BIT_INTEGER, value: 0

read 64743/2 1030 16
Reading server cluster Occupancy sensing (0406), PIROccupiedToUnoccupiedDelay
Cluster 0406, Attribute 16, type UNSIGNED_16_BIT_INTEGER, value: 30
```
The values are different. That's why I wrote

> Pair a device: When the thing is created the config provided by the config description should be sent to device. This doesn't happen so far.

cdjackson:
Ok, but then I still have the same comment. If the updated configuration is not transferred to the device, then this change will not help still and I think you need to understand why it's not transferred. IF this is a battery device, then it may just be that the device is sleeping and therefore the data is not transferred.

TomTravaglino:
We will elaborate another solution for this requirement. So let's close this one.

PR #379, opened 2019-02-03
Add static definition for Xiaomi Wireless 2-Button Switch

puzzle-star:
This adds support for the non-standard Xiaomi lumi.sensor_86sw2, the two button wireless (battery powered) switches.

It adds three channels (switch_onoff cluster at endpoints 1, 2 and 3), for left button (switch_1), right button (switch_2) and both buttons at the same time (switch_3)

### DIFF
--- a/org.openhab.binding.zigbee/ESH-INF/thing/xiaomi/lumisensor86sw2.xml
+++ b/org.openhab.binding.zigbee/ESH-INF/thing/xiaomi/lumisensor86sw2.xml
@@ -17,13 +17,13 @@
 			</channel>
 			<channel id="switch_2" typeId="switch_onoff">
 				<properties>
-					<property name="zigbee_endpoint">1</property>
+					<property name="zigbee_endpoint">2</property>
 					<property name="zigbee_inputclusters">6</property>
 				</properties>
 			</channel>
 			<channel id="switch_3" typeId="switch_onoff">
 				<properties>
-					<property name="zigbee_endpoint">1</property>
+					<property name="zigbee_endpoint">3</property>
 					<property name="zigbee_inputclusters">6</property>
 				</properties>
 			</channel>
--- a/org.openhab.binding.zigbee/ESH-INF/thing/xiaomi/lumisensor86sw2.xml
+++ b/org.openhab.binding.zigbee/ESH-INF/thing/xiaomi/lumisensor86sw2.xml
@@ -1,0 +1,46 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<thing:thing-descriptions bindingId="zigbee"
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xmlns:thing="http://eclipse.org/smarthome/schemas/thing-description/v1.0.0"
+	xsi:schemaLocation="http://eclipse.org/smarthome/schemas/thing-description/v1.0.0 http://eclipse.org/smarthome/schemas/thing-description-1.0.0.xsd">
+
+	<thing-type id="xiaomi_lumisensor86sw2">
+		<label>Xiaomi Wireless 2-Button Switch</label>
+		<description>Xiaomi Wireless 2-Button Switch</description>
+
+		<channels>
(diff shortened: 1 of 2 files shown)


puzzle-star:
Right. It reports the wrong set of clusters, but suports the (non reported) switch cluster

Ok to document in the XML (I assume there is a way it can be seen in the GUI). Will change it.

cdjackson:
When the descriptors are retrieved, they will be shown in the logs - you can se then what clusters are supported by the endpoint (assuming they are reported in the descriptor - which is not always the case with these devices!).

t-8ch:
@cdjackson , I know about these. Afaik I have to subscribe to these clusters to receive their commands.
Is there a way of subscribing all clusters while developing a new config. Then I could press the button a look at the zigbee debug log to see which command was actually sent.
(This is how I found out about the attribute reports of the other switch).
So I don't have to try each cluster own-by-one.

cdjackson:
Ok, you asked if the logging showed the clusters - so this was the answer to that question ;)

No - you can’t just subscribe (ie bind) to all clusters. A device will have a limit to the number of clusters that can be bound in its binding table - on some devices, this may be limited to a single entry (some Hue bulbs are like this) - other devices will allow 2 or 3, and some will allow many more. Subscribing to all clusters will likely cause problems, and then you also need to enable reporting.

puzzle-star:
Hi @hsudbrock and @cdjackson 

> Ok to document in the XML (I assume there is a way it can be seen in the GUI). Will change it.

done :-)

cdjackson:
Thanks @puzzle-star - but - I'm not sure this is the best approach. 

Do we really think that users need to know about this? Personally, I don't think so - I would personally suggest to put this as a comment in the XML rather than putting it into the thing type description. Ideally, users don't need to know if this is a static thing type, or dynamic, or the reason why it's one or the other - they just need to know it works and I think this should not be provided to users.

If @hsudbrock thinks that users need to know this information, then maybe I can be convinced, but my preference is not to have this displayed in the UI.

puzzle-star:
Hi @cdjackson,

> Ideally, users don't need to know if this is a static thing type, or dynamic, or the reason why it's one or the other - they just need to know it works and I think this should not be provided to users.

I tend to agree, but also I am seing the "add manually list" starting to grow, with lots of devices that cannot really be added manually, maybe a "Not to be added manually" note on a newline should anyway be added?

cdjackson:
We can fix this - there is an XML flag that can be added to stop it appearing in the list - I agree that we should add this (as we do in ZWave). I’ll dig it out...

cdjackson:
If we add ```listed="false"``` to the thing-type definition, it will not show up in the list -:

```
  <thing-type id="popp_700045_00_000" listed="false">
```

I actually suggest that we do this for all static thing types - I meant to start doing this a while back, but forgot...

puzzle-star:
OK. I think that approach is better, but I would suggest then to leave the comment in the description in a newline, spearated by "<br/>" (according to the thing xml documentation, this should be the way to have GUIs showing the first line in short representations).

I would suggest to keep the description separated by a BR, and adding the listed="false"

What do you thing?

cdjackson:
Why do you think we need this in the description?  Personally, I think it should be a comment in the XML - not part of the thing description. This is an implementation issue, and I don’t think the user should care about this.

puzzle-star:
OK, then, will leave it in a comment in the XML, out of the description

cdjackson:
Thanks.

puzzle-star:
Done. Once this is merged, I will do a PR to add the listed="false" to the other static thing definitions, to clean up the list

cdjackson:
Thanks Pedro.

hsudbrock:
> If @hsudbrock thinks that users need to know this information, then maybe I can be convinced, but my preference is not to have this displayed in the UI.

No, my suggestion above was also to merely add a comment to the XML - sorry if what I wrote was ambiguous.

Thanks for adding this thing type!

puzzle-star:
Hi @hsudbrock ,

Already discussed and agreed with @cdjackson  (and fixed) :-)

PR #381, opened 2019-02-05
Allowing integrators of the binding to implement an own thing handler factory 

hanneshofmann:
Integrators of the binding require to provide own thing types. For this purpose, https://github.com/openhab/org.openhab.binding.zigbee/pull/365 introduced an API (`ZigBeeThingTypeProvider`) which is further used by the `ZigBeeHandlerFactory` to create the corresponding thing handler using additionally the thing types of integrators.

However, it turned out that this approach won't work well as `ZigBeeHandlerFactory#supportsThingType` might be called *before* an external `ZigBeeThingTypeProvider` got bound. In consequence, the respective things remain uninitialized (e.g. after runtime start/bundle update).

To overcome this, the following changes were done in this PR:
- Removed the `ZigBeeThingTypeProvider` API and the no longer needed package
- Introduced an API for the `ZigBeeChannelConverterFactory`

Based on the `ZigBeeChannelConverterFactory` API, integrators are no longer required to expose a `ZigBeeThingTypeProvider` component as they are able to implement their own thing handler factories.

@cdjackson I would be very happy to hear your opinion on that topic.


hsudbrock:
Hi @hanneshofmann - do I understand it correctly that this PR - in summary - does the following two things?
1. Revert #365 
2. Extract an interface for the `ZigBeeChannelConverterFactory` which is in an OSGi-exported package (and rename the existing implemenation to `ZigBeeChannelConverterFactoryImpl`)

hanneshofmann:
@hsudbrock Yes (and the MR also reverts https://github.com/openhab/org.openhab.binding.zigbee/pull/370).

PR #383, opened 2019-02-08
Remove deadlock in the reconnect logic of the ZigBeeCoordinatorHandler

hsudbrock:
The ZigBeeCoordinatorHandler reconnect logic can currently run into a deadlock as follows:

1. In thread A, the reconnect Job (in `startReconnectJobIfNotRunning`) calls `initializeZigBee`.
2. The call to `initializeZigBee starts another thread B that initializes the ZigBee stack.
3. The initialization in thread B fails, resulting in `networkStateUpdated` being called with new network state `OFFLINE` in thread B. In consequence, thread B notifies the `reconnectLock` at the end of the `networkStateUpdated` method.
4. Thread A waits on the reconnectLock at the end of the `run` method within `startReconnectJobIfNotRunning`.

Because the `notify` call happens before the `wait` call, and the `wait` call has no timeout, the `wait` call blocks forever. I could observe this deadlock "in the wild" on a rather slow machine.

This PR removes the deadlock by introducing a boolean flag that is set in addition to notifying the `reconnectLock`, and testing this flag before starting to wait on the `reconnectLock`.

Additionally, this PR removes the superfluous notification of the `reconnectLock` when the network state changes to `UNINITIALIZED` (which could lead to one reconnect attempt starting while the previous one has not yet finished - which might be only a theoretical issue, but I think it is good to have it fixed anyways).

Signed-off-by: Henning Sudbrock <henning.sudbrock@telekom.de>

### DIFF
--- a/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeCoordinatorHandler.java
+++ b/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeCoordinatorHandler.java
@@ -125,6 +125,7 @@ public abstract class ZigBeeCoordinatorHandler extends BaseBridgeHandler
     private ScheduledFuture<?> reconnectPollingTimer;
     private ScheduledExecutorService reconnectPollingScheduler;
     private final Object reconnectLock = new Object();
+    private boolean currentReconnectAttemptFinished = false;
 
     /**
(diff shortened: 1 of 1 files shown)


hsudbrock:
@triller-telekom Maybe you could have a look here as well, as it is a change to the reconnect logic.

hsudbrock:
Thanks Chris!

PR #384, opened 2019-02-08
HEX prefix '0x' added to thing's versions.

YordanDZhelev:
This change is needed in order ESH firmware update mechanism to be able to distinguish between HEX and DEC version formats.


hsudbrock:
@cdjackson Is there anything from your side that speaks against adding the `0x` prefix to the firmware version property (as this property contains the firmware version written down in hexadecimal)?

cdjackson:
I'm not sure I fully understand why this is needed?

In any case, please can you run the formatter on the  code (or manually sort it out).

cdjackson:
Sorry - I missed the comments from @hsudbrock before I replied. As I said earlier, I'm not sure why this is needed - isn't it just a case of ensuring that the firmware provider is providing versions in a consistent way to how the binding provides them?

If we make this change, it will mean other firmware providers need to be changed to make this consistent again.

YordanDZhelev:
When ESH compares versions [here](https://github.com/eclipse/smarthome/blob/master/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/firmware/FirmwareImpl.java#L315) it relies on the following regex to check if the version part is an intiger `"^-?\\d+$"` and this will not match, e.g. `0000000f`, however, it wil "false" match e.g. `0000012`, because it is a DEC integer, but the actual DEC value is not 12 but 18. After the chek for integer ESH uses `Integer.parseInt(versionString)` which will not work for `a,b,c,d,e,f` in the version and will return false values for the rest HEX strings, which does not contain letters, but are still in HEX format. In our case should be `Integer.parseInt(versionString, 16)`. The solution in ESH will be to first check for the `0x` prefix, then parse it as HEX.

cdjackson:
Without following this code through to understand it better, I don't really understand why this is needed.  

What I'm worried about is that this is a breaking change that will stop other firmware providers working - ie if the firmware versions are currently defined in a specific format (ie decimal) and we now change this to hex, will that cause a problem with existing implementations?

hsudbrock:
> What I'm worried about is that this is a breaking change that will stop other firmware providers working - ie if the firmware versions are currently defined in a specific format (ie decimal) and we now change this to hex, will that cause a problem with existing implementations?

I think that it would be ok to add the `0x` prefix as suggested by @YordanDZhelev , let me explain why:

To my knowledge, the point where the thing property `firmwareVersion` is relevant for the firmware update process implemented in ESH is when the `FirmwareRegistryImpl` determines the set of `Firmware`s for a given thing:
1. All `FirmwareProvider`s are queried for the `Firmware`s that they can provide for a given `Thing`.
2. For each of the obtained `Firmware`s, it is checked via the method `Firmware.isSuitableFor(Thing thing)` whether the firmware is suitable to be installed on the device. A firmware is 'suitable' if (a) it has the required model, (b) the firmware on the thing is "high enough", and (c) the firmware's `FirmwareRestriction` function (of type `Function<Thing,Boolean>` says that it is ok to install the firmware on the thing.

I.e., it could actually be the case that some `FirmwareProvider` takes the value of the `firmwareVersion` property into account: Firstly, when returning firmwares in step 1. above. Secondly, by using the value of the property in a `FirmwareRestriction` function in step 2. above. So your worries justified, Chris.

If I'd have to guess, however, I'd say that it is rather unlikely that some `FirmwareProvider` actually decides which `Firmware`s to provide for a thing based on the thing's firmware version outside of Qivicon (for which we added this possibility a couple of months ago).

On the other side, I think that it is actually not so good to store the firmware version without the `0x` prefix in the thing property. Firstly, if the version does not contain one of the letters `A`-`F`, it remains unclear that the number is hexadecimal and not decimal. Secondly, ESH checks whether a firmware is "high enough" to be installed on a thing by comparing the firmware's so-called 'prerequisite version' to the thing's firmware property (in step 2.b above). This comparison first tries to parse the versions as decimal numbers and compares those; if this fails, it compares the versions lexicographically. I think that this should actually be OK (if both versions do not contain `A`-`F` the comparison of the parsed numbers will be OK; if one of the versions contains `A`-`F` the comparison is lexicographically, which is also OK). But: I think that it is not nice that versions are compared as numbers or as Strings only depending on whether `A`-`F` occurs in a version number.

WDYT, Chris?

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/discovery/ZigBeeNodePropertyDiscoverer.java
```diff
@@ -185,7 +186,7 @@ public void setAlwaysUpdate(boolean alwaysUpdate) {
         if (otaCluster != null) {
             Integer fileVersion = otaCluster.getCurrentFileVersion(Long.MAX_VALUE);
             if (fileVersion != null) {
-                properties.put(Thing.PROPERTY_FIRMWARE_VERSION, String.format("%08X", fileVersion));
+                properties.put(Thing.PROPERTY_FIRMWARE_VERSION, String.format("%s%08X",HEX_PREFIX, fileVersion));
```
hanneshofmann: I am aware that this code will be used as part of the device discovery and whenever the thing initialization takes place. However, the `Thing.PROPERTY_FIRMWARE_VERSION` property will be also modified after successfully updating the device firmware (see `ZigBeeThingHandler#updateFirmware`).

hsudbrock: So what you are saying is that there is still one point in the code missing that needs to be adapted in this PR, right?

hanneshofmann: Yes and therefore, it might probably make sense to extract the constants as well.

hsudbrock: OK, thanks @hanneshofmann ! Regarding the constants, I still think that `0x` is something that does not need a constant - but I am fine with it :)

cdjackson:
Thanks Henning. It’s been a while since I looked at the firmware provider implementations, but I’m worried that this is a breaking change that will cause other people problems. Maybe I need to spend some time to look over this some more as I really don’t want to merge this right now if it will break other implementations.

hsudbrock:
Of course, Chris, I did not want to rush you.

Issue #348, opened 2018-12-25
Bulb not turned on after brightness attribute report.


t-8ch:
### Outline

When turning off a Hue bulb  and waiting for ~1 Minute the bulb will not turn on again, when selecting same brightness level as before turning off. The same scenario works when the bulb is turned on faster again.

#### Working scenario:

* Set bulb to `$COLOR` at `$BRIGHTNESS`.
* `bulb switches to correct settings`
* Turn bulb off.
* `bulb shuts off`
* Do not wait.
* Set bulb to `$COLOR` at `$BRIGHTNESS`.
* `bulb restores its settings`

#### Non-working scenario

* Set bulb to `$COLOR` at `$BRIGHTNESS`.
* `bulb switches to correct settings`
* Turn bulb off.
* `bulb shuts off`
* Wait > 1 min
* Set bulb to `$COLOR` at `$BRIGHTNESS`.
* `bulb stays off`

### Cause

When turning off the bulb, the binding remembers this and normally sends a `MoveToLevelWithOnOffCommand` packet.
However this bulb send periodic (1 minute) attribute updates. When the update is received the binding will store this attribute. When trying to turn on the bulb again the binding [compares the new value to the old value](https://github.com/openhab/org.openhab.binding.zigbee/blob/79dfaec91fcad6ea625c485a4e29e92dd7ff9565/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterColorColor.java#L334) to determine if the command would be a NOOP.
Because the current brightness value as reported by the attribute update is the same as the one set via the command, `updateBrightness()` will not be called and the bulb will stay off.

### Possible fixes

1) Remove the check in `handleCommand()` (Tested locally, seems to work)
1) Only update the color characteristics when the bulb is not off.

(Personally I would prefer to keep the behaviour-changing state synchronization between the binding and the physical devices to a minum and go with number 1)

### Configuration

| Configuration    | Description                                                                    |
|------------------|--------------------------------------------------------------------------------|
| Coordinator used | CC2531 stick                                                                            |
| openHAB version  | 2.4.0
| Hardware         | RPi 3B+                                                                               |
| Memory           |                                                                                |
| Java version     |                                                                                |
| Devices          | Provide all device types and their addresses                                   |

cdjackson:
[...]

PR #386, opened 2019-02-09
Fix color control channel updates

puzzle-star:
Fix color control channel updates (related to issue #348 )

This changes the way attribute changes are reported in relation to current ColorMode and on/off status.

Also eliminated some floating point comparisons by storing the last values for x, y, hue and saturation as integers.

Finally, tests have been changed as when the light goes off the channel sets a HSBType(h,s,0) as state, to be consistent with the default type of the channel and for OH to remember the hue and saturation values even when off.


cdjackson:
@puzzle-star do you think in the ```handleRefresh``` we should also request the colorMode? Refresh is meant to refresh everything, but I'm not sure that this is getting the complete picture (and I would have the same question for @TomTravaglino for the color temperature converter).

puzzle-star:
Hi @cdjackson,

 > do you think in the `handleRefresh` we should also request the colorMode?

Probably yes. Adding it to the PR

Also, I do not have any (Philips) Hue lights to test, but it works with Hue/Saturation lights (I do not have only IKEA XY lights) :-)

Pedro

PR #387, opened 2019-02-12
Add sensible state tags to the new thermostat channels

hsudbrock:
This PR adds state tags to the new thermostat channels as follows:
* Adds the pattern used for the existing temperature channels to the new thermostat temperature channels
* Marks local and outdoor temperature as read-only
* Marks the running mode as read-only


hsudbrock:
Do you also think that it is sensible to add those `<state>` tags to the new thermostat channels, Chris? (Asking as you might have had a reason for not adding them in your PR, but maybe the missing tags also just slipped through...)

PR #394, opened 2019-02-17
Improve readme for first time setup

Polooo2:
This PR improves the documentation to set up a coordinator with the binding for the first time. Took me an afternoon to get started and wanted to improve it for other users.

### DIFF
--- a/org.openhab.binding.zigbee/README.md
+++ b/org.openhab.binding.zigbee/README.md
@@ -8,9 +8,15 @@ The ZigBee binding supports an interface to a wireless ZigBee home automation ne
 
 A ZigBee Coordinator is the network controller, and is therefore the heart of the ZigBee network. It also acts as the trust centre to control security access to the network.
 
-Coordinators need to be installed manually and the serial port and baud rate must be set. These are set to match the configuration that the dongle is in. Should you wish to use a different baud rate than the default speed of the device, you must change the configuration of the dongle using some other, and then configure the binding to match your change. If in doubt, you should leave the settings at their default values which should work in most cases. 
+Coordinators need to be installed manually and the serial port (```zigbee_port```) and baud rate (```zigbee_baud```) must be set. These are set to match the configuration that the dongle is in. Should you wish to use a different baud rate than the default speed of the device (get default baud rate with (Linux) e.g. ```stty < /dev/ttyACM0```) , you must change the configuration of the dongle using some other, and then configure the binding to match your change. If in doubt, you should leave the settings at their default values which should work in most cases. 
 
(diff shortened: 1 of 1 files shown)


org.openhab.binding.zigbee/README.md
```diff
@@ -8,9 +8,15 @@ The ZigBee binding supports an interface to a wireless ZigBee home automation ne
...
 
-Coordinators need to be installed manually and the serial port and baud rate must be set. These are set to match the configuration that the dongle is in. Should you wish to use a different baud rate than the default speed of the device, you must change the configuration of the dongle using some other, and then configure the binding to match your change. If in doubt, you should leave the settings at their default values which should work in most cases. 
+Coordinators need to be installed manually and the serial port (```zigbee_port```) and baud rate (```zigbee_baud```) must be set. These are set to match the configuration that the dongle is in. Should you wish to use a different baud rate than the default speed of the device (get default baud rate with (Linux) e.g. ```stty < /dev/ttyACM0```) , you must change the configuration of the dongle using some other, and then configure the binding to match your change. If in doubt, you should leave the settings at their default values which should work in most cases.
```
cdjackson: I don't understand what this command is? How can it read the dongle baud rate?

Polooo2: I was looking for ways to determine the baud rate of the device as I didn't know where get it from. After some Google research I found this command and got this output on my raspi:

speed 9600 baud; line = 0;
min = 1; time = 1;
-brkint -icrnl -imaxbel
-opost -onlcr
-isig -icanon -iexten -echo -echoe -echok -echoctl -echoke


Maybe I got it wrong here, but using this value I was able to get the coordinator working (before I always had a communication error and a NPE in thingUpdate in the logs).

cdjackson: I will test this tonight, but I can't see how it can work. What dongle were you using? It might work on dongles like Telegesis and XBee that have ASCII protocols, but most use a binary protocol.

Polooo2: I have the Texas Instruments cc2531. What would be another way of getting this info?

Polooo2: I just noticed that I set the baud rate to 9600 in the thing file. If you open the properties of the device in PaperUi the baud rate seems not to be set and in the drop down I can not select 9600 as a value...

cdjackson: Normally this information would be provided in the manual of the device.

cdjackson: This does not work on my machine. I tried with 2 Ember dongles with different firmware, an XBee dongle, a Telegesis dongle, and a CC2531 dongle. In all cases, there was no response after a few minutes.

I'm also a bit surprised that your system said that the CC2531 is running at 9600bd - have you actually tested this? I thought that the standard firmware works at 38400 (at least all the ones I've ever tested use this rate).

Polooo2: I just see the coordinator marked as online in PaperUI as I did not have time to include any device yet. Ok then I will remove this hint and reference the manual of the device.

Polooo2: I changed the note about getting the default baud rate and referenced the manual there. Sry for the inconvenience

PR #400, opened 2019-03-01
Read multiple attributes at once during discovery

### LINKED ISSUE #399: Speedup discovery by reading multiple attributes with single request
Currently, when discovering a new device, the discovery queries several attributes each with a single `Read Attribute` command (see [here](https://github.com/openhab/org.openhab.binding.zigbee/blob/0cd602d8e4e9b675a63e9a9bfbe20c35ee99e9dd/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/discovery/ZigBeeNodePropertyDiscoverer.java#L80)).

Since recently, the ZSS ZigBee library supports reading multiple attributes with a single request (see [here](https://github.com/zsmartsystems/com.zsmartsystems.zigbee/pull/530)).

It would be nice to use this feature for speeding up the discovery process by requesting all the attributes at once (with a fallback to using multiple `Read Attribute` requests for devices that cannot put all attribute values into a single response).

I could implement this, @cdjackson - so in case you already started with this please give me a ping :)

hsudbrock:
This resolves #399 by using a multi-attribute read request in the `ZigBeeNodePropertyDiscoverer`. I did some smaller refactorings in that class as well, and used separate commits for those to ease the review.

Please note that this PR needs the new ZSS version before it can be merged (the new version that supports multi-attribute-read commands).

The basic idea is to do a multi-attribute read first, and then read each attribute separately from the ZSS layer as is already done before this PR. If the multi-attribute read succeeds, then the subsequent separate reading of each attribute will be obtained from the cache; if the multi-attribute read fails, then the subsequent separate readings serve as a fallback.

As a small additional change I added in this PR is adding a thing property for the manufacturer code obtained from the node descriptor (as I think that this is helpful in particular with the upcoming support for manufacturer-specific clusters).


cdjackson:
Replaced by #464

PR #405, opened 2019-03-08
Make interface WarningTypeCommandDescriptionProvider and related enums API

hsudbrock:
Currently, the interface `WarningTypeCommandDescriptionProvider` is not exported. The interface should, however, be exported, because it is intended to be implemented by providers that might be in other bundles.

For this reason, I have moved the interface to the subpackage `warningdevice` of the non-internal `converter` package. Likewise, I have added related enums into the exported subpackage, as they can be used to create warning type command options in a type-safe way.


hsudbrock:
Thanks Chris!

PR #407, opened 2019-03-09
Update README.md

BlackEternity:
Updated the README.md file to add the Philips Hue Dimmer-Switch to the list of supported devices. 
Additionally, a new section has been added - Channels triggered event & rules to give an example on how to utilize Channels that cannot be bound to Items like the Hue Dimmer buttons.


cdjackson:
Thanks - looks fine to me, but as @hsudbrock is the king of the trigger channel, maybe he can take a quick peek to check if it ok as well...

hsudbrock:
Thanks @BlackEternity ! Looks fine for me as well.

Just as an idea, in case you want to improve the docs further, you could think of linking to the openHAB docs for trigger channels (https://www.openhab.org/docs/configuration/rules-dsl.html#channel-based-triggers) in an additional PR.

PR #409, opened 2019-03-11
Allow thing handler to skip device initialization and execute it on config update

### LINKED ISSUE #403: Initialize device only when it is first discovered
With #342, in `ZigBeeBaseChannelConverter` the device initialization and converter initialization has been split into two methods, so that it becomes possible to only initialize the device, or the converter, or both.

This issue is for implementing the following extension suggested by @cdjackson in #338:
> Initially I would simply propose to call both methods during initialisation, however a further extension is to only call configure when the device is first discovered (need to think how this is detected) and it may also be useful to provide a user action to allow the user to manually re-configure the device (which would be especially useful for battery devices).

The basic idea is to introduce a boolean config parameter `zigbee_initialise_device` for things. The `ZigBeeThingHandler` will initialize the device only of this parameter is set to `true`, and it will set the parameter to `false` afterwards. The parameter will be set to `true` by the discovery service, so that the device is always initialized after discovery. Users can manually re-configure the device by updating the value of the parameter to `true`.

TomTravaglino:
Resolves #403

This PR provides a solution that allows the thing handler to skip device initialization if it has been done previously. Furthermore it allows to execute a device (re-)initialization by updating the thing config.

In detail:
- Introduced thing property 'zigbee_initialised'. The property will be
  checked before the channels are initialized and skips the device
  initialization when it is 'true'. On the first initialization the
  property will be set to 'true' after the channel initialization.
- Introduced thing config parameter 'zigbee_initialise_device' in order
  to allow the user to (re-)initialize the device by updating the
  config. The config parameter will be checked when the config is
  updated and executes a device initialization by disposing and initializing the thing handler


TomTravaglino:
@hsudbrock As discussed, please join this PR and do a review.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeThingHandler.java
```diff
@@ -423,6 +431,22 @@ private synchronized void doNodeInitialisation() {
         coordinatorHandler.serializeNetwork();
     }
 
+    private synchronized void reinitializeDevice() {
```
hsudbrock: I do not completely understand the workings of this method. It reinitializes the complete node (via `initialiseZigBeeNode`), and adds some listeners, but does not uninitialise everything before (e.g., the channel converters are not disposed). Why is this OK?

I had thought that one could either dispose all channel converters and them reinitialize them (including device initialization), or that one could dispose the complete thing handler and reinitialize ist.

WDYT?

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/ZigBeeBindingConstants.java
```diff
@@ -218,6 +218,7 @@
     public static final String THING_PROPERTY_ASSOCIATEDDEVICES = "zigbee_devices";
     public static final String THING_PROPERTY_INSTALLCODE = "zigbee_installcode";
     public static final String THING_PROPERTY_STACKCOMPLIANCE = "zigbee_stkcompliance";
+    public static final String THING_PROPERTY_INITIALIZED = "zigbee_initialised";
```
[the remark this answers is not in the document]
TomTravaglino: Sound good to me

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeThingHandler.java
```diff
@@ -582,6 +608,9 @@ public void handleConfigurationUpdate(Map<String, Object> configurationParameter
                 case ZigBeeBindingConstants.CONFIGURATION_LEAVE:
                     coordinatorHandler.leave(nodeIeeeAddress);
                     break;
+                case ZigBeeBindingConstants.CONFIGURATION_INITIALIZE_DEVICE:
```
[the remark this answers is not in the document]
TomTravaglino: OK. I will do the change to the entire switch statement.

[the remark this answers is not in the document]
TomTravaglino: Or maybe to all usages of ZigBeeBindingConstants in this class? wdyt?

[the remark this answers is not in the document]
hsudbrock: Both fine for me. (I, personally, would stick to the changes relevant to this PR.)

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeThingHandler.java
TomTravaglino: You're right, something went totally wrong here. Will change it to this:
```
    private synchronized void reinitializeDevice() {
        dispose();
        initialize();
    }
```

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeThingHandler.java
```diff
@@ -350,6 +352,8 @@ private synchronized void doNodeInitialisation() {
                 updateThing(thingBuilder.build());
             }
 
+            Boolean initializeDevice = !Boolean.valueOf(thing.getProperties().get(THING_PROPERTY_INITIALIZED));
```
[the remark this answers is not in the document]
TomTravaglino: Not really. Can change it.

TomTravaglino:
@hsudbrock I updated the PR

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeThingHandler.java
```diff
@@ -364,7 +369,7 @@ private synchronized void doNodeInitialisation() {
                 }
 
                 logger.debug("{}: Initializing channel {} with {}", nodeIeeeAddress, channel.getUID(), handler);
-                if (handler.initializeDevice() == false) {
+                if (initializeDevice && handler.initializeDevice() == false) {
```
hsudbrock: Suggestion: Add a log statement somewhere that device initialization is skipped, because the device is already initialized. WDYT?

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeThingHandler.java
```diff
@@ -423,6 +432,11 @@ private synchronized void doNodeInitialisation() {
         coordinatorHandler.serializeNetwork();
     }
 
+    private synchronized void reinitializeDevice() {
+        dispose();
```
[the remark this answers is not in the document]
TomTravaglino: 👍

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeThingHandler.java
```diff
@@ -8,6 +8,8 @@
  */
 package org.openhab.binding.zigbee.handler;
 
+import static org.openhab.binding.zigbee.ZigBeeBindingConstants.*;
```
cdjackson: Personally I don't like static imports - especially importing all constants in this way as I think it's prone to error. Is this something DT do as standard? My preference is to define the class name with the constants - I know it's arguably less readable, but IMHO it's more understandable, and less prone to error.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeThingHandler.java
```diff
@@ -394,6 +399,10 @@ private synchronized void doNodeInitialisation() {
                     stateDescriptions.put(channel.getUID(), stateDescription);
                 }
             }
+
+            if (initializeDevice) {
```
cdjackson: Why is there the need for the ```if``` statement here? For the sake of simply writing the property on every startup (not a high frequency operation) you could remove this and just update the property?

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeThingHandler.java
```diff
@@ -394,6 +399,10 @@ private synchronized void doNodeInitialisation() {
...
+                thing.setProperty(THING_PROPERTY_DEVICE_INITIALIZED, Boolean.TRUE.toString());
```
cdjackson: It seems possible to incorrectly report that the device is initialised when it isn't?

This statement will always set the device_initialised to true - even if the device was not actually initialised. By this, I mean if there is an error above, you ignore the error and we still say the device is initialised.

Personally, I think that if the user, or system, wanted to initialise the device, and it fails, then we should not set this property to true. This would allow it to retry on next startup rather than leave the device unusable.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeThingHandler.java
hsudbrock: Not a standard at DT :) I, personally, am fine with both, as long as it is consistent within the class... So if you prefer not using static imports, fine with me, Chris.

TomTravaglino: OK. I will remove the static imports and use the class name instead.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeThingHandler.java
TomTravaglino: Will remove the `if` statement

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeThingHandler.java
TomTravaglino: You're right. I will change it and set the property to `true` only if all channels are initialized successfully.

hsudbrock:
@TomTravaglino Should the new config parameter also be added to the config description? (Probably as an advanced parameter?)

TomTravaglino:
> Should the new config parameter also be added to the config description

Yes. I add it to `device.xml`

hsudbrock:
> And to the static thing types too?

What would be the reason for *not*  adding it to the static thing types?

cdjackson:
Regarding this config parameter - this could get messy. I think that without this in the XML (ie ALL xmls for all things) then there might be problems. Certainly it will not be possible for users to reinitialise their devices.

I wonder if it might not be better to instead use a Config Parameter Provider to provide this to all ZigBee thing types rather than require it to be added to all XML files. This might be less error prone?

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeThingHandler.java
```diff
@@ -582,6 +605,9 @@ public void handleConfigurationUpdate(Map<String, Object> configurationParameter
                 case ZigBeeBindingConstants.CONFIGURATION_LEAVE:
                     coordinatorHandler.leave(nodeIeeeAddress);
                     break;
+                case ZigBeeBindingConstants.CONFIGURATION_INITIALIZE_DEVICE:
+                    initializeDevice = true;
```
TomTravaglino: I'm wondering if I need to check the value of the config param instead of just setting the variable to `true`

cdjackson: I don't see why? 
If it's previously false, you set it to true so it initialises as requested. If it was previously true, then you will still initialise as requested.

TomTravaglino:
> Regarding this config parameter - this could get messy. I think that without this in the XML (ie ALL xmls for all things) then there might be problems. Certainly it will not be possible for users to reinitialise their devices.
> 
> I wonder if it might not be better to instead use a Config Parameter Provider to provide this to all ZigBee thing types rather than require it to be added to all XML files. This might be less error prone?

@cdjackson

cdjackson:
Yes

TomTravaglino:
I discussed the ‘Config Parameter Provider’ idea with @triller-telekom and we came to the conclusion that such a solution needs ‘some’ work to be done. In addition, the mac address is currently also defined in the XML. My PR implements that the reportings and cluster bindings are only configured once for each device. Implementing a generic provider for various kinds of thing properties is IMHO out of scope for this PR. Our proposal is to let this PR as it is and create a new issue for the ‘Config Parameter Provider’ solution. WDYT?

cdjackson:
> we came to the conclusion that such a solution needs ‘some’ work to be done.

Yes, but it's not much is it? And it makes this PR usable - otherwise this PR leaves hidden dependencies.

> Implementing a generic provider for various kinds of thing properties is IMHO out of scope for this PR.

I disagree since without this, we end up with a dependency that will probably be forgotten, and will stop devices working.  What I'm worried about is if we implement these sort of changes that are only partly working, it's fine for Qivicon where Deutsche Telekom people can manage your system and your thing definitions, but it will likely be left to me to work out the problems with openHAB.

I'm happy to move it to another PR if you plan to provide that in the coming couple of weeks, otherwise I would prefer to see it here so that this feature doesn't have hidden dependencies.

hsudbrock:
I just discussed this issue with @triller-telekom. First, we see Chris' point that it might cause issues adding this to all static thing types; in particular, if there is someone out there having own static thing types that are not provided by the open source binding, then these would not get the new config parameter.

We took a look at what is currently implemented in the `ZigBeeThingHandler`; it already acts as a `ConfigDescriptionProvider`, providing additional config parameters for some things with the generic thing type. We could simply have this `ConfigDescriptionProvider` also provide this PR's new parameter `zigbee_initialise_device`. This appears (a) simpler than adding new `config parameter providers` as discussed above (as this is not yet provided by the openHAB core bundles, and would hence need more work), and (b) would nicely fit into the existing code regarding dynamically provided config parameters.

From the code perspective, this could look like this in the `ZigBeeThingHandler.doNodeInitialisation` (instead of the current code [here](https://github.com/openhab/org.openhab.binding.zigbee/blob/8276ff250a72c931bebd44a6da23865ca2088413/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeThingHandler.java#L254-L279)), where we add the new config parameter for all things:

```
        List<ConfigDescriptionParameter> parameters = new ArrayList<ConfigDescriptionParameter>();
        parameters.add(ConfigDescriptionParameterBuilder.create("zigbee_initialise_device", Type.BOOLEAN)
                .withAdvanced(true).withDescription("Set to true to reinitialise binds and reporting on the device")
                .withLabel("Reinitialise device").withRequired(false).build());

        if (getThing().getThingTypeUID().equals(ZigBeeBindingConstants.THING_TYPE_GENERIC_DEVICE)) {
            // Dynamically create the channels from the device
            // Process all the endpoints for this device and add all channels as derived from the supported clusters
            nodeChannels = new ArrayList<>();
            for (ZigBeeEndpoint endpoint : coordinatorHandler.getNodeEndpoints(nodeIeeeAddress)) {
                logger.debug("{}: Checking endpoint {} channels", nodeIeeeAddress, endpoint.getEndpointId());
                nodeChannels.addAll(channelFactory.getChannels(getThing().getUID(), endpoint));
            }
            logger.debug("{}: Dynamically created {} channels", nodeIeeeAddress, nodeChannels.size());

            for (ZclClusterConfigHandler handler : configHandlers) {
                parameters.addAll(handler.getConfiguration());
            }
        } else {
            // We already have the correct thing type so just use the channels
            nodeChannels = getThing().getChannels();
            logger.debug("{}: Using static definition with existing {} channels", nodeIeeeAddress, nodeChannels.size());
        }

        try {
            configDescription = new ConfigDescription(new URI("thing:" + getThing().getUID()), parameters);
        } catch (URISyntaxException e) {
            logger.debug("Error creating URI for thing description:", e);
        }
```

WDYT, @cdjackson and @TomTravaglino ? Would this meet both your concerns?

TomTravaglino:
I considered a very similar approach this morning. But to me it lookes like it works for generic devices since you create a new config description and put it into the device. But it was not yet clear to me if you can handle it that way when a static thing type already provides its own config description. I think this approach would overwrite it. Maybe I missed something so I first wanted to discuss it with someone of you.

hsudbrock:
> I think this approach would overwrite it.

The `ConfigDescriptionRegistry` merges the config description provided by a thing type with the config descriptions provided by `ConfigDescriptionProvider`s (in our case, the `ZigBeeThingHandler`).  (See the implementation of `ConfigDescriptionRegistry.getConfigDescription(URI uri, Locale locale)` for details.) I.e., it is not overwritten.

TomTravaglino:
Thanks for the input. Then the approach should be fine.

TomTravaglino:
I updated the PR

TomTravaglino:
PR is updated

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeThingHandler.java
```diff
@@ -251,6 +252,9 @@ private synchronized void doNodeInitialisation() {
 
         List<Channel> nodeChannels;
 
+        List<ConfigDescriptionParameter> parameters = new ArrayList<ConfigDescriptionParameter>();
+        parameters.addAll(ZigBeeConfigDescriptionParameters.getParameters());
```
hsudbrock: Why do you add those explicitly here, and not put them directly into the array list constructor?

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeThingHandler.java
```diff
@@ -363,10 +369,17 @@ private synchronized void doNodeInitialisation() {
...
+                if (initializeDevice) {
+                    logger.debug("{}: Initializing channel {} with {}", nodeIeeeAddress, channel.getUID(), handler);
+                    if (handler.initializeDevice() == false) {
+                        logger.info("{}: Channel {} failed to initialise device", nodeIeeeAddress, channel.getUID());
+                        channelInitializationSuccessful = false;
+                        continue;
+                    }
+                } else {
+                    logger.debug(
+                            "{}: Initialization of channel {} with {} will be skipped as the device is already initialized",
```
hsudbrock: This log message sounds wrong, as the channel is still initialized (but only the converter part - just the device initialization part is skipped).

Correct would be something like "Device initialization for channel {} with {} is skipped as the device is already initialized".

IMHO, it would be even better to not log this for every channel, but rather use a single log message stating that device initialization is skipped for all channels (in the case that `initializeDevice` is false).

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeThingHandler.java
```diff
@@ -423,6 +439,12 @@ private synchronized void doNodeInitialisation() {
         coordinatorHandler.serializeNetwork();
     }
 
+    private synchronized void reinitializeDevice() {
```
hsudbrock: Reading this now, the method name seems wrong. This method reinitializes the thing handler, and not the device. In fact, it will only reinitialize the device if the `THING_PROPERTY_DEVICE_INITIALIZED` was set to `false` before (which you do in `handleConfigurationUpdate`).

I suggest to either rename the method, or set the thing property in here, to fix this issue.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeThingHandler.java
```diff
@@ -597,6 +624,11 @@ public void handleConfigurationUpdate(Map<String, Object> configurationParameter
 
         // Persist changes
         updateConfiguration(configuration);
+
+        if (initializeDevice) {
```
hsudbrock: Question: What is the difference for you between `initializeDevice` and `reinitializeDevice`? (If there is none, you could use one of the words both for the variable and the method.)

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeThingHandler.java
```diff
@@ -251,6 +252,9 @@ private synchronized void doNodeInitialisation() {
 
         List<Channel> nodeChannels;
 
+        List<ConfigDescriptionParameter> parameters = new ArrayList<ConfigDescriptionParameter>();
```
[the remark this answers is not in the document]
TomTravaglino: I will change it.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/ZigBeeConfigDescriptionParameters.java
```diff
@@ -0,0 +1,41 @@
...
+
+import org.eclipse.smarthome.config.core.ConfigDescriptionParameter;
+import org.eclipse.smarthome.config.core.ConfigDescriptionParameterBuilder;
+
+/**
+ * Provides {@link ConfigDescriptionParameter}s which are supposed to exist in all things
+ */
+public class ZigBeeConfigDescriptionParameters {
+
+    private static final String PARAM_ZIGBEE_INITIALISE_DEVICE_NAME = "zigbee_initialise_device";
```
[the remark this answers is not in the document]
TomTravaglino: Will replace it.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeThingHandler.java
TomTravaglino: No special reason. But I can change it.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeThingHandler.java
TomTravaglino: I can remove this statement and put this one just before the loop:
```
            if (!initializeDevice) {
                logger.debug("{}: Device initialization will be skipped as the device is already initialized",
                        nodeIeeeAddress);
            }
```

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeThingHandler.java
TomTravaglino: OK. I will move the statement
```
getThing().setProperty(ZigBeeBindingConstants.THING_PROPERTY_DEVICE_INITIALIZED, Boolean.FALSE.toString());
```
into this method. I think then it's clear what the method is made for.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeThingHandler.java
TomTravaglino: I named the variable `initializeDevice` because it is related to the config param `zigbee_initialise_device`. The method name is `reinitializeDevice` because a re-initialization is what it does when the config param is set to `true`.
Do you think it's confusing or not good to understand?

hsudbrock: > Do you think it's confusing or not good to understand?

I think that it is ok, but not ideal.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeThingHandler.java
```diff
@@ -423,6 +439,12 @@ private synchronized void doNodeInitialisation() {
         coordinatorHandler.serializeNetwork();
     }
 
+    private synchronized void reinitializeDevice() {
+        logger.debug("{}: Reinitializing device", nodeIeeeAddress);
+        dispose();
+        initialize();
```
cdjackson: This seems a bit strange - maybe this is also Hennings issue.

This method ```reinitializeDevice``` doesn't reinitialise the device - it reinitialises the thing handler and the device. I would have expected that when the user wants to reinitialise the device, it just does that.

Or, do you think that because we have updated the device, that we also need to reinitialise all the converters? I'm not sure this is the case (I'd need to think it through some more - maybe you've already done that?). The clusters won't change, so I don't think reinitialising the converters and redetecting the channels and polling is really necessary?

I could be wrong though, and maybe a full reinitialisation and repoll etc is really needed, but this wasn't my expectation when I first introduced this concept.

TomTravaglino: The idea was to re-use the existing methods and so rely on the existing initialization. However, I understand your objection and so I took a closer look at the initialization part. As far as I can see we can initialize the device without the need to re-initialize anything else. I'm implementing a method that initializes only the devices. PR will follow soon.

TomTravaglino:
I update the PR. Now the device can be re-initialized without initializing the thing handler.

cdjackson:
Thanks @TomTravaglino this looks good to me. I'll just let CI re-run after the branch update and then it should be good to merge.

triller-telekom:
@cdjackson Pipeline looks successful now, I guess its ready to be merged now. Thanks for reviewing.

Issue #315, opened 2018-11-30
ONLINE/OFFLINE status on Coordinator and Zigbee devices


Hambootje:
I used both the CC2531EMK and Ember EM35x Coordinator via the Zigbee binding. 
When a zigbee lightbulb is switched off from the main power, the status of the thing remains online.
The same holds for the coordinator itself. This also show always online status, even if the dongle is removed.

For light bulbs this is not really a problem... You know when they don't work properl. It's different with for example doorsensors. Here you want to keep track that the device is still present and fire a notification when the thing goes offline.

From the log-file I can see that a regular polling cycle is started. If the device is online, the items are updated. If the zigbee-device is offline there are no values returned. I guess in this case the thing status can be set to OFFLINE.

### Configuration
Coordinator | Ember EM35x or CC2531EMK
openHAB version  | 2.4.0.SNAPSHOT
Hardware         |  Raspberry Pi 3+                                                                          
Java version     |    openjdk version "1.8.0_152"                                                                           
Devices            | zigbee:device:013C1ADE:000b57fffe909e96:000B57FFFE909E96_1_dimmer

cdjackson:
This is not super easy to resolve. For lightbulbs it is as they are mains powered, however for battery devices, most will not respond to requests, and therefore it is not easy (well, not possible even!) to decide if they are online or not.

Mains devices could be set offline after the polling period (or probably best to be 2x the polling period, or possibly 2x the reporting period). Battery devices I think should not be set offline unless we know that they are the sort that respond when asleep, which is a very short list of devices in my experience.

triller-telekom:
I would like to get back on this topic to improve the status handling of zigbee devices in openHAB.

> Mains devices could be set offline after the polling period (or probably best to be 2x the polling period, or possibly 2x the reporting period)

I agree, since devices have to poll the coordinator within that polling period. You are referring to 
EZSP_CONFIG_END_DEVICE_POLL_TIMEOUT and EZSP_CONFIG_END_DEVICE_POLL_TIMEOUT_SHIFT which are also mentioned here, right?

https://www.silabs.com/community/wireless/zigbee-and-thread/knowledge-base.entry.html/2012/06/29/how_long_can_my_end-jseP

> Battery devices I think should not be set offline unless we know that they are the sort that respond when asleep, which is a very short list of devices in my experience.

This is a certainly a problem if some devices are faulty. But should that prevent us from setting other devices to offline?

[...]

PR #412, opened 2019-03-14
Implemented IsAliveTracker to set Things to OFFLINE after a timeout

triller-telekom:
- Introduced a method in the ZigbeeBaseChannelConverter to obtain the minimum
of all configured max reporting intervals.
- Adjusted wrong comments with 10 minutes where code said 2 hours.
- Use pollingPeriod = POLLING_PERIOD_DEFAULT if there was no blocking get()
on the Future before this PR and POLLING_PERIOD_HIGH if it was previously
used in the case that the configuring of the reporting failed.
- Using !CommandResult.isSuccess() instead of sometimes isError() for
all converters.

Relates to #315

### DIFF
--- a/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/converter/ZigBeeBaseChannelConverter.java
+++ b/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/converter/ZigBeeBaseChannelConverter.java
@@ -152,6 +152,12 @@ public abstract class ZigBeeBaseChannelConverter {
     protected int pollingPeriod = POLLING_PERIOD_DEFAULT;
 
     /**
+     * The smallest of all maximum reporting periods configured for the attributes used by the converter. By default it
+     * is set to {@code Integer.MAX_VALUE}, because we assume that not every converter implements reporting.
+     */
+    protected int minimalReportingPeriod = Integer.MAX_VALUE;
+
+    /**
      * Constructor. Creates a new instance of the {@link ZigBeeBaseChannelConverter} class.
      *
      */
@@ -162,11 +168,11 @@ public abstract class ZigBeeBaseChannelConverter {
     /**
      * Creates the converter handler.
      *
-     * @param thing the {@link ZigBeeThingHandler} the channel is part of
-     * @param channel the {@link Channel} for the channel
+     * @param thing       the {@link ZigBeeThingHandler} the channel is part of
+     * @param channel     the {@link Channel} for the channel
      * @param coordinator the {@link ZigBeeCoordinatorHandler} this endpoint is part of
-     * @param address the {@link IeeeAddress} of the node
-     * @param endpointId the endpoint this channel is linked to
+     * @param address     the {@link IeeeAddress} of the node
+     * @param endpointId  the endpoint this channel is linked to
      */
     public void initialize(ZigBeeThingHandler thing, Channel channel, ZigBeeCoordinatorHandler coordinator,
             IeeeAddress address, int endpointId) {
@@ -307,10 +313,41 @@ public abstract class ZigBeeBaseChannelConverter {
     }
 
     /**
+     * Gets the minimum of all maximum reporting periods used in attribute reports. If no attribute reports are
+     * configured, the default {@code Integer.MAX_VALUE} will be returned.
+     *
+     * @return minimum of all maximum reporting periods in seconds
+     */
+    public int getMinimalReportingPeriod() {
+        return minimalReportingPeriod;
+    }
+
+    /**
+     * Sets the {@code pollingPeriod} and {@code maxReportingPeriod} depending on the success or failure of the given
+     * reporting response.
+     *
+     * @param reportResponse                    a {@link CommandResult} representing the response to a reporting request
+     * @param reportingFailedPollingInterval    the polling interval to be used in case configuring reporting has
+     *                                              failed
+     * @param reportingSuccessMaxReportInterval the maximum reporting interval in case reporting is successfully
+     *                                              configured
+     */
+    protected void handleReportingResponse(CommandResult reportResponse, int reportingFailedPollingInterval,
+            int reportingSuccessMaxReportInterval) {
+        if (!reportResponse.isSuccess()) {
+            // we want the minimum of all pollingPeriods
+            pollingPeriod = Math.min(pollingPeriod, reportingFailedPollingInterval);
+        } else {
+            // we want to know the minimum of all maximum reporting periods to be used as a timeout value
+            minimalReportingPeriod = Math.min(minimalReportingPeriod, reportingSuccessMaxReportInterval);
+        }
+    }
+
+    /**
      * Creates a standard channel UID given the {@link ZigBeeEndpoint}
      *
-     * @param thingUID the {@link ThingUID}
-     * @param endpoint the {@link ZigBeeEndpoint}
+     * @param thingUID    the {@link ThingUID}
+     * @param endpoint    the {@link ZigBeeEndpoint}
      * @param channelName the name of the channel
      * @return
      */
@@ -358,7 +395,7 @@ public abstract class ZigBeeBaseChannelConverter {
      * The currentConfiguration shall be updated.
      *
      * @param currentConfiguration the current {@link Configuration}
-     * @param updatedParameters a map containing the updated configuration parameters to be set
+     * @param updatedParameters    a map containing the updated configuration parameters to be set
      */
     public void updateConfiguration(@NonNull Configuration currentConfiguration,
             Map<String, Object> updatedParameters) {
--- a/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeThingHandler.java
+++ b/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeThingHandler.java
@@ -129,6 +129,15 @@ public class ZigBeeThingHandler extends BaseThingHandler implements ZigBeeNetwor
     private final int POLLING_PERIOD_DEFAULT = 1800;
     private int pollingPeriod = POLLING_PERIOD_DEFAULT;
 
+    /**
+     * We increase the timeout interval by this factor to allow for lost data
+     */
+    private final int POLLING_OR_REPORTING_FACTOR = 2;
+    /**
+     * Increase the timeout by a margin to capture the case where data arrives shortly after the specified interval
+     */
+    private final int POLLING_OR_REPORTING_MARGIN = 30;
+
     private boolean firmwareUpdateInProgress = false;
 
     private ExecutorService commandScheduler = ThreadPoolManager.getPool("zigbee-thinghandler-commands");
@@ -145,14 +154,23 @@ public class ZigBeeThingHandler extends BaseThingHandler implements ZigBeeNetwor
     private final ZigBeeChannelConverterFactory channelFactory;
 
     /**
+     * The service with timers to see if the device is still alive (ONLINE)
+     */
+    private final ZigbeeIsAliveTracker isAliveTracker;
+
+    /**
      * Creates a ZigBee thing.
      *
-     * @param zigbeeDevice   the {@link Thing}
-     * @param channelFactory the {@link ZigBeeChannelConverterFactory} to be used to create the channels
+     * @param zigbeeDevice         the {@link Thing}
+     * @param channelFactory       the {@link ZigBeeChannelConverterFactory} to be used to create the channels
+     * @param zigbeeIsAliveTracker the tracker which sets the {@link Thing} to OFFLINE after a period without
+     *                                 communication
      */
-    public ZigBeeThingHandler(Thing zigbeeDevice, ZigBeeChannelConverterFactory channelFactory) {
+    public ZigBeeThingHandler(Thing zigbeeDevice, ZigBeeChannelConverterFactory channelFactory,
+            ZigbeeIsAliveTracker zigbeeIsAliveTracker) {
         super(zigbeeDevice);
         this.channelFactory = channelFactory;
+        this.isAliveTracker = zigbeeIsAliveTracker;
     }
 
     @Override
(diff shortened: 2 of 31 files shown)


triller-telekom:
close/reopen to trigger new build

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/converter/ZigBeeBaseChannelConverter.java
```diff
@@ -306,11 +312,42 @@ public int getPollingPeriod() {
         return pollingPeriod;
     }
 
+    /**
+     * Gets the minimum of all maximum reporting periods used in attribute reports. If no attribute reports are
+     * configured, the default {@code Integer.MAX_VALUE} will be returned.
+     *
+     * @return minimum of all maximum reporting periods in seconds
+     */
+    public int getMinimalMaximumReportingPeriod() {
```
cdjackson: This is  a horrible name - can't it just be ```getMinimumReportingPeriod```

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/converter/ZigBeeBaseChannelConverter.java
```diff
@@ -151,6 +151,12 @@
      */
     protected int pollingPeriod = POLLING_PERIOD_DEFAULT;
 
+    /**
+     * The smallest of all maximum reporting periods configured for the attributes used by the converter. By default it
+     * is set to {@code Integer.MAX_VALUE}, because we assume that not every converter implements reporting.
+     */
+    protected int minimalMaximumReportingPeriod = Integer.MAX_VALUE;
```
cdjackson: This may need to be aligned with the reporting configuration?
Also, as per other comments - I don't really think this is a very descriptive name.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeThingHandler.java
```diff
@@ -129,6 +129,8 @@
     private final int POLLING_PERIOD_DEFAULT = 1800;
     private int pollingPeriod = POLLING_PERIOD_DEFAULT;
 
+    private final int POLLING_OR_REPORTING_FACTOR = 2;
```
cdjackson: I think this should be >2 to avoid mistriggers. you want to allow one loss, and setting this to 2 doesn't provide that - or at least it provides absolutely no margin.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeThingHandler.java
```diff
@@ -401,6 +412,11 @@ private synchronized void doNodeInitialisation() {
         }
         logger.debug("{}: Channel initialisation complete", nodeIeeeAddress);
 
+        int pollOrReportingInterval = findMinimumPollOrReportingInterval(channels);
```
cdjackson: I wonder if these names can't be changed? How about just ```expectedUpdatePeriod``` or ```refreshPeriod```?

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeThingHandler.java
```diff
@@ -423,6 +439,23 @@ private synchronized void doNodeInitialisation() {
         coordinatorHandler.serializeNetwork();
     }
 
+    private int findMinimumPollOrReportingInterval(Map<ChannelUID, ZigBeeBaseChannelConverter> channels) {
```
cdjackson: Likewise - I think this name is a mouthful and not so easy to understand.  How about just ```getExpectedUpdatePeriod``` or ```getRefreshPeriod```?

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigbeeIsAliveTracker.java
```diff
@@ -0,0 +1,83 @@
...
+
+    protected final ScheduledExecutorService scheduler = ThreadPoolManager.getScheduledPool("ZigbeeIsAliveTracker");
+
+    /**
+     * Adds a mapping from {@link ZigBeeThingHandler} to the interval in which it should have communicated
+     *
+     * @param zigBeeThingHandler      the {@link ZigBeeThingHandler}
+     * @param pollOrReportingInterval the interval in which the device should have communicated with us
+     */
+    public void addHandlerIntervalMapping(ZigBeeThingHandler zigBeeThingHandler, int pollOrReportingInterval) {
```
cdjackson: Do these methods need to have the ```IntervalMapping``` on the end? Why not just ```addHandler``` and ```removeHandler```? The fact that it uses a map is implementation detail that doesn't really need to be reflected in the method name?

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/converter/ZigBeeBaseChannelConverter.java
triller-telekom: I thought about this too, BUT...

There is a minimum reporting period and a maximum reporting period for **each** attribute reporting. And what this variable should represent is the minimum of all those maximum reporting times :)

So I find `getMinimumReportingPeriod` confusing as well, because you could think of this as the minimum reporting period for an attribute, whereas i am only considering the maximums...

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/converter/ZigBeeBaseChannelConverter.java
triller-telekom: Yes, it IS aligned with the reporting configurations, please see the changes I made to the converters.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeThingHandler.java
triller-telekom: Right, I was just afraid that 3 might be too large as the default is 7200 seconds = 2h maximum reporting time, times 3 is 6 hours until it goes offline.

But if we only want to allow **one** loss, we do not need 3, we could also use a slightly larger number than 2 and round it to an INT, WDYT?

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeThingHandler.java
triller-telekom: Sure, `expectedUpdatePeriod` sounds good.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeThingHandler.java
triller-telekom: See above, `getExpectedUpdatePeriod` sounds good to me as well.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigbeeIsAliveTracker.java
triller-telekom: I agree, will change it.

triller-telekom:
I have added a commit addressing the obvious changes.

> One thought though - how does this change fit in with the reporting configuration options? I know it's only just been implemented and hasn't really flown through to the converters, but is this all going to change when the configuration is added?

I am not aware of the configuration options you are planning to introduce. Do you have a link, pointing me to more information?

However, for now the `minimalMaximumReportingPeriod` gets set by each converter individually. This can still be done if you configure each converter to use different reporting intervals.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/converter/ZigBeeBaseChannelConverter.java
cdjackson: I understand that, but for the sake of clarity I think it's much clearer to use the shorter name. I don't think it really needs to reflect exactly what is happening in the converter and you can always describe this in the text.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/converter/ZigBeeBaseChannelConverter.java
cdjackson: Ok, maybe I missed this, but I didn't see any reference to the ```ZclReportingConfig``` class. My apologies - I will take a better look later.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeThingHandler.java
cdjackson: Yes, I didn't expect this to be 3 - I thought 2.1, or something like that. Or you could use 2, and then add 30 seconds or something similar - just something to give some margin.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeThingHandler.java
cdjackson: Ta.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/converter/ZigBeeBaseChannelConverter.java
triller-telekom: Maybe we have a misunderstanding here since I am not yet aware of your planned reporting configurations, see my comment https://github.com/openhab/org.openhab.binding.zigbee/pull/412#issuecomment-473198733 regarding this.

What I do in the converter is, I call the new method

```
handleReportingResponse(CommandResult reportResponse, int reportingFailedPollingInterval,
            int reportingSuccessMaxReportInterval)
```

which sets the `pollingPeriod` and `minimalMaximumReportingPeriod` fields accordingly, depending on whether the reporting configuration request was successful or not.

cdjackson: > What I do in the converter is, I call the new method

Yes, that I saw. My question is if this isn't the appropriate time to start using the new class rather than have to change again. This function was added a month or two back, but not rolled into the converters yet.

triller-telekom: Which function was added a month or two back?

The

```
handleReportingResponse(CommandResult reportResponse, int reportingFailedPollingInterval,
            int reportingSuccessMaxReportInterval)
```

is my function which I added in this PR :)

cdjackson: As above - it is the ```ZclReportingConfig``` class which is designed to allow users to configure all of these timers that are currently hard coded here.

I think (but might be wrong as I wrote this a month or two back) that it may just be a case of changing to use methods in this class to get the times, rather than hard coding them.

triller-telekom: Ah I see, so what you mean is that the converters should make use of the `ZclReportingConfig` instead of using hard coded values.

Since this is not the case for now, I have used these hardcoded values as input to my `handleReportingResponse` method. If the `ZclReportingConfig` gets used in the future, one can easily replace them.

I think this is out of scope of this PR and would only blow up the lines of code. I think a separate should take care of making the channel converters use the `ZclReportingConfig`, WDYT?

triller-telekom: @cdjackson may I get back to you on this issue?

I see a usage of the `ZclReportingConfig` in 

https://github.com/openhab/org.openhab.binding.zigbee/blob/0c7dbd66a6a7df79ce9e4c89e1b423060a6e8c14/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterSwitchOnoff.java#L119-L120

for example.

However, when the device is configured, this configuration is not used, but the hard-coded default values are used.

https://github.com/openhab/org.openhab.binding.zigbee/blob/0c7dbd66a6a7df79ce9e4c89e1b423060a6e8c14/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterSwitchOnoff.java#L68-L72

So from my point of view the change that I made does not touch the implementation you did with the `ZclReportingConfig` at all. it is totally independent from it.

Once the created `ZclReportingConfig` in a `initializeConverter()` is used in the `initializeDevice()`  the changes I make in this PR can flow into it.

Making use of the configuration set in the `initializeConverter()` inside the `initializeDevice()` should IMHO be done in a separate PR as it hasn't been done so far for any converter.

So how shall we proceed on this PR?

triller-telekom:
@cdjackson Could you please have a look at my comment https://github.com/openhab/org.openhab.binding.zigbee/pull/412#discussion_r268584384 again? Especially on the fact that hardly any converter makes use of the new configuration classes yet.

We have briefly talked about this on the phone and you mentioned that you are fine with having the converters use the new configuration classes in a separate PR. So I am wondering what is missing on this PR before we can merge it?

cdjackson:
I would have preferred to see this implemented here so that it gets implemented rather than just changing all this code again. But I guess you're not happy to do so, so let's just merge!

triller-telekom:
I have rebased the branch on the current master and squashed my commits.

cdjackson:
CI appears to be failing....

triller-telekom:
Since you closed and reopened the PR I assume its unrelated, is it? Because I cannot see the build logs anymore in this closed state now.

cdjackson:
The errors looked at first glance to be download related (ie it failed to download some dependencies). There were also a lot of errors about the license format, but I thought that these shouldn't fail the build (and it shouldn't have changed.

I closed it to to retrigger CI which only seems to work if the PR is closed for some (unknown) time (a few minutes doesn't seem to work, so I left it closed a little longer).

Let's see if it works, otherwise we'll need to look at it further.

cdjackson:
So it's failed again - looks like license issues and other checkstyle errors. I thought this was disabled, and I'm not sure why it's an issue here as it wasn't (I believe!) in master as the PR from @hsudbrock yesterda fixed this.

triller-telekom:
I broke the `ZigBeeConverterIasTamperTest` and have fixed it now. It needed a little more mocking to not run into an NPE.

cdjackson:
Thanks.

PR #413, opened 2019-03-19
Store BasicCluster.ApplicationVersion as Thing property

triller-telekom:
Currently the `ZigBeeNodePropertyDiscoverer` sets the thing properties as
defined in the `Thing` class:

PROPERTY_VENDOR
PROPERTY_MODEL_ID
PROPERTY_HARDWARE_VERSION

In addition it sets the following properties which are zigbee specific:

THING_PROPERTY_STKVERSION
THING_PROPERTY_ZCLVERSION
THING_PROPERTY_DATECODE

This PR also adds the applicationVersion as a property on the thing.

By definition it holds the "version number of the application software" of
the device and thus might be useful for certain devices.

For example, BitronHome stores their non upgradeable firmware version
(thus no ota cluster available) in this attribute on the 902010/24 smoke
detector.


hsudbrock:
Looks good to me, feel free to merge if you think that it is ok, Chris.

Issue #415, opened 2019-03-21
RuntimeException thrown when ZigBee stick is unplugged.


YordanDZhelev:
In case of unplugged zigbee stick the binding still tries to open it, which results in the following `RuntimeException` 
```
java.lang.RuntimeException: No such device in open
    at gnu.io.RXTXCommDriver.getCommPort(getCommPort:835)
    at gnu.io.CommPortIdentifier.open(open:383)
    at org.openhab.binding.zigbee.handler.ZigBeeSerialPort.open(open:136)
    at org.openhab.binding.zigbee.handler.ZigBeeSerialPort.open(open:112)
    at com.zsmartsystems.zigbee.dongle.ember.ZigBeeDongleEzsp.initialiseEzspProtocol(initialiseEzspProtocol:878)
    at com.zsmartsystems.zigbee.dongle.ember.ZigBeeDongleEzsp.initialize(initialize:310)
    at com.zsmartsystems.zigbee.ZigBeeNetworkManager.initialize(initialize:334)
    at org.openhab.binding.zigbee.handler.ZigBeeCoordinatorHandler.initialiseZigBee(initialiseZigBee:422)
    at org.openhab.binding.zigbee.handler.ZigBeeCoordinatorHandler.access$6(access$6:382)
    at org.openhab.binding.zigbee.handler.ZigBeeCoordinatorHandler$1.run(run:532)
    at java.util.concurrent.Executors$RunnableAdapter.call(call)
    at java.util.concurrent.FutureTask.runAndReset(runAndReset)
    at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$301(access$301)
    at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(run)
    at java.util.concurrent.ThreadPoolExecutor.runWorker(runWorker)
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(run)
    at java.lang.Thread.run(run)
```
The same error was observed in two different variants:
```
java.lang.RuntimeException: No such file or directory in open
    at gnu.io.RXTXCommDriver.getCommPort(getCommPort:835)
    at gnu.io.CommPortIdentifier.open(open:383)
    at org.openhab.binding.zigbee.handler.ZigBeeSerialPort.open(open:136)
```
and 
```
java.lang.RuntimeException: No such device in open
    at gnu.io.RXTXCommDriver.getCommPort(getCommPort:835)
[...]

PR #416, opened 2019-03-21
Handle RuntimeException in case of unplugged ZigBee stick.

YordanDZhelev:
Fixes https://github.com/openhab/org.openhab.binding.zigbee/issues/415
In case of unplugged ZigBee stick a `RuntimeException` is thrown. In that case we should not continue to try to open it.

### DIFF
--- a/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeSerialPort.java
+++ b/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeSerialPort.java
@@ -12,7 +12,9 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.util.Enumeration;
+import java.util.Set;
 import java.util.TooManyListenersException;
+import java.util.concurrent.ConcurrentHashMap;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -94,6 +96,8 @@ public class ZigBeeSerialPort implements ZigBeePort, SerialPortEventListener {
      */
     private final Object bufferSynchronisationObject = new Object();
 
+    private Set<String> portOpenRuntimeExcepionMessages = ConcurrentHashMap.newKeySet();
+
     /**
      * Constructor setting port name and baud rate.
      *
@@ -157,6 +161,7 @@ public class ZigBeeSerialPort implements ZigBeePort, SerialPortEventListener {
                 serialPort.notifyOnDataAvailable(true);
 
                 logger.debug("Serial port [{}] is initialized.", portName);
+                portOpenRuntimeExcepionMessages.clear();
             } catch (NoSuchPortException e) {
                 logger.error("Serial Error: Port {} does not exist.", portName);
                 return false;
@@ -168,6 +173,12 @@ public class ZigBeeSerialPort implements ZigBeePort, SerialPortEventListener {
                 return false;
             } catch (TooManyListenersException e) {
                 logger.error("Serial Error: Too many listeners on Port {}.", portName);
+                return false;
+            } catch (RuntimeException e) {
+                if (!portOpenRuntimeExcepionMessages.contains(e.getMessage())) {
+                    portOpenRuntimeExcepionMessages.add(e.getMessage());
+                    logger.error("Serial Error: Device cannot be opened on Port {}. Caused by {}", portName, e.getMessage());
+                }
                 return false;
             }
 


org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeSerialPort.java
```diff
@@ -170,7 +170,11 @@ public boolean open(int baudRate, FlowControl flowControl) {
                 logger.error("Serial Error: Too many listeners on Port {}.", portName);
                 return false;
             } catch (RuntimeException e) {
-                logger.debug("Device cannot be opened on Port {}. Caused by {}", portName, e.getCause());
+                if(e.getMessage().contains("No such file or directory in open") || e.getMessage().contains("No such device in open")) {
```
cdjackson: Does this work with different languages?

YordanDZhelev: I believe this messages come from the native part and are hardcoded there.

cdjackson: And is this text common for all JREs?  It just seems like a not very deterministic way to do this across all systems in all languages etc. I'm not really sure why this is needed to be honest?

So long as this doesn't continually log every few seconds in the event that the stick is removed, I don't see why we don't just log as error in the same way as the other errors logged here.

Also, can you make the text the same as the other errors (ie add "Serial Error: ").

triller-telekom: > So long as this doesn't continually log every few seconds in the event that the stick is removed, I don't see why we don't just log as error in the same way as the other errors logged here.

Unfortunately we have observed that (maybe depending on the serial library used underneath) that these RuntimeExceptions occur periodically when the stick is plugged out. Therefore we thought we "somehow" catch that case separately and log it on debug level, whereas other RuntimeExceptions should still be visible on the warn level.

So the question is: Are these messages localized somehow? I wonder where one could look that up...

cdjackson:
So you mean that everything is working normally, and then this exception occurs - or do you mean it occurs as the stick is plugged in?

So long as it only happens once in a while, I don't think it matters too much and we probably should just log it in a similar way to the others in this method.

Regarding localisation - I also wondered above if this message is also the same in other JREs? We probably shouldn't assume that everyone is using a specific JRE and I wouldn't be surprised if this is a slightly different message in different environments (ie I'm not sure this check is robust)?

hsudbrock:
I think that @cdjackson and @triller-telekom have a point that using the exception message might be a problem (due to (a) different languages and (b) different implementations of `gnu.io`).

I think that the 'good' solution would be to detect when a USB dongle is removed and replugged, and simply not try to reconnect when the dongle is plugged out. But this would be a bigger thing, and I do not suggest to do this just to get rid of too many log messages.

Here's one other idea how to solve the issue that does not have the "exception message language issue": Log the `RuntimeException` on level `warn` when it occurs with a disclaimer that following equal exceptions will be logged on level `debug`, and then log all subsequent equal `RuntimeException`s on level `debug`. When a not equal `RuntimeException` occurs, do this again on `warn` and so forth.

WDYT?

hsudbrock:
> So you mean that everything is working normally, and then this exception occurs - or do you mean it occurs as the stick is plugged in?

This exception happens every couple of seconds when the stick is plugged out, due to the "reconnect logic" in the coordinator handler.

cdjackson:
I don't mind this, but if you're going to change the logging from warn to debug after the first error, then why not inhibit the logging alltogether until the port is successfully opened again? 

Ie -:
* On first error, log an error and set a flag
* Subsequent errors then aren't logged
* When the port is opened successfully, reset the flag

I'm not sure if we can make this work for all of the exceptions that are logged in this method?

hsudbrock:
> I don't mind this, but if you're going to change the logging from warn to debug after the first error, then why not inhibit the logging alltogether until the port is successfully opened again?

I would like to get a `warn` log message if the dongle is plugged in again, but the serial port still cannot be opened. That's why I would log a `warn` message again if the message of the `RuntimeException` changes (which assumes that the message used when no dongle is plugged in differs from the message for other issues with plugged in dongles).

cdjackson:
Ok, but that’s fine - it get’s logged once only. So the flag can get reset when the port is opened successfully, and then at the first exception, it gets set so that any subsequent exceptions up to the next successful opening are not logged - that’s what I’m suggesting above at least, but maybe I’m missing a subtlety here?

Also, why a warn and not error? I’m just thinking that there are a lot of messages being logged in this method, and probably they should be consistent (either warn or error - I don’t mind, but they should be the same shouldn’t they?).

hsudbrock:
> Ok, but that’s fine - it get’s logged once only. So the flag can get reset when the port is opened successfully, and then at the first exception, it gets set so that any subsequent exceptions up to the next successful opening are not logged - that’s what I’m suggesting above at least, but maybe I’m missing a subtlety here?

I mean the case where the port cannot be opened successfully when the dongle is plugged in again. In this case, I want the log message to be logged not on `debug`, but on `warn` or `error`. (Or do I miss something?)

> Also, why a warn and not error? I’m just thinking that there are a lot of messages being logged in this method, and probably they should be consistent (either warn or error - I don’t mind, but they should be the same shouldn’t they?).

Both fine for me - I agree that this should be consistent.

cdjackson:
>  I want the log message to be logged not on debug, but on warn or error. (Or do I miss something?)

No problem - I suggested it is logged at warn or error. All I'm saying is do we need to log it at all every other time at debug? I don't think so. My suggestion is (expanded to try and avoid confusion) -:

* Open the port, and clear the flag
* On first error, log an ```error``` and set a flag
* Subsequent errors then aren't logged (at debug, or any other level)
* When the port is opened successfully, reset the flag and we start from the beginning

I feel that I'm simply missing something here - or at least I don't really understand what is now being suggested. I think that what I'm suggesting above is simple (both to understand, and to implement) although I've not looked at it in detail, so maybe there's something subtle I'm missing, but I feel that you guys at DT have something against this for some reason that I'm missing?

hsudbrock:
Let me try to clarify what I mean with an example. Assume that the following exceptions occur:
* Firstly, 5000 `RuntimeException`s, because the dongle is not plugged in and the reconnect job tries to open the port every couple of seconds.
* Then, a `PortInUseException`, because the dongle has been plugged in again but the port was grabbed by someone else when the binding tried to connect to the port.

As I understand your suggestion, @cdjackson , it would log the `PortInUseException` on level `debug`, because the port was *not* opened successfully after the first `RuntimeException` (and the flag is only reset after a successful port open). Please correct me if I am wrong.

What I would like to achieve is that the `PortInUseException` is logged on `warn` or `error`. This can be done as follows (taking your suggestion from above and adding the parts that I think are missing to achieve what I described above in italics):

* Open the port, and clear the flag
* On first error, log an error and set a flag *and store the type/message of the exception*
* Subsequent errors *with the same type/message* then aren't logged (at debug, or any other level)
* *Also, when the exception type/message differs from the stored one, log an error, and store the new type/message of the exception*
* When the port is opened successfully, reset the flag *and remove the stored type/message of the exception* and we start from the beginning

cdjackson:
>  it would log the PortInUseException on level debug

My proposal is not to log anything at debug - the logging should be at error or warn level. I don't see any point in logging the exception at debug level every 2 seconds when it's retrying the opening of the port and the stick isn't there.

I'm happy with your idea though - that's fine with me. Thanks.

YordanDZhelev:
I updated the PR to only log the first RuntimeException with a particular message. Following RuntimeExceptions with the same message will just be skipped untill a successful open of the port.

cdjackson:
Thanks - looks good. I was wondering if we shouldn't treat the other exceptions the same way - eg at the moment do we get the "No port" exception if the stick is unplugged?

As suggested previously, can you also add the "Serial Error:" statement to the front of the log so that it has the same format as the other log entries (I know - I'm being picky - sorry).

Can you rebase please - there have been changes in the openhab project and CI doesn't compile this branch now.

It might also be worth squashing the commits and signing them so that the DCO check doesn't fail due to unsigned commits.

YordanDZhelev:
@cdjackson do you think that there is anything else needed to be implemented as part of this change?

cdjackson:
Sorry - I missed that you'd updated this.

LGTM - thanks.

PR #418, opened 2019-03-22
Add support for IAS movement and vibration alarms

### LINKED ISSUE #417: Add support for IAS movement & vibration alarms
The `ZoneType` `0x002d` of the IAS Zone cluster is for 'Vibration/Movement sensors'; it uses the `Alarm1` for movement indication, and the `Alarm2` for vibration.

This issue is for adding corresponding channel types and converters to the binding, I will provide a PR.

One example device that supports this zone type is bitron's glass break sensor AV 2010/33.

hsudbrock:
This resolves #417 

Here's a screenshot of how an IAS device with zonetype `0x002d` looks like in paper UI with the two new converters (after shaking the device, and, hence, with vibration alarm shown):

image

### DIFF
--- a/org.openhab.binding.zigbee/ESH-INF/thing/channels.xml
+++ b/org.openhab.binding.zigbee/ESH-INF/thing/channels.xml
@@ -136,6 +136,24 @@
         <category>Sensor</category>
         <state readOnly="true" />
     </channel-type>
+    
+    <!-- IAS Movement Sensor Channel -->
+    <channel-type id="ias_movement">
+        <item-type>Switch</item-type>
+        <label>Movement Sensor</label>
+        <description>Movement Sensor Alarm</description>
+        <category>Sensor</category>
+        <state readOnly="true" />
+    </channel-type>
+    
+    <!-- IAS Vibration Sensor Channel -->
+    <channel-type id="ias_vibration">
+        <item-type>Switch</item-type>
+        <label>Vibration Sensor</label>
+        <description>Vibration Sensor Alarm</description>
+        <category>Sensor</category>
+        <state readOnly="true" />
+    </channel-type>
 
     <!-- IAS Tamper Channel -->
     <channel-type id="ias_tamper">
--- a/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterIasMovement.java
+++ b/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterIasMovement.java
@@ -1,0 +1,45 @@
+/**
+ * Copyright (c) 2010-2018 by the respective copyright holders.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.openhab.binding.zigbee.internal.converter;
+
+import org.eclipse.smarthome.core.thing.Channel;
+import org.eclipse.smarthome.core.thing.ThingUID;
+import org.eclipse.smarthome.core.thing.binding.builder.ChannelBuilder;
+import org.openhab.binding.zigbee.ZigBeeBindingConstants;
+
+import com.zsmartsystems.zigbee.ZigBeeEndpoint;
+import com.zsmartsystems.zigbee.zcl.clusters.iaszone.ZoneTypeEnum;
+
+/**
+ * Converter for the IAS movement sensor.
+ *
+ * @author Henning Sudbrock - Initial Contribution
+ */
+public class ZigBeeConverterIasMovement extends ZigBeeConverterIas {
+
+    @Override
+    public boolean initializeConverter() {
+        bitTest = CIE_ALARM1;
+        return super.initializeConverter();
+    }
+
+    @Override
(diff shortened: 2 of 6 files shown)


org.openhab.binding.zigbee/ESH-INF/thing/channels.xml
```diff
@@ -136,6 +136,24 @@
         <category>Sensor</category>
         <state readOnly="true" />
     </channel-type>
+    
+    <!-- IAS Movement Sensor Channel -->
+    <channel-type id="ias_movement">
+        <item-type>Switch</item-type>
+        <label>Movement Sensor</label>
+        <description>Movement Sensor Alarm</description>
+        <category>Sensor</category>
```
triller-telekom: I wonder why `ias_water` has category `Sensor` though the description mentions "Alarm" and `ias_tamper` has category `Alarm`.

So IMHO all alarm related channels should have category `Alarm`. WDYT?

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterIasMovement.java
```diff
@@ -0,0 +1,45 @@
+/**
+ * Copyright (c) 2010-2018 by the respective copyright holders.
```
triller-telekom: Is this a bug in the maven pom from openHAB? because IMHO this should state 2019...

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterIasMovement.java
```diff
@@ -0,0 +1,45 @@
...
+
+    @Override
+    public boolean initializeConverter() {
+        bitTest = CIE_ALARM1;
+        return super.initializeConverter();
+    }
+
+    @Override
+    public Channel getChannel(ThingUID thingUID, ZigBeeEndpoint endpoint) {
+        if (!supportsIasChannel(endpoint, ZoneTypeEnum.VIBRATION_MOVEMENT_SENSOR)) {
```
triller-telekom: Just out of curiosity: Both of your two converters: `ZigBeeConverterIasMovement` and `ZigBeeConverterIasVibration` have the same condition, so devices will always get both channels. is this intended or does one of the converters use a wrong enum type?

hsudbrock: It would be nice to do better here, but I do not see how one can do better, due to how the IAS zone cluster works: The `zone state` of a device implementing the cluster is a bitmap, where bits `0` and `1` provide two alarm flags. Based on the `zone type` provided by the device, one knows what the two alarms mean (in this case, vibration and movement). But one cannot see which of the two alarms are actually implemented by a device (a device could only use the first, the second, or both alarm flags, i.e., here, depending on whether the device supports vibration alarms only, movement alarms only, or both).

org.openhab.binding.zigbee/ESH-INF/thing/channels.xml
hsudbrock: Good question; there are other IAS-based channels that also have a category that is not `Alarm`, and where I think that they should keep that category, viz the motion detection channels.

Regarding the water alarm, I don't know whether `Sensor` or `Alarm` is better. But I think that this is out of scope of this PR.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterIasMovement.java
hsudbrock: Hm, I just took the same copyright message that is currently used for the other classes in this binding. I guess that you are right that this should be changed, but I also think that this larger change should not be mixed into this PR - wdyt?

Out of curiosity: Why to you think that this is a bug in the maven pom from openHAB, i.e., where is the connection from the pom to the copyright message?

org.openhab.binding.zigbee/ESH-INF/thing/channels.xml
triller-telekom: I am fine with mixing tags, I was just wondering whether all these channels with `Alarm` in their description should have the `Alarm` category. But lets leave it as a sensor now.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterIasMovement.java
triller-telekom: > Out of curiosity: Why to you think that this is a bug in the maven pom from openHAB, i.e., where is the connection from the pom to the copyright message?

What I meant is that someone should use maven to correct it. But I agree , this should not be mixed with this PR. if the whole binding is still 2018, this is fine for this PR.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterIasMovement.java
triller-telekom: Thanks for the clarification!

hsudbrock:
@cdjackson Is it ok to merge this from your side?

cdjackson:
lgtm - thanks Henning.

PR #408, opened 2019-03-10
add support for xiaomi wireless mini switch

t-8ch:
https://www.aqara.com/en/wireless_mini_switch.html

shortpress requires https://github.com/zsmartsystems/com.zsmartsystems.zigbee/pull/556
doublepress requires support for proprietary attributes.

PR #424, opened 2019-04-08
Aqara mini switch with gyro is not supported

schwabe:
With 2.5.0MR1 the variant with Gyro is not supported and I don't see commits changing this, so add a note here until it is supported.


schwabe:
Looks like removing it again and adding it again for the 5th time actually worked this time, strange device ....

schwabe:
It is being detected as switch but no button press or shaking triggers any log of ChannelTriggeredEvent nor is the channel "switch" changing.

t-8ch:
@schwabe https://github.com/openhab/org.openhab.binding.zigbee/pull/408 lists the requirements for the thing to actually work.
I was not aware of the shake capabilities.

schwabe:
@t-8ch there seem to be two different variants of the mini switch. One with shake (called gyro/WXKG12LM) and one without. 

I would test the changes and I am quite familiar but last week when I tried I gave up after an hour since could not find any tutorial how this whole Karaf, building bundles for Openhab/Karaf etc. works.

PR #425, opened 2019-04-16
Refactor ZigBeeConverterAtmosphericPressure

YordanDZhelev:
Refactor ZigBeeConverterAtmosphericPressure to split the device and channel initialization
Initialization logic split in two methods similar `ZigBeeConverterSwitchOnoff` refactor done in [this commit](https://github.com/openhab/org.openhab.binding.zigbee/commit/45f74b9c7e7a7e5a59bb68c536f4a3ed64535bb5)

### DIFF
--- a/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterAtmosphericPressure.java
+++ b/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterAtmosphericPressure.java
@@ -52,40 +52,56 @@ public class ZigBeeConverterAtmosphericPressure extends ZigBeeBaseChannelConvert
     private Integer enhancedScale = null;
 
     @Override
-    public synchronized boolean initializeConverter() {
+    public boolean initializeDevice() {
+        ZclPressureMeasurementCluster serverCluster = (ZclPressureMeasurementCluster) endpoint
+                .getInputCluster(ZclPressureMeasurementCluster.CLUSTER_ID);
+        if (serverCluster == null) {
+            logger.error("{}: Error opening device pressure measurement cluster", endpoint.getIeeeAddress());
+            return false;
+        }
+
+        // Check if the enhanced attributes are supported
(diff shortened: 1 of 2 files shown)


org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterAtmosphericPressure.java
```diff
@@ -52,40 +52,56 @@
...
+                } else {
+                    reportingResponse = serverCluster.setMeasuredValueReporting(1, REPORTING_PERIOD_DEFAULT_MAX, 0.1)
+                            .get();
+                    handleReportingResponse(reportingResponse, POLLING_PERIOD_DEFAULT, REPORTING_PERIOD_DEFAULT_MAX);
+                }
             } else {
-                reportingResponse = cluster.setMeasuredValueReporting(1, REPORTING_PERIOD_DEFAULT_MAX, 0.1).get();
-                handleReportingResponse(reportingResponse, POLLING_PERIOD_DEFAULT, REPORTING_PERIOD_DEFAULT_MAX);
+                logger.error("{}: Error 0x{} setting server binding", endpoint.getIeeeAddress(),
+                        Integer.toHexString(bindResponse.getStatusCode()));
```
cdjackson: I think we actually should return false if there's an error here. I suspect you've copied this from the converter I wrote originally which also returns true in the event of an exception, but it seems incorrect to me. WDYT?

YordanDZhelev: >  I suspect you've copied this from the converter I wrote originally
That is correct, I used the `ZigBeeConverterSwitchOnoff` as a reference. I can see now that it is also described in the `ZigBeeBaseChannelConverter.initializeDevice()` Javadoc that:
- this methos should return true if the device was configured correctly. Therefore, it looks that we should return `false` here. 
- this methos should set the `pollingPeriod` variabl to `POLLING_PERIOD_HIGH` if it fails. 

I will implement both changes.

PR #427, opened 2019-04-19
Refactor ZigBee binding channel converters

YordanDZhelev:
Extract device initialization from `initializeConverter()` method to `initializeDevice()`.
This PR only split the existing functionality without changing the logic. If any refactoring is needed e.g. additional setting of the polling interval, or returning `false` during device initialization in case of an exception, it can be done in a separate PR.


cdjackson:
Thanks @YordanDZhelev I will take a look over the coming days.

cdjackson:
I think in most cases, the ```initialiseDevice``` method is not returning ```false``` when there is an error (as per our discussion on your previous PR). This should be updated.

YordanDZhelev:
> I think in most cases, the `initialiseDevice` method is not returning `false` when there is an error (as per our discussion on your previous PR). This should be updated.

PR is updated with one commit. Please take a closer look at ZigBeeConverterColorColor.

YordanDZhelev:
@cdjackson could you please take a look again

openhab-5iver:
@YordanDZhelev, have you seen #430? It appears this PR has broken something. Please let me know if you need any more details or if I can help test anything.

PR #434, opened 2019-05-03
Refactoring of some converters to always initialize variables

### LINKED ISSUE #432: NPE in some converters when calculating updated attributes
Caused by #427 and #409 

I'm experiencing a NPE in the method `ZigBeeConverterMeasurementPower.attributeUpdated(...)` thrown by this statement:
```
BigDecimal valueInWatt = BigDecimal.valueOf(value * multiplier / divisor);
```
The reason is that `multiplier` and `divisor` are both null. This happens because since #427 these variables are initialized in the method `initializeDevice()` which since #409 is only invoked on the very first initialization and then only `initializeConverter()` will be called on startup. Such variables should be better initialized in the method `initializeConverter()`.

Other converters which currently use to initialize variables in `initializeDevice()` will need to be changed too (didn't check them all).

TomTravaglino:
Resolves #432 

Some converters used to initialize variables on device initialization in
order to handle attributes updates. After the initilization has been
been split into a device and a converter initialization these variables
were no longer initialized when only the converter was initialized. This
made the initialization run into NPEs. This commit ensures that the
initialization is properly done when only the converter is initialized.

### DIFF
--- a/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterColorTemperature.java
+++ b/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterColorTemperature.java
@@ -62,8 +62,6 @@ public class ZigBeeConverterColorTemperature extends ZigBeeBaseChannelConverter 
             return false;
         }
 
-        determineMinMaxTemperature(serverClusterColorControl);
-
         try {
             CommandResult bindResponse = bind(serverClusterColorControl).get();
             if (bindResponse.isSuccess()) {
--- a/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterColorTemperature.java
+++ b/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterColorTemperature.java
@@ -62,24 +62,7 @@ public class ZigBeeConverterColorTemperature extends ZigBeeBaseChannelConverter 
             return false;
         }
 
-        Integer minTemperatureInMired = serverClusterColorControl.getColorTemperatureMin(Long.MAX_VALUE);
-        Integer maxTemperatureInMired = serverClusterColorControl.getColorTemperatureMax(Long.MAX_VALUE);
-
-        // High Mired values correspond to low Kelvin values, hence the max Mired value yields the min Kelvin value
-        if (maxTemperatureInMired == null) {
-            kelvinMin = DEFAULT_MIN_TEMPERATURE_IN_KELVIN;
-        } else {
-            kelvinMin = miredToKelvin(maxTemperatureInMired);
-        }
-
-        // Low Mired values correspond to high Kelvin values, hence the min Mired value yields the max Kelvin value
-        if (minTemperatureInMired == null) {
-            kelvinMax = DEFAULT_MAX_TEMPERATURE_IN_KELVIN;
-        } else {
-            kelvinMax = miredToKelvin(minTemperatureInMired);
-        }
-
-        kelvinRange = kelvinMax - kelvinMin;
+        determineMinMaxTemperature(serverClusterColorControl);
 
         try {
             CommandResult bindResponse = bind(serverClusterColorControl).get();
@@ -112,6 +95,8 @@ public class ZigBeeConverterColorTemperature extends ZigBeeBaseChannelConverter 
                     endpoint.getEndpointId());
             return false;
         }
+
+        determineMinMaxTemperature(clusterColorControl);
 
         clusterColorControl.addAttributeListener(this);
         return true;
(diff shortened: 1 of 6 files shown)


org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterColorTemperature.java
```diff
@@ -62,24 +62,7 @@ public boolean initializeDevice() {
...
-
-        // Low Mired values correspond to high Kelvin values, hence the min Mired value yields the max Kelvin value
-        if (minTemperatureInMired == null) {
-            kelvinMax = DEFAULT_MAX_TEMPERATURE_IN_KELVIN;
-        } else {
-            kelvinMax = miredToKelvin(minTemperatureInMired);
-        }
-
-        kelvinRange = kelvinMax - kelvinMin;
+        determineMinMaxTemperature(serverClusterColorControl);
```
cdjackson: Why is this needed here? kelvinRange is not used in this method.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterMeasurementPower.java
```diff
@@ -71,12 +71,8 @@ public boolean initializeDevice() {
             return false;
         }
 
-        divisor = serverClusterMeasurement.getAcPowerDivisor(Long.MAX_VALUE);
-        multiplier = serverClusterMeasurement.getAcPowerMultiplier(Long.MAX_VALUE);
-        if (divisor == null || multiplier == null) {
-            divisor = 1;
-            multiplier = 1;
-        }
+        determineDivisorAndMultiplier(serverClusterMeasurement);
```
cdjackson: Why is this called here - this is at the end of the method, so surely nothing from this method is used after this?

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterMeasurementRmsCurrent.java
```diff
@@ -72,12 +72,8 @@ public boolean initializeDevice() {
             return false;
         }
 
-        divisor = serviceClusterMeasurement.getAcCurrentDivisor(Long.MAX_VALUE);
-        multiplier = serviceClusterMeasurement.getAcCurrentMultiplier(Long.MAX_VALUE);
-        if (divisor == null || multiplier == null) {
-            divisor = 1;
-            multiplier = 1;
-        }
+        determineDivisorAndMultiplier(serverClusterMeasurement);
```
cdjackson: Why do we need to call this here? It's right at the end of the initialisation method, so anything done there is not used.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterMeasurementRmsVoltage.java
```diff
@@ -72,12 +72,7 @@ public boolean initializeDevice() {
             return false;
         }
 
-        divisor = serverClusterMeasurement.getAcVoltageDivisor(Long.MAX_VALUE);
-        multiplier = serverClusterMeasurement.getAcVoltageMultiplier(Long.MAX_VALUE);
-        if (divisor == null || multiplier == null) {
-            divisor = 1;
-            multiplier = 1;
-        }
+        determineDivisorAndMultiplier(serverClusterMeasurement);
```
cdjackson: Same comment - why is this called?

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterColorTemperature.java
TomTravaglino: Will remove it.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterMeasurementPower.java
TomTravaglino: Will remove it.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterMeasurementRmsCurrent.java
TomTravaglino: Will remove it.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterMeasurementRmsVoltage.java
TomTravaglino: Will remove it.

TomTravaglino:
I updated the PR

PR #438, opened 2019-05-10
Create Discovery Results only if a startScan() was triggered

### LINKED ISSUE #437: Discovery results from deleted Things reappear in the Inbox

triller-telekom:
Fixes #437


org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/discovery/ZigBeeDiscoveryService.java
```diff
@@ -86,7 +87,9 @@ protected void addZigBeeCoordinatorHandler(ZigBeeCoordinatorHandler coordinatorH
         ZigBeeNetworkNodeListener listener = new ZigBeeNetworkNodeListener() {
             @Override
             public void nodeAdded(ZigBeeNode node) {
-                ZigBeeDiscoveryService.this.nodeDiscovered(coordinatorHandler, node);
+                if (searchStarted) {
```
TomTravaglino: Maybe `scanStarted` is a better name for the variable as it better describes its purpose. WDYT?
I'm wondering if `searchStarted` needs to be synchronized as it can be accessed from different threads.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/discovery/ZigBeeDiscoveryService.java
```diff
@@ -128,9 +133,11 @@ public void startScan() {
             }
 
             logger.debug("Starting ZigBee scan for {}", coordinator.getUID());
-            coordinator.scanStart();
+            coordinator.scanStart(SEARCH_TIME);
         }
 
+        Runnable searchFinishedIndicator = () -> searchStarted = false;
```
TomTravaglino: Does `searchStarted` need to be synchronized? See also my above.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/discovery/ZigBeeDiscoveryService.java
triller-telekom: I have changed the name to `scanStarted`. It makes more sense, since the rest of the code is also talking about a "scan", rather than a "search". Thanks for pointing this out.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/discovery/ZigBeeDiscoveryService.java
triller-telekom: That's why I made the the variable `volatile` :-) See https://www.javamex.com/tutorials/synchronization_volatile.shtml

cdjackson:
@triller-telekom I don't really agree with this. As we discussed this morning, this leaves users open to devices still being in their network, but they never know this. This is IMHO a security risk for starters and I think that users should have visibility that a device still exists. As this currently stands, this is artificially hiding devices that exist in the network. As an example - if a user has a keypad that they want to remove from the network, and they think it's removed, if it is really still usable, this is not good and I think it should be exposed to the user in some way.

As I suggested this morning, we could instead (or as well) add an option to allow this to be enabled or disabled. This would allow Qivicon to disable this feature and hide this security issue from users, but for openHAB I would prefer to allow the user to make this decision.

triller-telekom:
> As an example - if a user has a keypad that they want to remove from the network, and they think it's removed, if it is really still usable

It "somehow" is usable... It *might* be able to still send data to the zigbee network, but not to the openHAB framework, since the corresponding `Thing` is gone. But I see your point.

> As I suggested this morning, we could instead (or as well) add an option to allow this to be enabled or disabled

I will add a configuration option. But not entirely sure where to put it. I would suggest to put it into the configuration for the `ZigBeeDiscoveryService`. Would that be an appropriate solution?

triller-telekom:
I have added an additional commit to make the service configurable. By default the new feature is turned off. So no need to change to have the current behavior.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/discovery/ZigBeeDiscoveryService.java
```diff
@@ -227,4 +249,17 @@ public void run() {
...
+    private void parseAndSetCreateResultsOnlyDuringActiveScansConfigProperty(Map<String, Object> properties) {
+        if (properties != null) {
+            Object property = properties.get(CONFIG_PROPERTY_CREATE_RESULTS_ONLY_DURING_ACTIVE_SCANS);
+            if (property != null) {
+                if (property instanceof Boolean) {
+                    createResultsOnlyDuringActiveScans = (Boolean) property;
+                } else if (property instanceof String) {
+                    createResultsOnlyDuringActiveScans = Boolean.parseBoolean((String) property);
+                }
+            }
```
TomTravaglino: I think we need to handle `property == null`. The property could have been removed after is was true. In that case `createResultsOnlyDuringActiveScans` should be set to true.

triller-telekom: Right, it has to be set to `false` to restore the default behavior. I have amended my commit to reflect this.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/discovery/ZigBeeDiscoveryService.java
```diff
@@ -227,4 +249,20 @@ public void run() {
 
         scheduler.schedule(pollingRunnable, 10, TimeUnit.MILLISECONDS);
     }
+
+    private void parseAndSetCreateResultsOnlyDuringActiveScansConfigProperty(Map<String, Object> properties) {
```
cdjackson: Isn't it possible to use a shorter method name?

triller-telekom: Yes sure, I'll have a look probably tomorrow.

triller-telekom:
I have renamed the long method name in an additional commit and rebased the branch against the current master again.

cdjackson:
Thanks. CI seems to be busted again. I've tried a build but it's not working so I'll try again later...

triller-telekom:
Close and reopen for triggering rebuild

triller-telekom:
Close/reopen to try to retrigger a build again.

cdjackson:
@triller-telekom apparently the issue with CI is related to the version of maven that jenkins is using -:

https://community.openhab.org/t/problems-with-latest-ide-install/76859/2

PR #442, opened 2019-05-17
Adapt the binding to the changes in the ZSS core framework coming with release 1.2.0

triller-telekom:
This PR adjusts the binding to be compatible with the changes introduced into Zsmartsystems by
https://github.com/zsmartsystems/com.zsmartsystems.zigbee/pull/572

~Before this PR can be merged, the dependency to zsmartsystems has to be raised to 1.2.0 as well, see
https://github.com/openhab/openhab-deps-repo/pull/14~

And a release of the 1.2.0 ZSS version has to be deployed.

Also-by: Henning Sudbrock <henning.sudbrock@telekom.de>

### DIFF
--- a/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/ZigBeeNetworkStateSerializerImpl.java
+++ b/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/ZigBeeNetworkStateSerializerImpl.java
@@ -12,6 +12,9 @@
  */
 package org.openhab.binding.zigbee.internal;
 
+import static java.util.Collections.emptyList;
+import static java.util.stream.Collectors.toSet;
+
 import java.io.BufferedReader;
 import java.io.BufferedWriter;
 import java.io.File;
@@ -21,10 +24,12 @@ import java.io.IOException;
 import java.io.InputStreamReader;
 import java.io.OutputStreamWriter;
 import java.nio.file.Files;
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Date;
 import java.text.SimpleDateFormat;
+import java.util.ArrayList;
+import java.util.Date;
+import java.util.List;
+import java.util.Objects;
+import java.util.Set;
 
 import org.eclipse.smarthome.config.core.ConfigConstants;
 import org.openhab.binding.zigbee.ZigBeeBindingConstants;
@@ -36,12 +41,12 @@ import com.thoughtworks.xstream.io.xml.PrettyPrintWriter;
 import com.thoughtworks.xstream.io.xml.StaxDriver;
 import com.zsmartsystems.zigbee.IeeeAddress;
 import com.zsmartsystems.zigbee.ZigBeeNetworkManager;
-import com.zsmartsystems.zigbee.ZigBeeNetworkStateSerializer;
 import com.zsmartsystems.zigbee.ZigBeeNode;
-import com.zsmartsystems.zigbee.dao.ZclClusterDao;
-import com.zsmartsystems.zigbee.dao.ZigBeeEndpointDao;
-import com.zsmartsystems.zigbee.dao.ZigBeeNodeDao;
-import com.zsmartsystems.zigbee.zcl.ZclAttribute;
+import com.zsmartsystems.zigbee.database.ZclAttributeDao;
+import com.zsmartsystems.zigbee.database.ZclClusterDao;
+import com.zsmartsystems.zigbee.database.ZigBeeEndpointDao;
+import com.zsmartsystems.zigbee.database.ZigBeeNetworkDataStore;
+import com.zsmartsystems.zigbee.database.ZigBeeNodeDao;
 import com.zsmartsystems.zigbee.zdo.field.BindingTable;
 import com.zsmartsystems.zigbee.zdo.field.NodeDescriptor.FrequencyBandType;
 import com.zsmartsystems.zigbee.zdo.field.NodeDescriptor.MacCapabilitiesType;
@@ -53,7 +58,7 @@ import com.zsmartsystems.zigbee.zdo.field.PowerDescriptor.PowerSourceType;
  *
  * @author Chris Jackson
  * @author Pedro Garcia - Atomic (mostly) file operations
- * 
+ *
  *         Serialization of the network state to a file is done in the most atomic way portability allows for:
  *         - Serialization is done to a temporary file so that the network state file, if present, is always a complete,
  *         backupable, working version
@@ -69,7 +74,7 @@ import com.zsmartsystems.zigbee.zdo.field.PowerDescriptor.PowerSourceType;
  *
  */
 
-public class ZigBeeNetworkStateSerializerImpl implements ZigBeeNetworkStateSerializer {
+public class ZigBeeNetworkStateSerializerImpl implements ZigBeeNetworkDataStore {
     /**
      * The logger.
      */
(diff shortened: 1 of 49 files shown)


triller-telekom:
@cdjackson Can you please take care of the ZSS 1.2.0 release and the PR I filed on the openhab-deps-repo? Once this is done, you can merge this PR.

kaikreuzer:
> Before this PR can be merged, the dependency to zsmartsystems has to be raised to 1.2.0 as well, see openhab/openhab-deps-repo#14

This repo has already been migrated to bnd, so the dependencies need to be updated within the pom.xmls, like e.g. [here](https://github.com/openhab/org.openhab.binding.zigbee/blob/master/org.openhab.binding.zigbee.cc2531/pom.xml#L23-L28).

triller-telekom:
Thanks Kai.

I have amended my commit to reference the 1.2.0 directly from the pom.xml files.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterGenericButton.java
```diff
@@ -117,7 +126,7 @@ public void commandReceived(ZclCommand command) {
     }
 
     @Override
-    public void attributeUpdated(ZclAttribute attribute) {
+    public void attributeUpdated(ZclAttribute attribute, Object val) {
```
t-8ch: `val` has to be used in `AttributeReportSpec.matches()`.
I think changing `boolean matches(ZclAttribute attribute)` to `boolean matches(ZclAttribute attribute, Object value)` should be fine.
(I can propose a PR next week)

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/ZigBeeNetworkStateSerializerImpl.java
```diff
@@ -8,6 +8,9 @@
  */
```
cdjackson: I think this class is probably too complex now. It still contains the hacks we added to avoid concurrency, but that should not be required now. I'd suggest to use the new ZigBeeDataStore class from ZSS. I think you've partly used that, but also kept the concurrency hacks - do you think this is really needed? One of the main reasons to rewrite the class was to avoid this sort of thing.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterBatteryAlarm.java
```diff
@@ -154,17 +154,17 @@ public Channel getChannel(ThingUID thingUID, ZigBeeEndpoint endpoint) {
...
+    public void attributeUpdated(ZclAttribute attribute, Object val) {
         if (attribute.getCluster() == ZclClusterType.POWER_CONFIGURATION
                 && attribute.getId() == ZclPowerConfigurationCluster.ATTR_BATTERYALARMSTATE) {
 
             logger.debug("{}: ZigBee attribute reports {}", endpoint.getIeeeAddress(), attribute);
 
-            // The value is a 32-bit bitmap, represented by an Integer
-            Integer value = (Integer) attribute.getLastValue();
-            if (value == null) {
+            if (!(val instanceof Integer)) {
```
cdjackson: I see these ```instanceof``` checks throughout this code - do you think that's really necessary? It can only either be (in this case) Integer or null so I'm not sure we need to change all these null checks?

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterColorColor.java
```diff
@@ -475,51 +481,51 @@ public void updateConfiguration(@NonNull Configuration currentConfiguration,
...
-    public void attributeUpdated(ZclAttribute attribute) {
+    public void attributeUpdated(ZclAttribute attribute, Object val) {
         logger.debug("{}: ZigBee attribute reports {}", endpoint.getIeeeAddress(), attribute);
 
         synchronized (colorUpdateSync) {
             try {
                 if (attribute.getCluster().getId() == ZclOnOffCluster.CLUSTER_ID) {
                     if (attribute.getId() == ZclOnOffCluster.ATTR_ONOFF) {
-                        Boolean value = (Boolean) attribute.getLastValue();
+                        Boolean value = (Boolean) val;
```
cdjackson: This seems inconsistent? Most other places where you get the value from the library there is an ```instanceof``` check. Why trust the type here and not everywhere else?

triller-telekom:
Thank you for your review @cdjackson.

This PR (as the original commit is VERY old already) is targeted on making the binding compatible, i.e. able to be compiled against with the newest Zsmartsystems changes.

I agree that one can implements the `ZigBeeNetworkStateSerializerImpl` in a smarter way, now that you have introduced the new `DataStore`. However I would like to keep this PR limited to the changes necessary that the binding still compiles against the new ZSS code. New features from ZSS, like the `DataStore` can then be addressed afterwards as the current code works as before and the PR only adjusts it so it compiles again.

Regarding the `instanceof` checks within the converters I agree with you that it should be done in a consistent way. Given the age of this commit where the PR is based on and the various changes that occurred within ZSS over the time, the consistency suffered a bit unfortunately.
For the `null` checks that we need, IMHO a check for the type is the nicer way to do it, because it ensures that the type we are casting too is correct too, so I think
```
 if (!(val instanceof Integer)) {
       (Integer) val
...
```

looks nicer than

```
if (value == null) {
...
```

Regarding the other point that you mention where I do not check the type or `null` I am a bit unsure:
```
Boolean value = (Boolean) val;
```

If we have to do a `null` check before, IMHO we should also check the type, so this is then missing in the PR.

We are currently using commit b81ff9d43f862a5b5d54d6bc7d7522429e443296 from ZSS and with this Zigbee is working. For all newer commits, to which you are probably referring to, I cannot tell yet.

cdjackson:
> New features from ZSS, like the DataStore can then be addressed afterwards


The DataStore is required to be implemented since the old serialiser is deprecated in the library. I’m not sure how this can be considered optional and this implementation is new? It was also implemented in the library, so the source is already there.  If we are going to implement it then I think we should implement something correctly as this is totally new code and not from the existing codebase - right? I don’t think it makes sense to implement new code “badly” and then say we have to update it later?

>> IMHO a check for the type is the nicer way to do it, because it ensures that the type we are casting too is correct too, so I think
>> 
Maybe, but remember the type should be guaranteed in the library so this is not necessary. If you don’t trust that this is happening then you should also have the type checks in the attribute listeners as well, but this should not be necessary.

Null checks, or type checks in attributeUpdated should not be needed. If you are getting nulls received here, we should resolve this in the library since it makes no sense to say the attribute is updated to null as a device cannot send this.

>> Given the age of this commit where the PR is based on and the various changes that occurred within ZSS over the time, the consistency suffered a bit unfortunately.
>> 
Understood, but now we are making the updates into the binding and again I think this should be done properly. We shouldn’t have a quick hack to get something added and then another PR to do it properly.

triller-telekom:
@cdjackson: Some quick feedback from our way too hot office: I have just rebased our software on the 86a11b422f4ef3e57115fef599f3328b99ce3db0 commit from ZSS and so far it looks like the Ember and telegesis sticks are still working. Also our Ember-SPI bridge is working.

t-8ch:
@triller-telekom Please don't forget https://github.com/triller-telekom/org.openhab.binding.zigbee/pull/1
(The generic button converter is the original reason for the change of the interface in ZSS 1.2, it would be tragic to miss it in the migration)

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/ZigBeeNetworkStateSerializerImpl.java
triller-telekom: What my PR is doing is simply implementing the new interface, but still using the very same "dump" implementation as before, i.e. serializing all nodes instead of just the one which was updated, etc.

So are you suggesting to create a new implementation for your `ZigBeeNetworkDataStore` class?

triller-telekom:
@t-8ch Thanks for mentioning your PR again, I have cherry-picked your commit.

I have also removed the `instanceof` and `null` checks.

cdjackson:
Closed by #454

PR #452, opened 2019-06-29
Add BOM

cdjackson:
@kaikreuzer can you check this please? I'm not completely sure if this is correct, but I think it's better to have a BOM here, and then just reference this BOM in the openhab-deps repo.

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -1,0 +1,59 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.openhab.addons.bundles</groupId>
+    <artifactId>org.openhab.addons.zigbee.reactor</artifactId>
+    <version>2.5.0-SNAPSHOT</version>
+  </parent>
+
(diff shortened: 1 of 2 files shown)


bom/pom.xml
```diff
@@ -0,0 +1,59 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.openhab.addons.bundles<</groupId>
+    <artifactId>org.openhab.addons.zigbee.reactor/artifactId>
```
kaikreuzer:
```suggestion
    <artifactId>org.openhab.addons.zigbee.reactor</artifactId>
```

kaikreuzer: Please check yourself that the project at least builds locally - thanks!

cdjackson:
Sorry - I've not managed to get it to build here, and as I said, I'm not really sure how this new build system works which is why I’ve been trying to get someone to help with this for the past few weeks.  If you have suggestions as to how make it work, then I’d really appreciate it.

I'll close this then - if someone who knows the system can update it, it would be appreciated.

t-8ch:
Few notes:

The versions in the bom should not be properties, as these would be resolved in the importing project.
(maybe the maven release plugin would update these versions too. Or the flatten plugin)

The dependencies in the bom should be wrapped in `pluginManagement`.

In general is a BOM for this project not very useful, as all component share the same version, and can easily be managed with a maven property in the consumer.

cdjackson:
Thanks - are you able to create a PR to get this resolved? It would be appreciated if you could.

PR #453, opened 2019-06-30
Revised BOM

t-8ch:
As discussed in #452 

The only change is wrapping the `dependencies` with `dependencyManagement`

The explicit in-lining of versions was not necessary with my version of maven. I will test it next week with the maven where I thought to remember experiencing this problem.

The test pom is:

```
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
  <modelVersion>4.0.0</modelVersion>

  <groupId>xxx</groupId>
  <artifactId>xxx</artifactId>
  <version>1.0.0-SNAPSHOT</version>

  <packaging>jar</packaging>

  <dependencyManagement>
    <dependencies>
      <dependency>
        <groupId>org.openhab.addons.bundles</groupId>
        <artifactId>org.openhab.binding.zigbee.bom</artifactId>
        <version>2.5.0-SNAPSHOT</version>
        <type>pom</type>
        <scope>import</scope>
      </dependency>
    </dependencies>
  </dependencyManagement>

  <dependencies>
    <dependency>
      <groupId>org.openhab.addons.bundles</groupId>
      <artifactId>org.openhab.binding.zigbee</artifactId>
    </dependency>
    <dependency>
      <groupId>org.openhab.addons.bundles</groupId>
      <artifactId>org.openhab.binding.zigbee.cc2531</artifactId>
    </dependency>
  </dependencies>
</project>
```

Test with `mvn dependency:tree`


bom/pom.xml
```diff
@@ -0,0 +1,61 @@
...
+    <artifactId>org.openhab.addons.zigbee.reactor</artifactId>
+    <version>2.5.0-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>org.openhab.binding.zigbee.bom</artifactId>
+  <packaging>pom</packaging>
+
+  <name>openHAB ZigBee :: BOM :: openHAB ZigBee Binding</name>
+
+  <dependencyManagement>
```
kaikreuzer: Could you elaborate why you said that it need to be a `dependencyManagement` element here instead of defining it directly as dependencies?
If I compare it with the openhab2-addons bom [here](https://github.com/openhab/openhab2-addons/blob/master/bom/openhab-addons/pom.xml), it is done differently (which might explain why the Zigbee binding is actually not available in the app.bndrun list... //cc @triller-telekom).

triller-telekom: Bingo @kaikreuzer !

If I remove the wrapping `<dependencyManagement>` and just keep the `<dependencies>`, it see the zigbee binding in the demo app and can add it to the run configuration (FYI: @cdjackson)

@t-8ch I am also interested why you chose to use the `<dependencyManagement>` tags. What was the reason to use them?

t-8ch: https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#

Documents it this way (search for "bom").
How are you using the bom from the consumer?
A bom is only provides dependency versions, not the dependencies themselves.
You specify a scope=import dependency on the BOM and *in addition* a normal dependency on thing you actually want (without an explicit) version.

If you just want to expose a meta-dependency that pulls in everything it would be an aggregator instead of a bom (I think)

t-8ch: The pom example in this PR shows how to use a bom.
If it is supposed to work like an aggregator the `dependencyManagement` should be removed indeed.

kaikreuzer: @maggu2810 introduced BOMs to openHAB, so he is probably the best to comment here.
From my understanding, the boms are indeed meant to be aggregators and not BOMs as described in the Maven docs that you reference, so I'd suggest to change it here.

maggu2810: @t-8ch I know about that naming in the Maven documentation.
IIRC it used that naming (long ago) because it has been used by other ones, too -- without a big differentiation if provides the one or the other.

If "bill of material" in the Maven world must not be used for other use cases then the one in that documentation I absolutely agree that it should be changed.

t-8ch: @maggu2810 "must not be used" is a bit harsh. It's just that it is commonly understood to be something different.
(Aggregator would actually also be the wrong name, as I found out...)

maggu2810: If you have suggestion for a better name...

EnRoute uses as a directory name "indexes" but that is IMHO used because they generate an OSGi repository index.
https://github.com/osgi/osgi.enroute/tree/master/indexes

So, let me know if there are clever ideas. :wink:

t-8ch: Naming is hard, I prefer to keep out of it :-P
Maybe `meta` (or `meta-dependency`), `all`?

maggu2810: What about `deps` instead of `boms`.

At least it is exactly used that way. Dependencies for compilation  and dependencies for running an application.
Another option would be `tp` for target platform (but IMHO that's a Eclipse PDE specific naming).

WDYT about `deps`?

t-8ch: Sounds reasonable. Another one I have would be `all-modules` or `all-components`.

maggu2810: @t-8ch While the OSGi projects I am working for is growing steadily, I started to publish an OSGi playground repository.
I used the naming `deps` now as it consists of the dependencies / dependency chain for specific "target" (e.g. to depend on compilation, to depend on runtime, for tests, ...).

PR #456, opened 2019-07-04
removed dependencyManagement section

kaikreuzer:
As discussed in https://github.com/openhab/org.openhab.binding.zigbee/pull/453#discussion_r300365311


cdjackson:
@kaikreuzer thanks for looking into this. I understood from speaking to @triller-telekom yesterday that with this change, and the updated ```openhab-distro```, that this should now work? Unfortunately at least for me it is still not listing the zigbee projects in the bnd app.  I've tried updating all maven poms in Eclipse, and refreshed, cleaned and restarted, but it still is not working here.

Is there something else missing - other than using this PR to restore the dependendies, and the latest ```openhab-distro```?

Thanks
Chris

kaikreuzer:
@cdjackson Not sure what needs to be done to test this locally. Easiest thing would be to merge this PR and I'd assume that the zigbee bundles then become available in the list, since this is the case for all other addons as well.

cdjackson:
I’ll merge this anyway, but as this is just merging back to how I originally defined it earlier I’m not sure that will help? What I have currently done is to fork this branch locally - merging it will be the same as there’s only been one other commit to master and that is documentation.

cdjackson:
As expected, merging this hasn't helped - I still don't get the zigbee bundles listed in the IDE.

kaikreuzer:
You will have to wait for the next snapshot build.

cdjackson:
Hmmm - ok, thanks. What snapshot creates this - is it the distro?

Strangely I was able to test this locally with other repositories I’ve updated without it requiring a snapshot build...

kaikreuzer:
The build that builds this PR publishes the snapshot of the bom, which is then referenced by the IDE, i.e. the openhab-bundles build. It just succeeded a few minutes ago.

triller-telekom:
What helped me was to do re-create the dependencies locally as @t-8ch mentioned in his PR: https://github.com/openhab/org.openhab.binding.zigbee/pull/453#issue-293095348

> Test with `mvn dependency:tree`

After that my IDE had picked up the zigbee bundles in the example app.

cdjackson:
I think my system picks up the bundles now, but they don’t resolve. There’s another issue where these are being tracked in the openhab-distro repository.

https://github.com/openhab/openhab-distro/issues/930#issuecomment-509008898 <https://github.com/openhab/openhab-distro/issues/930#issuecomment-509008898>

PR #457, opened 2019-07-05
Added details about allowed values in configuration file

bkummel:


org.openhab.binding.zigbee/README.md
```diff
@@ -18,6 +18,18 @@ demo.things:
 Thing zigbee:coordinator_cc2531:stick1 "Zigbee USB Stick" [zigbee_port="/dev/ttyACM0", zigbee_baud="38400"]
 ```
 
+#### Serial port Configuration
+
+If you are using a config file, the serial port needs to be configured as shown in the above example. Below is a list of the relevat settings concerning the serial port.
+
+| Setting               | Allowed values  |
+| --------------------- | --------------- |
+| `zigbee_baud`         | `38400`, `57600`, `115200` |
```
cdjackson: This is not correct. The allowable values are not really defined and are dependant on what coordinator you're using.

org.openhab.binding.zigbee/README.md
```diff
@@ -18,6 +18,18 @@ demo.things:
...
+| `zigbee_flowcontrol ` | `0` = None, `1` = Hardware (CTS/RTS), `2` = Software (XOn/XOff) |
+| `zigbee_port`         | The name of the serial port, e.g. `/dev/ttyUSB1` or `/dev/ttyZigbee` or whatever the port is named on your system |
+
+Note that these serial port configuration values should be provided as strings in the config file, that is, inside double quotes, as shown in the example above.
```
cdjackson: I'm surprised that this is the case - the baud rate is defined as an integer in the configuration file, so it should not contain the quotes.

org.openhab.binding.zigbee/README.md
```diff
@@ -43,16 +55,52 @@ Once a child is removed from the child table of a router, it will be asked to re
 
 Note that ZigBee compliant devices should rejoin the network seamlessly, however some non-compliant devices may not rejoin which may leave them unusable without a manual rejoin.
 
+**Values:** Timeout time in seconds. The table below lists the options that are shown in PaperUI and the equivalent values that can be set in a configuration file:
```
cdjackson: I guess you could simply state that this is notionally defined in seconds rather than defining how many seconds are in each period?

org.openhab.binding.zigbee/README.md
```diff
@@ -43,16 +55,52 @@ Once a child is removed from the child table of a router, it will be asked to re
...
+
+- `0` = Low RAM concentrator
+- `1` = High RAM concentrator
+
+
 ##### Mesh Update Period (zigbee_meshupdateperiod)
 
 The binding is able to search the network to get a list of what devices can communicate with other devices. This is a useful diagnostic feature as it allows users to see the links between devices, and the quality of these links. However, this can generate considerable traffic, and some battery devices may not poll their parents often enough to provide these updates, and users may consider that it is better to reduce the period, or disable this feature.
 
+**Value:** the update period in seconds. ```0``` means "never update". In PaperUI, a drop down list is shown, the options from that list are shown in te table below, with their equivalent values that can be put in a config file.
```
cdjackson: I guess you could simply state that this is notionally defined in seconds rather than defining how many seconds are in each period?

org.openhab.binding.zigbee/README.md
```diff
@@ -43,16 +55,52 @@ Once a child is removed from the child table of a router, it will be asked to re
...
+
 ##### Concentrator Type (zigbee_concentrator)
 
 The Concentrator is used to improve routing within a ZigBee network, and is especially useful in a network where much of the traffic is sent to or from a central coordinator. If the coordinator has sufficient memory, it can store routing information, thus reducing network traffic.
 
 If supported, the High RAM concentrator should be used.
 
+**Values:**
+
+- `0` = Low RAM concentrator
```
cdjackson: Can I suggest to use a table the same as other configs?

org.openhab.binding.zigbee/README.md
bkummel: You are right. It works with and without quotes, so it seems. I'll remove that note.

bartkummel:
I created a new PR, for technical reasons. I have processed the review comments in the new PR. This one can be closed.

PR #468, opened 2019-08-10
Illuminance configuration improvements

cdjackson:
This PR includes changes to the reporting configuration class ```ZclReportingConfig```. This now supports changing the reporting change, and also the reporting min/max times.

@triller-telekom I'd welcome your review on this.

Note that there will be a subsequent PR to improve the handling of reporting configuration which is  possible with the way reporting is now handled in the latest ZSS, but I'd like to get this merged as it already covers more than one feature.

### DIFF
--- a/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/config/ZclReportingConfig.java
+++ b/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/config/ZclReportingConfig.java
@@ -70,16 +70,16 @@ public class ZclReportingConfig implements ZclClusterConfigHandler {
     public ZclReportingConfig(Channel channel) {
         Configuration configuration = channel.getConfiguration();
         if (configuration.containsKey(CONFIG_REPORTINGMIN)) {
-            reportingTimeMin = Integer.parseInt((String) configuration.get(CONFIG_REPORTINGMIN));
+            reportingTimeMin = ((BigDecimal) configuration.get(CONFIG_REPORTINGMIN)).intValue();
         }
         if (configuration.containsKey(CONFIG_REPORTINGMIN)) {
-            reportingTimeMin = Integer.parseInt((String) configuration.get(CONFIG_REPORTINGMAX));
+            reportingTimeMin = ((BigDecimal) configuration.get(CONFIG_REPORTINGMAX)).intValue();
         }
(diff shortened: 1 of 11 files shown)


cdjackson:
I would prefer not to have to waste time refactoring to be honest. The development of these features was interlinked which is why they ended up in a single branch and splitting them seems unnecessary even if in an ideal world we should only have one change per branch.

Is that ok? Otherwise it will be next week before I can look at this I think.

cdjackson:
Updated.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/config/ZclReportingConfig.java
```diff
@@ -66,14 +67,30 @@
     private BigDecimal minimumChange;
     private BigDecimal maximumChange;
 
+    public ZclReportingConfig(Channel channel) {
+        Configuration configuration = channel.getConfiguration();
+        if (configuration.containsKey(CONFIG_REPORTINGMIN)) {
+            reportingTimeMin = ((BigDecimal) configuration.get(CONFIG_REPORTINGMIN)).intValue();
+        }
+        if (configuration.containsKey(CONFIG_REPORTINGMIN)) {
```
triller-telekom: I think there is a copy&paste error here: should be `CONFIG_REPORTINGMAX`

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/config/ZclReportingConfig.java
```diff
@@ -66,14 +67,30 @@
...
+            reportingTimeMin = ((BigDecimal) configuration.get(CONFIG_REPORTINGMAX)).intValue();
+        }
+        if (configuration.containsKey(CONFIG_REPORTINGMIN)) {
+            reportingTimeMin = ((BigDecimal) configuration.get(CONFIG_REPORTINGCHANGE)).intValue();
```
triller-telekom: I think there is a copy&paste error here: should be `CONFIG_REPORTINGCHANGE`

cdjackson:
> Then, this PR contains the illuminance fix and a preparation of the channelconverter to support reporting intervals as a channel configuration, I am fine with that.

No - I split the illuminance fix into #469 

I'll update with the errors - well spotted, thanks.

cdjackson:
Thanks Stefan

Issue #474, opened 2019-08-30
[Telegesis] Zigbee binding stops working after some time


yoadey:
### Outline

The zigbee binding stops working after some commands are send, afterwards only a `java.util.concurrent.RejectedExecutionException: Task com.zsmartsystems.zigbee.dongle.telegesis.ZigBeeDongleTelegesis$2@107f71 rejected from java.util.concurrent.ThreadPoolExecutor occurs.
` appears.

This was reproducible at least 5 times, to work with the binding again openhab has to be restarted.
If no command is sent, binding runs fine for several days. After a few commands being send the mentioned error occurs.
The commands which are send are processed normally, the lights switch as requested.
Please see logs for more details.

I think this happened already with 2.4.0, but I didn't analyzed the issue there any further as I switched to another gateway.

### Configuration

| Configuration    | Description                                                                    |
|------------------|--------------------------------------------------------------------------------|
| Coordinator used | Qivicon Zigbee USB Stick (Telegesis)                                   |
| openHAB version  | 2.5.0-M2     |
| Hardware         |  Synology Diskstation ds213j                                                                 |
| Memory           |  512MB                                                                         |
| Java version     |   Java(TM) SE Embedded Runtime Environment (build 1.8.0_121-b13)                         |
| Devices          | 6 Osram RGBW gu10 bulbs, 1 Osram e27 bulb, 1 Osram flex strip  |

### Logs

[openhab.log](https://github.com/openhab/org.openhab.binding.zigbee/files/3559346/openhab.log)

cdjackson:
This is not really a problem as far as I can see. This issue is caused by the binding (or something in the system) sending a command AFTER the system has been shut down. This exception is simply stating that the command is rejected as the system is shutdown.

The bigger question is why the system is shutdown - it should restart, but I don't see that in the log. Maybe it's not implemented for Telegesis.

yoadey:
With your feedback I've analysed the code and I think, the issue is in TelegesisFrameHandler: the variables retries is increased, but never set back to 0 if the transaction was successfully completed.
According to the comment at ACK_TIMEOUTS it should be consecutive fails, but currently it is 2 timeouts during whole run.

If my conclusions are correct, the stopTimer() method should be changed to the following:
```
    private void stopTimer() {
        if (timeoutTimer != null) {
            logger.trace("TELEGESIS Timer: Stop");
            timeoutTimer.cancel(false);
            timeoutTimer = null;
            retries = 0;
[...]

PR #477, opened 2019-09-06
Create new ZSS transport and config on handler restart

triller-telekom:
Relates to #474


cdjackson:
Thanks for this @triller-telekom. 

IMHO this needs a little additional refactoring. With this PR, ```initialize``` and ```reinitialiseZigBee``` do the same thing - they are duplicated (except in the case of the XBee where the ```reinitialiseZigBee``` is empty so presumably won't work!).

I would suggest that instead of overloading the ```initialize``` method in each dongle which calls ```super.initialize```, you should instead remove the ```initialize``` method which now seems to only calls ```createDongle```, and just call ```reinitialiseZigBee``` from ```ZigBeeCoordinatorHandler.initialize()```.

This would simplify things and avoid duplication. WDYT?

cdjackson:
@triller-telekom any progress on this?

triller-telekom:
Hi Chris,

sorry that I haven't updated this PR since a while and that I am also replying late. Unfortunately I am very busy with other stuff at the moment, which is why I haven't updated this PR.

If you like to take over feel free, otherwise I can have a look once I find some time. Sorry.

cdjackson:
Hi Stefan,
Ok, it’s no problem for me - I think you guys probably have more need to resolve this issue than me at the moment :)

Cheers
Chris

cdjackson:
Replaced by #564

PR #484, opened 2019-09-15
Fix bug setting level control reporting configuration

cdjackson:
Fix incorrect reference when configuring reporting. This doesn't really change anything since these all have the same value, but is clearly referencing the wrong constant.


cdjackson:
@triller-telekom

cdjackson:
Thanks Stefan

PR #497, opened 2019-10-01
Ensure leave command is send to the device

triller-telekom:
Once a Thing is about to be removed, we send out a leave command, however
this is done asynchronously. Meanwhile we delete the node from the
ZigbeeNetworkManager. Now the ZigbeeTransactionManager wants to dispatch the
leave command over the air and fails to do so because the destination is unknown

ZigBeeTransactionManager] - Attempt to send command with unknown destination: ZigBeeTransaction [queueTime=6, state=DISPATCHED, sendCnt=1, command=ManagementLeaveRequest [0/0 -> 23592/0, cluster=0034, TID=7A, deviceAddress=00124B00043D6DE1, removeChildrenRejoin=false]]

The removeNode call is unneccessary because if the leave command gets a reply
the node will be removed anyhow.

### DIFF
--- a/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeThingHandler.java
+++ b/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeThingHandler.java
@@ -637,7 +637,6 @@ public class ZigBeeThingHandler extends BaseThingHandler implements ZigBeeNetwor
     @Override
     public void handleRemoval() {
         coordinatorHandler.leave(nodeIeeeAddress);
-        coordinatorHandler.removeNode(nodeIeeeAddress);
         updateStatus(ThingStatus.REMOVED);
     }
 


cdjackson:
If you do this, the node will not be removed from the network. The ZSS libraries will not remove the ```ZigBeeNode``` instance from the lists and when you do another search the device will end up back in the inbox.

triller-telekom:
Why won't it be deleted?

The `coordinatorHandler.leave(nodeIeeeAddress);` call lead to a call of `ZigBeeNetworkManager#leave` which then will finally call:

https://github.com/zsmartsystems/com.zsmartsystems.zigbee/blob/master/com.zsmartsystems.zigbee/src/main/java/com/zsmartsystems/zigbee/ZigBeeNetworkManager.java#L1267

So the node will be removed if it replied to the leave command.

if it does not reply to the leave command, the node stay and will appear in the inbox again, right. But I guess this is expected and OK, since the device is still part of the zigbee network.

cdjackson:
Ok, in this instance this is correct - I was thinking this was acting on the leave notification which no longer removes a device from the network list.

PR #500, opened 2019-10-05
Add support for IncreaseDecreaseType in LevelControl

cdjackson:
Adds support for the ```LevelControlCluster``` ```Move``` command. This really requires https://github.com/openhab/openhab-core/pull/1097 to allow support of the ```Stop``` command otherwise it's not of use, so I don't plan to merge this until there is a mechanism for stopping the travel.


cdjackson:
Since the ```Stop``` command in OH has been rejected, this feature requires that the user repeatedly send the ```IncreaseDecrease``` command. If an ```IncreaseDecrease``` command is not received within 200ms the ZigBee ```stop``` command will be sent.

PR #509, opened 2019-10-18
Fix problems reading from serial port.

sesuncedu:
This PR addresses a group of interconnected issues when reading from the serial port. 

#### Bytes being read one-at-a-time from the native serial port

The first reason for this is the setting of a read threshold on the serial port to 1. 
This causes the input stream to return after reading at most 1 byte, regardless of how many are  available. 

The second reason is that bytes were being read from  inputstream one at a time.  This results in a separate read for each byte.

#### Data always read  reading until a timeout occurred when processing data-received events 

After receiving a notification that data is available for reading, the event handler did not check to see how many bytes were available, and read continuously until halted by a timeout. Since this was performed whilst a lock was being held, no bytes were made available to the application until no bytes had been available for the configured timeout period. This could cause readers to block indefinitely. Also, the number of bytes that could be read was unbounded. 

#### No checks for buffer overrun
If the circular buffer capacity was exceeded, this  was not detected.  The write pointer was advanced and the read pointer  left unmodified. This caused all content read up until the overrun  to be discarded, including data that might be part of the current frame being read.  A  typical behavior for serial ports is to use a head-drop strategy, with the oldest bytes being discarded . 

 ### Impact
The following log extracts shows the Ember drivers periodic keep-alive activity, which sends an EZSP NetworkStateRequest, which generates a NetworkStateResponse.

The first line in each trace is generated immediately before the request frame is written to the port.
The second line is generated after the response frame has been received. 

Before the changes, the time per transaction is 0.110ms. 
After the changes, the time per transaction is 0.014ms 

Note that there are no changes to write behavior in this PR.  Data is still written to the file descriptor one byte at a time. This is suboptimal, but is not as significant as the read issues. 

#### Before changes
```
2019-10-18 14:04:56.129 [DEBUG] [e.ember.internal.ash.AshFrameHandler] - --> TX ASH frame: AshFrameData [frmNum=7, ackNum=7, reTx=false, data=D1 00 18]
2019-10-18 14:04:56.239 [DEBUG] [e.ember.internal.ash.AshFrameHandler] - <-- RX ASH frame: AshFrameData [frmNum=7, ackNum=0, reTx=false, data=D1 80 18 02]
2019-10-18 14:04:56.249 [DEBUG] [e.ember.internal.ash.AshFrameHandler] - ASH: Frame acked and removed AshFrameData [frmNum=7, ackNum=7, reTx=false, data=D1 00 18]
2019-10-18 14:04:56.252 [WARN ] [zigbee.dongle.ember.ZigBeeDongleEzsp] - Unhandled EZSP response received: EzspNetworkStateResponse [status=EMBER_JOINED_NETWORK]
2019-10-18 14:04:56.254 [DEBUG] [e.ember.internal.ash.AshFrameHandler] - --> TX ASH frame: AshFrameAck [ackNum=0, notRdy=false]
```

#### After changes
```
15:21:03.007 [DEBUG] [le.ember.internal.ash.AshFrameHandler] - --> TX ASH frame: AshFrameData [frmNum=3, ackNum=1, reTx=false, data=EF 00 18]
15:21:03.021 [DEBUG] [le.ember.internal.ash.AshFrameHandler] - <-- RX ASH frame: AshFrameData [frmNum=1, ackNum=4, reTx=false, data=EF 80 18 02]
15:21:03.025 [DEBUG] [le.ember.internal.ash.AshFrameHandler] - ASH: Frame acked and removed AshFrameData [frmNum=3, ackNum=1, reTx=false, data=EF 00 18]
15:21:03.030 [WARN ] [.zigbee.dongle.ember.ZigBeeDongleEzsp] - Unhandled EZSP response received: EzspNetworkStateResponse [status=EMBER_JOINED_NETWORK]
15:21:03.034 [DEBUG] [le.ember.internal.ash.AshFrameHandler] - --> TX ASH frame: AshFrameAck [ackNum=2, notRdy=false]
```

### DIFF
--- a/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeSerialPort.java
+++ b/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeSerialPort.java
@@ -280,7 +280,7 @@ public class ZigBeeSerialPort implements ZigBeePort, SerialPortEventListener {
                 synchronized (bufferSynchronisationObject) {
                     int available = inputStream.available();
                     logger.trace("Processing DATA_AVAILABLE event: have {} bytes available", available);
-                    byte[] buf = new byte[available];
+                    byte buf[] = new byte[available];
                     int offset = 0;
                     while (offset != available) {
                         if (logger.isTraceEnabled()) {
@@ -293,11 +293,11 @@ public class ZigBeeSerialPort implements ZigBeePort, SerialPortEventListener {
                                     available - offset, offset);
                         }
                         if (n <= 0) {
-                            logger.error("Processing DATA_AVAILABLE event: Expected to be able to read " + available
-                                    + " bytes, but saw error after " + offset);
-                            throw new IOException(
+                            final String message =
                                     "Expected to be able to read " + available + " bytes, but saw error after "
-                                            + offset);
+                                            + offset;
+                            logger.warn("Processing DATA_AVAILABLE event: {}", message);
+                            throw new IOException(message);
                         }
                         offset += n;
                     }
--- a/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeSerialPort.java
+++ b/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeSerialPort.java
@@ -293,11 +293,9 @@ public class ZigBeeSerialPort implements ZigBeePort, SerialPortEventListener {
                                     available - offset, offset);
                         }
                         if (n <= 0) {
-                            final String message =
+                            throw new IOException(
                                     "Expected to be able to read " + available + " bytes, but saw error after "
-                                            + offset;
-                            logger.warn("Processing DATA_AVAILABLE event: {}", message);
-                            throw new IOException(message);
+                                            + offset);
                         }
                         offset += n;
                     }
@@ -316,7 +314,7 @@ public class ZigBeeSerialPort implements ZigBeePort, SerialPortEventListener {
                     }
                 }
             } catch (IOException e) {
-                logger.debug("Masking IOException in serial port event", e);
+                logger.warn("Processing DATA_AVAILABLE event: received IOException in serial port event", e);
             }
 
             synchronized (this) {
--- a/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeSerialPort.java
+++ b/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeSerialPort.java
@@ -159,7 +159,6 @@ public class ZigBeeSerialPort implements ZigBeePort, SerialPortEventListener {
                         break;
                 }
 
-                serialPort.enableReceiveThreshold(1);
                 serialPort.enableReceiveTimeout(100);
                 serialPort.addEventListener(this);
                 serialPort.notifyOnDataAvailable(true);
@@ -279,15 +278,45 @@ public class ZigBeeSerialPort implements ZigBeePort, SerialPortEventListener {
         if (event.getEventType() == SerialPortEvent.DATA_AVAILABLE) {
             try {
                 synchronized (bufferSynchronisationObject) {
-                    int recv;
-                    while ((recv = inputStream.read()) != -1) {
-                        buffer[end++] = recv;
+                    int available = inputStream.available();
+                    logger.trace("Processing DATA_AVAILABLE event: have {} bytes available", available);
+                    byte[] buf = new byte[available];
+                    int offset = 0;
+                    while (offset != available) {
+                        if (logger.isTraceEnabled()) {
+                            logger.trace("Processing DATA_AVAILABLE event: try read  {} at offset {}",
+                                    available - offset, offset);
+                        }
+                        int n = inputStream.read(buf, offset, available - offset);
+                        if (logger.isTraceEnabled()) {
+                            logger.trace("Processing DATA_AVAILABLE event: did read {} of {} at offset {}", n,
+                                    available - offset, offset);
+                        }
+                        if (n <= 0) {
+                            logger.error("Processing DATA_AVAILABLE event: Expected to be able to read " + available
+                                    + " bytes, but saw error after " + offset);
+                            throw new IOException(
+                                    "Expected to be able to read " + available + " bytes, but saw error after "
+                                            + offset);
+                        }
+                        offset += n;
+                    }
+                    for (int i = 0; i < available; i++) {
+                        buffer[end++] = buf[i] & 0xff;
                         if (end >= RX_BUFFER_LEN) {
                             end = 0;
+                        }
+                        if (end == start) {
+                            logger.warn("Processing DATA_AVAILABLE event: Serial buffer overrun");
+                            if (++start == RX_BUFFER_LEN) {
+                                start = 0;
+                            }
+
                         }
                     }
                 }
             } catch (IOException e) {
+                logger.debug("Masking IOException in serial port event", e);
             }
 
             synchronized (this) {


cdjackson:
Thanks @sesuncedu 

I guess you realise, but just for clarity - this is only applicable for the demo console app. Of course, other implementations (eg OH) may (ie does!) have the same issue.

I will take a better look at this tomorrow, but it looks ok at first glance. Please can you sign this PR / commit though.

sesuncedu:
This IS openhab!

sesuncedu:
I have made no attempt at fixing the default driver

cdjackson:
Ah - ok, I thought you’d pushed this into the ZSS libs - sorry.

sesuncedu:
Addressing the write issue unfortunately requires changing both, and also  requires a minor version bump on the com.zsmartsystems.zigbee.transport package (which thus requires  a minor version bump  in the  com.zsmartsystems.zigbee bundle version)

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeSerialPort.java
```diff
@@ -279,15 +278,45 @@ public void serialEvent(SerialPortEvent event) {
...
+                            logger.trace("Processing DATA_AVAILABLE event: try read  {} at offset {}",
+                                    available - offset, offset);
+                        }
+                        int n = inputStream.read(buf, offset, available - offset);
+                        if (logger.isTraceEnabled()) {
+                            logger.trace("Processing DATA_AVAILABLE event: did read {} of {} at offset {}", n,
+                                    available - offset, offset);
+                        }
+                        if (n <= 0) {
+                            logger.error("Processing DATA_AVAILABLE event: Expected to be able to read " + available
```
cdjackson: Please can we drop the logging level down to debug (or possibly warn). Also, please use logging with individual parameters rather than string concatenation.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeSerialPort.java
```diff
@@ -279,15 +278,45 @@ public void serialEvent(SerialPortEvent event) {
...
+                        }
+                        offset += n;
+                    }
+                    for (int i = 0; i < available; i++) {
+                        buffer[end++] = buf[i] & 0xff;
                         if (end >= RX_BUFFER_LEN) {
                             end = 0;
                         }
+                        if (end == start) {
+                            logger.warn("Processing DATA_AVAILABLE event: Serial buffer overrun");
```
cdjackson: I'd suggest to drop this down to debug since the system will recover from this.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeSerialPort.java
cdjackson: In fact I'd probably suggest to remove this log entry since there is an exception that also provides the same information.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeSerialPort.java
sesuncedu: > I'd suggest to drop this down to debug since the system will recover from this.

That's why warn is appropriate instead of error -  an unusual event, which may cause user visible anomalous behavior, and which may indicate more serious problems, but for which automatic recovery can be performed.

cdjackson: Fine - let's leave this one, but please resolve the other one.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeSerialPort.java
sesuncedu: This situation ought never to occur; the amount being read has been determined to be available, so if it can't be read it probably means that another thread or process is simultaneously  reading from the port, which is likely more serious than a warning.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeSerialPort.java
sesuncedu: I would be uncomfortable lowering the priority.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeSerialPort.java
cdjackson: Within the OH context though, this is only meant for issues that are system wide. I would really prefer to reduce this level, or simply remove it since there is already an exception thrown which is caught and logged so there seems little point in logging the same thing twice does there?

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeSerialPort.java
cdjackson: As I said, let's leave this as it is.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeSerialPort.java
sesuncedu: I don't know what other  IOExceptions, if any, might have previously been   masked here. 

But I will  stop logging this error as an error, and make the logging statement in the catch clause log at warning. 

```
/var/log/openhab2/openhab.log.5:2019-10-18 14:55:34.553 [ERROR] [ding.zigbee.handler.ZigBeeSerialPort] - Serial Error: Port /dev/ttyUSB1 in use.
```

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeSerialPort.java
```diff
@@ -279,15 +278,45 @@ public void serialEvent(SerialPortEvent event) {
...
+                        int n = inputStream.read(buf, offset, available - offset);
+                        if (logger.isTraceEnabled()) {
+                            logger.trace("Processing DATA_AVAILABLE event: did read {} of {} at offset {}", n,
+                                    available - offset, offset);
+                        }
+                        if (n <= 0) {
+                            final String message =
+                                    "Expected to be able to read " + available + " bytes, but saw error after "
+                                            + offset;
+                            logger.warn("Processing DATA_AVAILABLE event: {}", message);
```
cdjackson: I see you've updated this, but I still don't understand why there is a log statement, and then throwing this as an exception which logs exactly the same log statement? Is there really a need for logging this twice?

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeSerialPort.java
cdjackson: Thanks, but we are still logging this statement twice - it's exactly the same string that is logged, then thrown as an exception and caught a few lines further down and then logged again. I don't think there is there really a need for this duplication?

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeSerialPort.java
sesuncedu: One place was logging a known error level error (as a warning now). The exception was thrown to take things into the exception handling path. 
The exception logging was previously only at debug, since I didn't know what else might have been throwing them, and they previously weren't logged at all. 
They are now logged as warnings, and the separate higher priority logging removed.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeSerialPort.java
sesuncedu: The change removed separately logging the error as an warning, and now logs all IOExceptions as warnings.

cdjackson:
Thanks @sesuncedu

PR #517, opened 2019-11-14
Add methods to get supported cluster list from converters

cdjackson:
This PR is incomplete and will fail CI - do not merge!!

From recent versions of the Z-Smart Systems ZigBee library (1.2.6 I think), we need to tell the framework what clusters we support. This is to allow the framework to only process messages that the application supports, and more importantly, to reject ones that are not supported with the appropriate error. 

***This is required for ZigBee certification and there is a test for this***.

To achieve this, I've added two methods to the converters, and the converter factory to allow the coordinator handler to get the list of clusters that are implemented by the system. Each cluster will return a Set of cluster IDs it supports through the ```Set<Integer> getSupportedClientClusters()``` and ```Set<Integer> getSupportedServerClusters()``` methods. The converter factory has a similar set of methods and will return the consolidated Set of cluster IDs supported by all converters so that this can be passed to the framework.

@hsudbrock or @triller-telekom you should review this concept please.

Design questions - I have one thought - the framework methods set the client/server clusters the framework supports, but the converters internally use client/server clusters on the remote device, so the naming is reversed. Within the converter, does the ```getSupportedServerClusters``` method return the server clusters consumed by the converter, or the server clusters implemented by the converted?

---

This PR currently only implements the initial concept, and the methods in a single converter for comment.

Note that we probably also need to add the converter factory provider to the coordinator factories as this information is required in the ```ZigBeeCoordinatorHandler``` class (I've added this as a parameter on the constructor, but not hooked this all together yet).

### DIFF
--- a/org.openhab.binding.zigbee.cc2531/src/main/java/org/openhab/binding/zigbee/cc2531/internal/CC2531HandlerFactory.java
+++ b/org.openhab.binding.zigbee.cc2531/src/main/java/org/openhab/binding/zigbee/cc2531/internal/CC2531HandlerFactory.java
@@ -30,6 +30,7 @@ import org.eclipse.smarthome.core.thing.binding.ThingHandlerFactory;
 import org.eclipse.smarthome.io.transport.serial.SerialPortManager;
 import org.openhab.binding.zigbee.cc2531.CC2531BindingConstants;
 import org.openhab.binding.zigbee.cc2531.handler.CC2531Handler;
+import org.openhab.binding.zigbee.converter.ZigBeeChannelConverterFactory;
 import org.openhab.binding.zigbee.handler.ZigBeeCoordinatorHandler;
 import org.osgi.framework.ServiceRegistration;
 import org.osgi.service.component.annotations.Activate;
@@ -53,9 +54,21 @@ public class CC2531HandlerFactory extends BaseThingHandlerFactory {
 
     private final SerialPortManager serialPortManager;
 
+    @Nullable
+    private ZigBeeChannelConverterFactory zigbeeChannelConverterFactory;
+
     @Activate
     public CC2531HandlerFactory(final @Reference SerialPortManager serialPortManager) {
         this.serialPortManager = serialPortManager;
+    }
+
+    @Reference
+    protected void setZigBeeChannelConverterFactory(ZigBeeChannelConverterFactory zigbeeChannelConverterFactory) {
+        this.zigbeeChannelConverterFactory = zigbeeChannelConverterFactory;
+    }
+
+    protected void unsetZigBeeChannelConverterFactory(ZigBeeChannelConverterFactory zigbeeChannelConverterFactory) {
+        this.zigbeeChannelConverterFactory = null;
     }
 
     @Override
@@ -69,7 +82,7 @@ public class CC2531HandlerFactory extends BaseThingHandlerFactory {
 
         ZigBeeCoordinatorHandler coordinator = null;
         if (thingTypeUID.equals(CC2531BindingConstants.THING_TYPE_CC2531)) {
-            coordinator = new CC2531Handler((Bridge) thing, serialPortManager);
+            coordinator = new CC2531Handler((Bridge) thing, serialPortManager, zigbeeChannelConverterFactory);
         }
 
         if (coordinator != null) {
--- a/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeCoordinatorHandler.java
+++ b/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeCoordinatorHandler.java
@@ -447,9 +447,9 @@ public abstract class ZigBeeCoordinatorHandler extends BaseBridgeHandler
 
         // Add all the clusters that we are supporting.
         // If we don't do this, the framework will reject any packets for clusters we have not stated support for.
-        channelFactory.getSupportedClientClusters().stream()
+        channelFactory.getImplementedClientClusters().stream()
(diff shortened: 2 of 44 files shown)


cdjackson:
I have renamed ```getSupportedClientClusters``` and ```getSupportedServerClusters``` to ```getImplementedClientClusters``` and ```getImplementedServerClusters``` to try and make it easier to understand that the definition is about what is implemented in the converter. Otherwise there is confusion about "supported" - if it means supported by the converter, or supported by the remote device, since they are opposite.

cdjackson:
I've now added the ```ZigBeeChannelConverterFactory``` to the Ember coordinator factory. Other than updating the other coordinators, and all of the converters to add the implemented clusters, this should be complete.

@hsudbrock or @triller-telekom I would really appreciate it if you can take a quick look at this ASAP as I would like to get this included into the latest OH for the 2.5 release so we can update the ZSS libraries. (Thanks).

triller-telekom:
Hi Chris,

sorry to get back to you this late again. I wish I could use more time on this here...

While seeing the latest changes in ZSS I also thought about a way that the converters should announce their used clusters by themselves. It certainly makes sense to define the clusters where we use them, so in the converters is a good idea.

I also like "implemented" over  "supported" due to the confusion mentioned by you.

However, what we should not miss, is those clusters which are used by instances of `com.zsmartsystems.zigbee.app.ZigBeeApplication`. Those have to be registered to the `ZigbeeNetworkManager` as well. For example the `ZclIasZoneClient` which is in ZSS.

I don't know if you plan to register them within ZSS, or if any program who wants to make use of these applications has to register the clusters.

cdjackson:
Thanks for your thoughts @triller-telekom.

> I don't know if you plan to register them within ZSS, or if any program who wants to make use of these applications has to register the clusters.

Yes, there is a plan to do this, but it's not all joined up - https://github.com/zsmartsystems/com.zsmartsystems.zigbee/issues/411 is basically linked to this. In the first instance I might add these explicitly to the binding to make sure it's registered, but the above issue is becoming more important now that we have code to reject packets received from clusters that aren't registered...

You will find as well that there are now 2 APIs - one to set the clusters in the```ZigBeeManager```, and another that is setting the clusters in the dongle and these should be aligned. In the past we never bothered to set the clusters in the Ember dongle as it was not really required, but you may find that certification requires this ;)

I will progress with this then and try and finish it tonight or tomorrow - I think it's all done now so it just needs testing. Do you want to review? Or should I ask - do you have time to review ;)

triller-telekom:
> Do you want to review? Or should I ask - do you have time to review ;)

Rather the later, but I can try to have a deeper look at this PR during this week. Sorry can't promise.

org.openhab.binding.zigbee.cc2531/src/main/java/org/openhab/binding/zigbee/cc2531/internal/CC2531HandlerFactory.java
```diff
@@ -47,6 +49,18 @@
     private static final Set<ThingTypeUID> SUPPORTED_THING_TYPES_UIDS = Collections
             .singleton(CC2531BindingConstants.THING_TYPE_CC2531);
 
+    @Nullable
+    private ZigBeeChannelConverterFactory zigbeeChannelConverterFactory;
+
+    @Reference
```
t-8ch: Adding the `@Reference` directly on the field would remove the need for the `set` and `unset` methods.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterAtmosphericPressure.java
```diff
@@ -50,6 +54,16 @@
 
     private ZclPressureMeasurementCluster cluster;
 
+    @Override
+    public Set<Integer> getImplementedClientClusters() {
+        return Stream.of(ZclPressureMeasurementCluster.CLUSTER_ID).collect(Collectors.toSet());
```
t-8ch: `Collections.singleton()`

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeCoordinatorHandler.java
```diff
@@ -439,6 +445,13 @@ private synchronized void initialiseZigBee() {
                 return;
         }
 
+        // Add all the clusters that we are supporting.
+        // If we don't do this, the framework will reject any packets for clusters we have not stated support for.
+        channelFactory.getImplementedClientClusters().stream()
```
t-8ch: According to the OSGi references in the handler factories the channelFactory can be `null`.
This would throw a NPE here.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/handler/ZigBeeCoordinatorHandler.java
```diff
@@ -439,6 +445,13 @@ private synchronized void initialiseZigBee() {
...
+                .forEach(clusterId -> networkManager.addSupportedClientCluster(clusterId));
+        channelFactory.getImplementedServerClusters().stream()
```
t-8ch: Could be reduced to `channelFactory.getImplementedServerClusters().forEach(networkManager::addSupportedServerCluster)`

cdjackson:
Thanks for the review @t-8ch

cdjackson:
Closes #531

cdjackson:
This now updates to the ZSmart System libraries ```1.2.8```.

https://github.com/openhab/openhab-distro/pull/1043

Issue #506, opened 2019-10-12
[Telegesis] Baud rate cannot be configured


petero-dk:
I have a Telegesis usb device where I have had the Baud rate on the device configured to a non-standard 115200 and not the standard 19200. When I change the baud rate on the device in openHAB the settings wont stick. It cannot connect and when I open the device settings again the baudrate is 19200 again. (When I reconfigure my device to 19200 baudrate via the Telegesis Terminal utility I can get it to work in openHAB

When I look at the source code it seems the device is hard configured with the default 19200. 

https://github.com/openhab/org.openhab.binding.zigbee/blob/c768e658b282eea32651b246ecf0856e4daf0f4d/org.openhab.binding.zigbee.telegesis/src/main/java/org/openhab/binding/zigbee/telegesis/internal/discovery/ZigBeeTelegesisUsbSerialDiscoveryParticipant.java#L76-L77

I am correct in that this is the place where the device is initiated with the default baud rate?

cdjackson:
The code that you reference is not related to opening the device. This is only related to automatic discovery of dongles.

The configuration of the Telegesis dongle uses the baud rate that you configure in the UI.

https://github.com/openhab/org.openhab.binding.zigbee/blob/c768e658b282eea32651b246ecf0856e4daf0f4d/org.openhab.binding.zigbee.telegesis/src/main/java/org/openhab/binding/zigbee/telegesis/handler/TelegesisHandler.java#L68

cdjackson:
@petero-dk shall we close this or is there further information available?

petero-dk:
Give me one week then I will do a little more testing, ok?

petero-dk:
Ok, i finally got to test a little more. 

(The starting point is a fully working openHAB with a standard telegesis usb stick.)

First I checked the baud rate on the usb device via the telegesis terminal:
```
Command: ATS12?
Response: 0510
```
This means 19200 BAUD and no command response.

Then I set the new BAUD rate
```
Command: ATS12=0C10
Response (closed connection)
```
This sounds good, reconnecting to the device with 115200 connects and works now ATS12? gives 0C10.

Before closing openhab I set the BAUD rate of the device to 115200 then shutdown openHAB, upon restarting the device I only get the following error in the zigbee log (will logging down to info)
```

[...]

PR #524, opened 2019-11-30
Removed baud configuration on telegesis dongle discovery 

petero-dk:
So the thing discovery will not overwrite configured items, see #506


cdjackson:
Thanks @petero-dk . Unfortunately this causes a test failure. Can you take a look in the ```ZigBeeTelegesisUsbSerialDiscoveryParticipantTest``` file and you'll find a few tests that check that the baud rate is not null - these lines should now be removed.

cdjackson:
thanks @petero-dk

PR #537, opened 2019-12-29
Update README.md

marfrh:
Add firmware requirement R309 for Telegesis Dongle


cdjackson:
Thanks. The firmware should also be displayed in the properties, and once I have a development environment working again I'll look at why this is.

Also, with the latest changes, the dongle will not start at all and an error will be logged if the firmware is not R309.

PR #541, opened 2019-12-30
Update POM parent version

cdjackson:
@kaikreuzer please can you check this - I assume this wasn't left at ```2.5.0``` for a reason?

I'm currently struggling to get the ZigBee binding working after the 2.5 release - the dependencies seem to be messed up in the IDE (it's another bnd related issue I think).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
   <parent>
     <groupId>org.openhab.addons</groupId>
     <artifactId>org.openhab.addons.reactor</artifactId>
-    <version>2.5.0</version>
+    <version>2.5.1-SNAPSHOT</version>
     <relativePath/>
   </parent>
 


cdjackson:
I realised that there are also other dependencies linked to ```2.5.0``` in this pom. As with https://github.com/openhab/openhab-distro/pull/1035 though, changing them to ```2.5.1-SNAPSHOT``` introduces dependency errors.

kaikreuzer:
>  I assume this wasn't left at 2.5.0 for a reason?

It was - all 2.5.x add-ons should be compiled against the released 2.5.0 core, since there is no 2.5.x branch for the core artifacts.

kaikreuzer:
You can see at https://ci.openhab.org/view/Integration%20Builds%20(2.5.x)/job/openHAB2.5.x-Zigbee/ that it builds nicely.

cdjackson:
The problem I have is I can’t change either the dependencies (ie the ZSmart Systems libs) or the binding in the IDE. If I run the binding in the IDE, it seems to run the version from pre 2.5. So, yes, it might build just fine here, but the IDE is really a mess at the moment and I’m unable to get a working environment.

https://community.openhab.org/t/tutorial-migrate-your-binding-to-the-maven-bnd-based-build-system/81389/45 <https://community.openhab.org/t/tutorial-migrate-your-binding-to-the-maven-bnd-based-build-system/81389/45>

pom.xml
```diff
@@ -4,7 +4,7 @@
   <parent>
     <groupId>org.openhab.addons</groupId>
     <artifactId>org.openhab.addons.reactor</artifactId>
-    <version>2.5.0</version>
+    <version>2.5.1-SNAPSHOT</version>
```
kaikreuzer: This one might be ok to change to 2.5.1-SNAPSHOT as it is about the addons.
So let's have a try - but please remove the other changes.

cdjackson:
Thanks. I reverted the second commit which had the other dependency changes. I don’t think this really helps with the problem I have though unfortunately.

PR #556, opened 2020-02-07
Create feature file in ZigBee repository

cdjackson:


cdjackson:
@kaikreuzer this (mostly!) creates the feature file from the ZigBee dependencies. The only thing I've not managed to work out is how to add the OH feature dependencies - eg the serial transport. Do you happen to know how this is done?

I have another PR ready for ```openhab-distro``` to use this feature file. If you're happy with this I'll create that PR as well.

Partly fixes https://github.com/openhab/openhab-distro/issues/1060

cdjackson:
@kaikreuzer I've now got this working with the extra feature dependencies. Is there any reason that this can't be merged? It means that the feature file is available for use, but doesn't cause any problems if other projects decide not to consume it - right?

Thanks.

PR #559, opened 2020-02-24
Update ZSmart Systems libraries to 1.3.1

giannello:
https://github.com/zsmartsystems/com.zsmartsystems.zigbee/issues/986

Fixes serial buffer overruns

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -32,7 +32,7 @@
 
   <properties>
     <report.fail.on.error>false</report.fail.on.error>
-    <zsmartsystems.version>1.3.0</zsmartsystems.version>
+    <zsmartsystems.version>1.3.1</zsmartsystems.version>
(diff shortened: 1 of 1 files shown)


cdjackson:
Unfortunately this doesn’t update the libraries. I am currently holding off doing an update as there may be some other changes coming through. You can use the script by 5iver to update the ZSS libraries.

cdjackson:
Unfortunately this doesn’t update the libraries. I am currently holding off doing an update as there may be some other changes coming through. You can use the script by 5iver to update the ZSS libraries.

giannello:
I had no idea about the script from 5iver, ended up compiling the binding with this patch and manually copying the ZSS jars from my maven local repository to the addons directory :)

Closing the PR then, waiting for the other round of updates :)

PR #571, opened 2020-04-29
Remove unnecessary discoveries

giannello:
Reduce the traffic on the network, especially when a zigbee `scan` is initiated.

This fixes the issue of new nodes failing to join an existing network (~5 ROUTERs and 15 END_DEVICEs) due to excessive traffic.

### DIFF
--- a/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/discovery/ZigBeeDiscoveryService.java
+++ b/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/discovery/ZigBeeDiscoveryService.java
@@ -133,7 +133,7 @@ public class ZigBeeDiscoveryService extends AbstractDiscoveryService {
 
         for (ZigBeeCoordinatorHandler coordinator : coordinatorHandlers) {
             for (ZigBeeNode node : coordinator.getNodes()) {
-                if (node.getNetworkAddress() == 0) {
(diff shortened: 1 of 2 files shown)


cdjackson:
@giannello sorry for taking so long with this. I don't think this is the right approach. The `node.isDiscovered()` method is looking at the ZigBee network information and not the binding. The problem is that the `DiscoveryService` has no way to know if a device is already added as a Thing, so it adds all nodes that are known on the network in case they are not included in openHAB as a thing.

There used to be an "enhanced" abstract discovery service that (IIRC) provided the ability to look at discovered Things in OH, but this was removed.

So, we can do two things - what we are doing now where we add all devices to the network and we ensure nothing is missed - or - we remove this altogether. The risk there is that we could miss adding things.

There is a specific case here that will not work with your change. Currently OH has no way to update thing definitions - other than to delete the thing and add it back again. This is common on ZWave especially where the thing definitions are updated often, but it's the same issue here. To update the thing, we need to delete the thing, and add it back. If we don't add all things during the discovery, then I think this will no longer be possible, and instead users will have to remove the devices from the network and add them back for them to be discovered.

I'll have a think if there is a better way for the discovery service to know what devices are already Things.

cdjackson:
Also, looking at the change you made in the Thing Handler to not call rediscoverNode, again this is incorrect and may lead the devices getting lost.

I will close this PR and have opened #583 to capture any discussion.

giannello:
Fair enough :)

Thanks for taking your time to investigate.

PR #600, opened 2020-08-23
Only perform discovery on existing devices

### LINKED ISSUE #583: Only perform discovery on existing devices
Currently the binding performs discovery on all devices - even those that are already `Things`. This is because the discovery service has no way to know if a device is already discovered and added as a `Thing`.

Since devices may (re)join the network at any time we either need to add devices when discovery is started, or add support for the OH background discovery concept.

We also need to handle one other case which is if Things are deleted, but not removed from the network. In this case, if we don't add all known things during discovery, we will loose devices completely.

falmanna:
Don't perform discovery on discovered things
Closes #583

Signed-off-by: Feras Almanna <firas.almanna3@gmail.com>


cdjackson:
Thanks @falmanna LGTM.

Issue #601, opened 2020-08-24
Xiaomi door window sensor not working - on off channel consolidation issue


mattibal:
Hi all,

I have a Xiaomi door window sensor MCCGQ01LM (the rounded non-Aqara ones) that's not working with the Zigbee binding (with both the stable 2.5.7 and snapshot version).
The thing is correctly recognized and a "Level Control" channel shows up, but when you link an item to that channel, it doesn't get any event and doesn't change status when you open or close the door.

I looked at what's happening with a debugger and I think there is an issue in the Zigbee binding code.
The sensor seems to support On Off and Level Control as client clusters, but only On Off as server cluster. The Zigbee binding creates anyway just a Level Control channel (and not also an On Off channel) due to "consolidation", without taking into account that the level control cluster is supported by the sensor only as client and not as server.
So a ```ZigBeeConverterSwitchLevel``` is getting created, and it does initializes the device as "client" instead of as "server" because it gets confused by the fact that the level control server cluster is missing, so it doesn't register an AttributeListener for the On Off cluster and doesn't get On Off notifications.

As a workaround, I tried to disable the On Off - Level Control channels "consolidation" by commenting this line: https://github.com/openhab/org.openhab.binding.zigbee/blob/db37991a066c428835b6851c9a1da348bad41d24/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeChannelConverterFactoryImpl.java#L65-L66
and it seems to work fine. 
A "Switch" channel gets created alongside to the "Level Control" one, and when linking an item to the Switch channel, it does receive events and changes status normally when the door is open or closed.

I know that's not a proper fix for the issue, but might it be a good enough temporary solution? 
I've tested it with my IKEA Tradfri lights and an Innr SP120 plug, and it doesn't seems to break anything. 

BTW thank you for your work on the binding!

cdjackson:
> The sensor seems to support On Off and Level Control as client clusters, but only On Off as server cluster. 

That's incorrect. Both client and server are supported - they have been for a long time as we support wall switches that send OnOff (ie client) and lights that receive OnOff (ie server).

I don't know why a window sensor is a level control device anyway - it should be an IAS device - level doesn't really make sense. However, that's an asside.

I won't disable the channel consolidation if that's what you are suggesting - this goes against the OH design philosophy which is why it was added in the first place.  You are better off considering a static device definition.

mattibal:
[...]

PR #602, opened 2020-08-25
Add lumi.sensor_magnet static thing definition

mattibal:
This is a workaround for the #601 issue that was causing the Zigbee binding don't get any attribute update report from the On Off server cluster of that device.

I've tested this with my MCCGQ01LM Xiaomi Door Window Sensor, and it's working fine.

I'm not sure if the channel name is acceptable for Openhab standards, but I haven't found a better name :) 
It's an hacky solution anyway, since the magnetic contact is seen as a Switch instead of a Contact. If from Openhab you try to toggle that switch, the operation will obviously fail.

### DIFF
--- a/org.openhab.binding.zigbee/src/main/resources/ESH-INF/thing/xiaomi/lumisensor-magnet.xml
+++ b/org.openhab.binding.zigbee/src/main/resources/ESH-INF/thing/xiaomi/lumisensor-magnet.xml
@@ -1,0 +1,37 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<thing:thing-descriptions bindingId="zigbee"
+                          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                          xmlns:thing="https://openhab.org/schemas/thing-description/v1.0.0"
+                          xsi:schemaLocation="https://openhab.org/schemas/thing-description/v1.0.0 https://openhab.org/schemas/thing-description-1.0.0.xsd">
+
+    <!-- This is needed to avoid the device being considered just a Level Control client -->
+
+    <thing-type id="xiaomi_lumisensormagnet" listed="false">
+        <label>Xiaomi Door Window Sensor</label>
+
+        <channels>
+            <channel id="contact_switch" typeId="switch_onoff">
+                <label>Contact open (Switch)</label>
+                <description>The magnetic contact, seen by the device as a switch that is on when the contact is open</description>
+                <properties>
+                    <property name="zigbee_endpoint">1</property>
+                    <property name="zigbee_inputclusters">6</property>
+                </properties>
+            </channel>
+        </channels>
+
+        <properties>
+            <property name="vendor">Xiaomi</property>
+            <property name="modelId">lumi.sensor_magnet</property>
+            <property name="zigbee_logicaltype">END_DEVICE</property>
+        </properties>
+
+        <representation-property>zigbee_macaddress</representation-property>
+
+        <config-description>
+            <parameter name="zigbee_macaddress" type="text" readOnly="true" required="true">
+                <label>MAC Address</label>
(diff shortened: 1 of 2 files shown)


cdjackson:
Thanks - this looks fine to me. I guess the other option would be to use a trigger channel?

Closes #601

mattibal:
That would be nice, but if I understood correctly the ```ZigBeeConverterSwitchOnoff``` code, it's currently not possible to create a trigger channel that gets triggered when a On or Off command is received.

I'm not a Zigbee expert, but I think it would be useful anyway to have a switch channel (or better a contact channel) alongside the trigger channel, to always have the contact status information updated periodically by attribute reports even if the On or Off command message got lost for some reason (RF interference, routing issues?) when the door/window has been opened or closed.

cdjackson:
It’s not related to the on off converter or the on off channel at all. You can instead try to define a trigger channel that can be configured (I think) to respond to any command. Take a look at some of the remote control thing definitions in the binding.

mattibal:
Ok, I think I've found how to do what you were saying, I tried to add this to the thing definition:
```xml
<channel id="contact_opened" typeId="system.button">
    <label>Contact opened</label>
    <description>Contact opened event</description>
    <properties>
        <property name="zigbee_endpoint">1</property>
        <property name="zigbee_shortpress_cluster_id">0x0006</property>
        <property name="zigbee_shortpress_command_id">0x01</property>
    </properties>
</channel>

<channel id="contact_closed" typeId="system.button">
    <label>Contact closed</label>
    <description>Contact closed event</description>
    <properties>
        <property name="zigbee_endpoint">1</property>
        <property name="zigbee_shortpress_cluster_id">0x0006</property>
        <property name="zigbee_shortpress_command_id">0x40</property>
    </properties>
</channel>
```
Openhab correctly tries to bind to the client cluster 6 (On Off) of the device, but unfortunately the bind fails:
```
2020-08-25 23:52:52.607 [ERROR] [nverter.ZigBeeConverterGenericButton] - 00158D000460E3C9: Error 0xffff setting client binding for cluster 6
```
by looking with a debugger, that error seems to be a timeout while waiting for the bind response. 
I though that maybe that's somewhat related to the fact that the Xiaomi devices "sleep too much", so I tried to press the pairing button once a second while Openhab sends the bind message, but I got the same error.
As a consequence, I don't receive any command when the contact is opened or closed.

katiuskt:
Sorry guys, I don't fully understand the whole situation here but in OpenHAB 3 (on a rPi with OpenHABian) I have this very same issue:
CC2531 USB Controller + MCCGQ01LM door sensor.
Sensor thing properties shows it is recognized as a lumi.sensor_magnet but openhab only creates the dimmer channel. Not the on/off one.

All exactly as it was reported.

Into logs (IP:9001) I can see these errors:

> 2021-01-06 08:21:57.312 [ERROR] [converter.ZigBeeConverterSwitchLevel] - 00158D00063312E4: Error 0xffff setting client binding
> 2021-01-06 08:22:43.253 [ERROR] [converter.ZigBeeConverterSwitchLevel] - 00158D00063312E4: Error 0xffff setting client binding


Sensor itself is working (tested on Home Assistant + ZHA and with Zigbee2MQTT). They also work using a ConBee USB stick + DeConz binding. So it seems sensor only fails with OpenHAB Zigbee binding but I can't tell much more than that due to my knowledge :-(

I see this bug is open but for OpenHAB 2.5.x branch... Is there any chance to get sensor working on OpenHAB 3.x? I don't know if there is any change under the hood which could help...

cdjackson:
The error you mention is reported when the device reports an error, or possibly when the bind command times out. It could be caused by the device being asleep.

I'd suggest to discuss your issue on the support forum and not in this PR. Full logs would also be needed, but please provide and discuss on the forum.

PR #612, opened 2020-10-03
Apply and enable Spotless

wborn:
This will make and keep the code spotless!


cdjackson:
This completely destroys the history - is this really necessary??????? Personally I'd prefer to avoid this as I don't want to rewrite history in this way.

wborn:
The history is still there, you can still use blame/praise and got to the next diff.

PR #619, opened 2020-12-21
Update README.md

Marty-O:
Bitron Video Stick needs 57600baud rate and Software flowcontrol to work on Openhab (2.5.x).


org.openhab.binding.zigbee/README.md
```diff
@@ -137,9 +137,9 @@ The following coordinators are known to be supported.
...
-|[Bitron Video ZigBee USB Funkstick](http://www.bitronvideo.eu/index.php/produkte/smart-home-produkte/zb-funkstick/)|[Ember](#ember-ezsp-ncp-coordinator)|115200bps<br>Hardware flow control<br>High RAM||
+|[Bitron Video ZigBee USB Funkstick](http://www.bitronvideo.eu/index.php/produkte/smart-home-produkte/zb-funkstick/)|[Ember](#ember-ezsp-ncp-coordinator)|57600bps<br>Aoftware flow control<br>High RAM||
```
cdjackson, line 140: Software is spelt wrong.

org.openhab.binding.zigbee/README.md
```diff
@@ -137,9 +137,9 @@ The following coordinators are known to be supported.
...
 |[Elelabs ELU013/ELR023](https://elelabs.com/shop/)|[Ember](#ember-ezsp-ncp-coordinator)|115200bps<br>Hardware flow control<br>High RAM| Both the stick and the hat can be upgraded without additional hardware, firmware available [here](https://github.com/Elelabs/elelabs-zigbee-ezsp-utility).|
-|[Cortet EM358 USB Stick](https://www.cortet.com/iot-hardware/cortet-usb-sticks/em358-usb-stick)|[Ember](#ember-ezsp-ncp-coordinator)|57600bps<br>Software flow control<br>High RAM| |
+|[Cortet EM358 USB Stick](https://www.cortet.com/iot-hardware/cortet-usb-sticks/em358-usb-stick)|[Ember](#ember-ezsp-ncp-coordinator)|57600bps<br>Software flowcontrol<br>High RAM| |
```
cdjackson, line 142: For consistency, please can we use "flow control" everywhere rather than "flowcontrol" here. It might also pay to say that hardware flow control and 115k2 is also used - it depends on the software and since these devices don't come with software loaded, it's a little hard to define that here.

cdjackson (changes requested):
Thanks - just a couple of small comments.

org.openhab.binding.zigbee/README.md
Marty-O: Agreed. If the 56700 baud rate is verified to work for multiple users and the 115k2 isn't. Couldn't we assume that in regards to OH 56700 is the correct setting to use and should be the (only) one mentioned in the binding docs?

cdjackson: I don't think so, no. There are many versions of software for the EM358 that uses hardware flow control and 115k2. Silabs provide firmware for both, so we can't assume that someone will only use the software flow control - especially this device comes without any firmware loaded at all. Both should be mentioned here I think to avoid further confusion.

Marty-O: Ok. How about not changing the baud date / flow control, keeping it at 115k2 / hardware, but using the comment section to indicate the device is verified to work with OH2.5 using  @ 57600 baud / Software? Or am i missing some basic's here?

cdjackson: It will not work with 57600 and software control if the firmware set to 115k2 and hardware. Likewise, it will not work with 115k2 and hardware control if the firmware is set to 57k6. There are two firmware versions possible - this device comes with no firmware loaded, so there both options could be loaded - depending on the user.
So, we can't say that it is 115k2, but verified to work at 57k6 - that is incorrect. We can only say that there are two firmware versions, and if users are unsure, they should try both.

cdjackson:
Sorry - I hadn't spotted that you had updated the commit.

Thanks.

PR #622, opened 2020-12-27
Fix a small formatting mistake in README.md

LinqLover:
The reason was the use of a \&nbsp; instead of a regular space character.

I have checked the repository for other incidents, but there were no others.


kaikreuzer (approved):
Thank you!

PR #624, opened 2021-01-01
Fix build

wborn:
The code was using some deprecated constructors which were hidden by openhab/openhab-core#1995 to enforce builders being used instead.


cdjackson:
Thanks @wborn

PR #627, opened 2021-01-03
Extended and clarified documentation for ZigBee Binding

bartkummel:
- I added some more introductory text to place the ZigBee binding in context.
- Corrected inline monospaced text to use only one backtick.
- Changed "Paper UI" into "UI" where appropriate.


bartkummel:
Sorry, something went terribly wrong when I forgot to signoff my PR and tried to correct it with a rebase. I'm running out of time no. Perhaps I'll try to make a new PR later.

PR #633, opened 2021-01-07
Fix XStream security framework not initialized warnings

wborn:
See also this [community thread](https://community.openhab.org/t/xstream-not-initalized/110879?u=wborn)


cdjackson:
Thanks @wborn

PR #661, opened 2021-07-08
Fix RFC2217 for Zigbee

dschall:
This change skips enumerating all available COM ports when rfc2217 is used.
It also adds a call to flush() to both write() methods to make remote connections work.
Finally, it refactors some of the exception handling to take into account that some calls may fail without bringing down the whole connection.


dschall:
@cdjackson , not sure if you'll see new PRs automatically or if you need a ping..

cdjackson:
Is it not possible to implement this in the serial implementation rather than the binding. That would also then (presumably) resolve it for other bindings as well without having to make this change everywhere?

dschall:
I don't think it's possible to fix this in the serial implementation.
We'll have to skip the call to `serialPortManager.getIdentifiers()` here, because RCF2217 ports can't be discovered, so we'll always get an empty list back if no other serial ports are available on the system (or they are in use already).

Plus, we'll have to make at least one change in this binding:
RFC2217 won't work without the call to `flush` after writing.

giuliano-favro:
Shall be also possible to use 'socket://' instead 'rfc2217'?

PR #670, opened 2021-08-30
Added firmware upgrade instructions for some Ember

rkoshak:
Added instruction for upgrading the firmware for HUSBZB-1 and Telegesis ETRX357USB and perhaps other Ember based controllers using https://github.com/walthowd/husbzb-firmware. I tested these instructions on a HUSBZB-1.

As far as I can tell this tool is Linux only and, even if it did support Windows I don't have a Windows machine to test it on so would not be comfortable writing that section.

### DIFF
--- a/org.openhab.binding.zigbee/README.md
+++ b/org.openhab.binding.zigbee/README.md
@@ -201,7 +201,9 @@ If the usb dongle is not recognized, it might be necessary to make the dongle's 
 ##### Flashing/Upgrading on Linux
 
 Some Ember EZSP NCP coordinators can be flashed to upgrade the firmware using walthowd/husbzb-firmware (https://github.com/walthowd/husbzb-firmware).
-It has been tested with the HUSZBZ-1 and Telegesis ETRX357USB and may work with other coordinators based on similar EM358x/EM3581, ETRX35x/ETRX357, and EFR32MGxx MCU chips from Silicon Labs.
+It has been tested with the HUSZBZ-1 and Telegesis ETRX357USB and may work with other coordinators. 
+You are responsible to ensure you load the correct firmware for your device. 
+Loading the wrong firmware may render your device inoperable
 
 Review the `husbzb-firmware` readme for detailed instructions.
 Using docker to upgrade a HUSZBZ-1 is as follows:
--- a/org.openhab.binding.zigbee/README.md
+++ b/org.openhab.binding.zigbee/README.md
@@ -198,6 +198,28 @@ If the usb dongle is not recognized, it might be necessary to make the dongle's 
   ```
 - Plug in the dongle. It should now be recognized properly as ttyUSBx.
 
+##### Flashing/Upgrading on Linux
+
+Some Ember EZSP NCP coordinators can be flashed to upgrade the firmware using walthowd/husbzb-firmware (https://github.com/walthowd/husbzb-firmware).
+It has been tested with the HUSZBZ-1 and Telegesis ETRX357USB and may work with other coordinators based on similar EM358x/EM3581, ETRX35x/ETRX357, and EFR32MGxx MCU chips from Silicon Labs.
+
+Review the `husbzb-firmware` readme for detailed instructions.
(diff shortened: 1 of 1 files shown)


rkoshak:
Shoot, I forgot the sign-off and I can't see how to amend the commit message after the fact. I did the change through the browser and not via a local clone.

org.openhab.binding.zigbee/README.md
```diff
@@ -198,6 +198,28 @@ If the usb dongle is not recognized, it might be necessary to make the dongle's
   ```
 - Plug in the dongle. It should now be recognized properly as ttyUSBx.
 
+##### Flashing/Upgrading on Linux
+
+Some Ember EZSP NCP coordinators can be flashed to upgrade the firmware using walthowd/husbzb-firmware (https://github.com/walthowd/husbzb-firmware).
+It has been tested with the HUSZBZ-1 and Telegesis ETRX357USB and may work with other coordinators based on similar EM358x/EM3581, ETRX35x/ETRX357, and EFR32MGxx MCU chips from Silicon Labs.
```
cdjackson, line 204: Are you REALLY sure about this???

The EFR32 and EM357/358 are COMPLETELY different. Their firmware is not compatible at all so I'm surprised that this states that it updates all types of Silabs devices. I understood from the discussion on the forum that this was for the HUSBZB1 which is an EM357 - if so it is not going to work for other devices.

rkoshak: I'm just repeating what was stated on the tool's github readme. That's actually a direct copy and paste from the tool. It's no biggie to me to remove it. I figured the developer of the tool knew what they were talking about.

cdjackson: That sounds dodgy. With the EM357 and EM358 it's reasonably safe, but the EFR32 devices have a lot more configuration options, and if you program the wrong code (ie code that is configured to use a different oscillator, or different memory, or different ports) then you will brick the device. I've recently done exactly this when a customer changed their hardware configuration in an EFR32 and you then need to have the programmer to reprogram the device at low level.

I just don't want to be responsible for bricking peoples devices so maybe we should add a "use at own risk" disclaimer to this.

The other option of course is always to use the binding, although the user still needs to have the correct firmware in order to flash the device.

rkoshak: Always use the binding only works if the binding supports it. My understanding is that's not the case for the HUSZBZ-1.

How about I end the sentence at "may work with other coordinators. Use at your own risk."

cdjackson: > Always use the binding only works if the binding supports it. My understanding is that's not the case for the HUSZBZ-1.

Yes, that was also my understanding, but as I don't have one of these devices I've never been able to test it. However clearly this has a bootloader right, or it would be impossible to bootload using this other software. And presumably it is using the standard Silabs bootloader as it claims to be compatible with every Silabs chip. In this case, why wouldn't it also work with the binding, or the OH console at least? I added bootloader support originally for Eclipse SmartHome when German Telecom were directly using the binding and they needed it to support their customers.

> How about I end the sentence at "may work with other coordinators. Use at your own risk."

I would state something like "You are responsible to ensure you load the correct firmware for your device. Loading the wrong firmware may render your device inoperable". This makes it a bit clearer (IMHO) that the consequences of getting this wrong are a bricked stick that most people will not be able to recover from.

cdjackson:
I'll merge this an them make further updates. Thanks @rkoshak.

Hedda:
FYI, for EFR32 you can also refer to Elelabs EZSP Firmware Update Utility -> https://github.com/Elelabs/elelabs-zigbee-ezsp-utility/

I have however successfully tested both walthowd + Elelabs tool Itead's dongle -> https://itead.cc/product/zigbee-3-0-usb-dongle/

cdjackson:
As mentioned above, this can be done directly in the binding - this was added a long time ago (For Germany Telekom), but not documented. I will be updating this shortly.

PR #676, opened 2021-10-02
Upgrade Z-Smart Systems library to 1.4.1

cdjackson:
This brings the following major changes -:

* [Improve error handling during Ember network initialisation](https://github.com/zsmartsystems/com.zsmartsystems.zigbee/pull/1266)
* [XBee: Handle interrupted/timed out requests when getting channel.](https://github.com/zsmartsystems/com.zsmartsystems.zigbee/pull/1263)
* [Improve TelegesisFirmwareUpdateHandler to follow the XModen-CRC protocol](https://github.com/zsmartsystems/com.zsmartsystems.zigbee/pull/1262)
* [Move firmware command to dedicated command handler](https://github.com/zsmartsystems/com.zsmartsystems.zigbee/pull/1258)
* [Fix EmberFirmwareUpdateHandler to follow the XModem-CRC protocol and also fix retry logic](https://github.com/zsmartsystems/com.zsmartsystems.zigbee/pull/1255)
* [Update Ember ncphandler command to support reading statistics counters](https://github.com/zsmartsystems/com.zsmartsystems.zigbee/pull/1249)
* [Allow hex addresses in node ID in console commands](https://github.com/zsmartsystems/com.zsmartsystems.zigbee/pull/1247)
* [Improve channel selection command for Ember](https://github.com/zsmartsystems/com.zsmartsystems.zigbee/pull/1246)

@kaikreuzer or @wborn I don't think that anything further is now required in the distro as I think you are pulling in dependencies from the feature - right? So merging this will automatically make the updated dependencies available?

If so, please feel free to merge.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -33,7 +33,7 @@
 
   <properties>
     <report.fail.on.error>false</report.fail.on.error>
-    <zsmartsystems.version>1.4.0</zsmartsystems.version>
+    <zsmartsystems.version>1.4.1</zsmartsystems.version>
     <spotless.version>2.0.3</spotless.version>
     <spotless.check.skip>true</spotless.check.skip> <!-- Spotless disabled for now -->
     <sat.version>0.10.0</sat.version>


wborn:
> I think you are pulling in dependencies from the feature - right?

Yes that was done with https://github.com/openhab/openhab-distro/pull/1158.

> So merging this will automatically make the updated dependencies available?

It will probably build fine with Jenkins. But I wonder if these libs also need to be uploaded to the openHAB JFrog Maven repo. :thinking:
This was for instance recently an issue with the system info binding libraries (https://github.com/openhab/openhab-addons/issues/11294). If the libraries are embedded into the bundles it should be no issue.

cdjackson:
Thanks. The libs are not embedded - they need to be pulled in with maven. I assumed this was best rather than embedding the JAR within the bundle.

The libraries are on central - why do they need to be uploaded to an OH repo?

wborn:
The distro currently only uses the [openHAB online repo](https://github.com/openhab/openhab-distro/blob/34c684842beec47b1e51febd5ffb670f537504e7/distributions/openhab/src/main/filtered-resources/userdata/etc/org.ops4j.pax.url.mvn.cfg#L90-L91) for downloading online artifacts and not Maven Central. IIRC on Bintray the openHAB repos would also proxy Maven Central artifacts, but there was not enough available storage with the free JFrog plan.  It's certainly an annoying issue. :slightly_frowning_face:

cdjackson:
So how best do we proceed so we can get this merged? I'm not super keen to embed the JARs as it removes the flexibility to debug using different libraries. It seems (to me at least) like the best option is for OH to pull directly from central, but I guess that's not a quick option (if at all).

Shall I merge this and then sort out any library updates required in the distro if needed?

wborn:
> the best option is for OH to pull directly from central

If it is not possible to proxy artifacts from Maven Central on the openHAB JFrog repo anytime soon, it might also be an option to add Maven Central release artifacts to the `org.ops4j.pax.url.mvn.repositories` list. WDYT @kaikreuzer?

> Shall I merge this and then sort out any library updates required in the distro if needed?

I would go with that. There was a recent 3.2.0.M3 that people can revert to if they have issues. Another way to workaround these download issues is by using the add-on KAR file.

cdjackson:
> Without having all the compile/test dependencies of Maven Central in our own repo with a proxy, we won't be able to build openHAB when Central goes down one day.


Sure, but how likely is that? Half of the Java world uses central, so openHAB won't be alone. It's obviously up to you guys, but it just seems to me that this unlikely scenario that "one day" central might fail for a few hours, is not worth the hassle it causes and maintaining things could be a lot simpler (IMHO).

wborn:
Yes the openHAB repo (Bintray) has more availability issues than Maven Central. :wink: 

If we truly want to stay independent, we should also make sure our build servers don't download anything from Maven Central. So we can spot that we are and stay independent and everything is in our repos. Maybe the openHAB Foundation can help paying for a  [plan](https://jfrog.com/pricing/) that allows for proxying Maven Central?

cdjackson:
> I think it makes a big difference if we speak about the build or the product.

I agree, but it seems to me that you are doing the job of the build manually - aren't you? You are manually copying the dependencies into your own repository when you should be able to do this as part of the build process when generating the product. That way we should be able to define dependencies in the bindings and not have to manually upload them to the repo?

kaikreuzer:
In the past we used the KAR file and exploded that as a remote repo - it contains everything that is needed, but nothing more. If we could find a similar mechanism, this would imho be ideal.

PR #699, opened 2021-11-13
Build KAR to simplify dependency management

cdjackson:
@wborn you may have noticed I up-reved the ZSmart Systems libraries a day or so ago. As suggested previously you wanted a KAR to simplify the dependency management, so this PR now creates the KAR which includes these dependencies.

### DIFF
--- a/feature/pom.xml
+++ b/feature/pom.xml
@@ -1,11 +1,13 @@
-<?xml version="1.0" encoding="UTF-8" standalone="no"?><project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
 
 	<modelVersion>4.0.0</modelVersion>
 
 	<parent>
 		<groupId>org.openhab.addons.bundles</groupId>
 		<artifactId>org.openhab.addons.zigbee.reactor</artifactId>
-    	<version>3.2.0-SNAPSHOT</version>
+		<version>3.2.0-SNAPSHOT</version>
 	</parent>
 
 	<groupId>org.openhab.addons.features.karaf</groupId>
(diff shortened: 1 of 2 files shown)


wborn:
How does another KAR exactly simplify dependency management?
All zigbee dependencies are already part of openhab-addons-3.x.y.kar.
It's the online repo where these dependencies currently need to be manually added, see https://github.com/openhab/org.openhab.binding.zigbee/pull/676.

cdjackson:
> How does another KAR exactly simplify dependency management?

Ok, sorry - I thought that the comment from Kai was requesting that the KAR be added here so that the dependencies were all in one place so that you didn't need to manage them separately upstream. 

In the link you referenced, Kai said -:

> In the past we used the KAR file and exploded that as a remote repo - it contains everything that is needed, but nothing more. If we could find a similar mechanism, this would imho be ideal.

As I understand at the moment, you, or Kai, need to manually download the JAR from central and upload them to the OH repo. I understood the comment as being  you wanted me to create the KAR with the dependencies included, so that they could be exploded in the addons processing?

wborn:
> I understood the comment as being you wanted me to create the KAR with the dependencies included, so that they could be exploded in the addons processing?

I think @kaikreuzer meant to use the current add-ons KAR, and either automatically add all the content to Artifactory using some script... or maybe it can even be added to any CDN as static content. I did notice there are no POM files in a KAR, but if it worked previously like that, that may not be an issue for Karaf. :slightly_smiling_face:

cdjackson:
Ok, I've not looked at where the various OH repositories pull dependencies from and I understood from the previous discussion that you were pulling everything from an OH specific repo even at build time. Therefore I thought that providing the dependencies here meant that you didn't need to manage this elsewhere.

I'll leave it to you guys to manage then - I'm just trying to make it easier so that the zigbee binding can be built with the minimum of fuss.

kaikreuzer:
> I think @kaikreuzer meant to use the current add-ons KAR, and either automatically add all the content to Artifactory using some script... 

That's correct.

> I did notice there are no POM files in a KAR, but if it worked previously like that, that may not be an issue for Karaf.

I'm not 100% certain here. It might be that Bintray used to create dummy pom.xmls when uploading artifacts and hence Karaf was happy with it. But Artifactory doesn't do this and therefore the artifacts are not resolved.

>  I'm just trying to make it easier 

Thanks for trying. I'll upload the libs manually again for now.

kaikreuzer:
Btw @cdjackson, is it correct that there is no version 1.4.2 of the console bundle for Ember, see https://mvnrepository.com/artifact/com.zsmartsystems.zigbee/com.zsmartsystems.zigbee.console.ember?

cdjackson:
Strange - no - everything should have been built together, so this should also be there. It’s strange that the binding builds if this is missing (both locally and on Jenkins). 

It does seem to be there if I link directly though -:

https://mvnrepository.com/artifact/com.zsmartsystems.zigbee/com.zsmartsystems.zigbee.console.ember/1.4.2

https://repo1.maven.org/maven2/com/zsmartsystems/zigbee/com.zsmartsystems.zigbee.console.ember/1.4.2/

So it seems to just be an issue with this web index and the artefacts are really there.

PR #706, opened 2021-11-18
Add support for round Xiaomi Mijia wireless switch.

elfman03:
The Xiaomi Mijia wireless switch is a round button.  Like a lot of the
other LUMI devices, it does not report its clusters correctly and needs
to be defined via xml to register correctly.  This patterns an xml file
based on the other similar Xiaomi devices.

I verified that it works with a Xiaomi Mijia button that I have.

Signed-off-by: Chris Elford <elfman03@gmail.com> (github: elfman03)


elfman03:
for some reason signoff failed.  will try agin.

PR #712, opened 2021-11-26
Fix build

wborn:
The build fails because the code uses deprecated APIs which were removed in Mockito 4.

Related to: openhab/openhab-core#2574


cdjackson:
Thanks @wborn . I'm not really sure where the build is failing though - ci.openhab is building it fine, and it works ok here locally, but thanks for updating all the same.

Issue #733, opened 2022-01-24
Support extra configuration on Legrand / BTCino with Netatmo devices


pgaufillet:
The devices of this manufacturer offer optional behaviors (LED, dimmer). It should be possible to configure them.

As of OpenHAB 3.2, the following commands works in OpenHAB console:
* enable/disable Led in Dark: `zigbee write <node>/<endpoint> 0xfc01 1 [true|false] BOOLEAN`
* enable/disable Led if On: `zigbee write <node>/<endpoint> 0xfc01 2 [true|false] BOOLEAN`

Attribute Dimmer mode cannot be changed because its data type (0x09 / DATA_16_BIT) is not supported in OpenHAB.

More information at [Domoticz-Zigate-Wiki](https://github.com/zigbeefordomoticz/wiki/blob/master/en-eng/Corner_Legrand-Netatmo.md#cluster-0xfc01).

pgaufillet:
I am far from being an expert of this library so I am probably wrong :-)

This assumption is based on a simple search of `DATA_16_BIT` in its repository:
<img width="756" alt="image" src="https://user-images.githubusercontent.com/5228249/150837977-6113ad5d-dc54-451a-9745-2e25deaf8021.png">

You can see that in the method `parseValue()` the corresponding case is commented out.

And `zigbee write <node>/<endpoint> 0xfc01 0 0 DATA_16_BIT` in OpenHAB returns `Error: Exception during command execution (IllegalArgumentException): Data type DATA_16_BIT is not supported.`

cdjackson:
You're right, there is no serialisation defined for this data type. I will add that to the library and it will be available in a few weeks or so, but if you want support for these device I suggest to please try and add it to OH if you can.

pgaufillet:
Thank you for your help Chris! 

I'll try to implement a preliminary support for this non standard cluster 0xfc01 in the coming weeks.

cdjackson:
Just to note that if this is implemented "properly" - ie the device implements the ZCL on this cluster and these are just standard attributes, then all that is needed is a static definition - ie an XML file needs to be added. You don't need to add code.

Unfortunately it's not documented very well but maybe the following recent thread on the forum helps -:

https://community.openhab.org/t/sinope-zigbee-thermostat-th1123zb/129479/20

[...]

PR #736, opened 2022-02-03
Legrand with Netatmo cluster 0xfc01 support

pgaufillet:
Implement Legrand with Netatmo cluster 0xfc01 support as described in issue #733.

Signed-off-by: Pierre Gaufillet <pierre.gaufillet@bergamote.eu>

### DIFF
--- a/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/ZigBeeThingTypeMatcher.java
+++ b/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/ZigBeeThingTypeMatcher.java
@@ -26,10 +26,10 @@ import java.util.Map.Entry;
 import java.util.Set;
 
 import org.eclipse.jdt.annotation.NonNull;
+import org.openhab.binding.zigbee.ZigBeeBindingConstants;
 import org.openhab.core.config.core.ConfigDescriptionRegistry;
 import org.openhab.core.thing.ThingTypeUID;
 import org.openhab.core.thing.type.ThingType;
-import org.openhab.binding.zigbee.ZigBeeBindingConstants;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -127,7 +127,7 @@ public class ZigBeeThingTypeMatcher {
                     continue;
                 }
 
-                newProperties.add(new RequiredProperty(discoveryElement[0].trim(), discoveryElement[1].trim()));
+                newProperties.add(new RequiredProperty(discoveryElement[0].trim(), unescape(discoveryElement[1].trim())));
             }
 
             if (newProperties.isEmpty()) {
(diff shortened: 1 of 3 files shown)


org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/ZigBeeThingTypeMatcher.java
```diff
@@ -172,4 +172,30 @@ public RequiredProperty(String name, String value) {
...
         }
     }
+    
+	/**
+	 * Transforms \x&lt;nnnn&gt; occurrences into corresponding utf16 char.
+	 * 
+	 * @param s String that shall be unescaped.
+	 * @return UTF16 string
+	 */
+	private static final String unescape(String s) {
```
cdjackson, line 182: Why does this need to be static? I can't immediately see a need for that so if it's not required then I'd suggest to remove this.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/ZigBeeThingTypeMatcher.java
```diff
@@ -172,4 +172,30 @@ public RequiredProperty(String name, String value) {
             this.value = value;
         }
     }
+
```
cdjackson, line 175: It looks like this needs formatting as this method doesn't line up with the method above. It looks like tabs have been used rather than spaces?

cdjackson (changes requested):
Thanks for this. It generally looks fine to me - just a couple of small comments/questions/changes.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/ZigBeeThingTypeMatcher.java
pgaufillet: unescape is rather a utility method than a member of TigBeeThingTypeMatcher, as it does not apply to its properties but to a String argument. In this kind of case, I usually create a Util class for holding such methods. Here, with only one util method, keeping it in the user class looked to be the best choice.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/ZigBeeThingTypeMatcher.java
pgaufillet: I accidentally applied spotless settings to all the code and had to cancel it later. This is probably a resulting error. I am going to fix it.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/ZigBeeThingTypeMatcher.java
cdjackson: Ok, but I see it as just another method in the class and I think in general making all (or really, any) methods static is not highly desirable. If this was a helper class where you were providing static methods externally, then I could see some merit, but as it is it's all internal and private.

I'm happy for it to be in the user class - it certainly makes no sense for it to be elsewhere - I just don't see the need to make it static.

pgaufillet: Fixed. Removed "final" too after having thought about it.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/ZigBeeThingTypeMatcher.java
pgaufillet: I applied openhab formatter style to this method. It should be ok now. Same thing for the import order.

cdjackson:
Thanks for this @pgaufillet - looks good now.

cdjackson (approved):

Issue #740, opened 2022-03-02
Zigbee devices going offline with zigbee-addon 3.2.0


nailyk-fr:
## Description

Since some time I am using Openhab to control lights in my place. 
I migrated (new installation) from Openhab 2.x to Openhab 3.x (zigbee-addon 3.1.0) with quite success. However, my VM was to slow to handle my setup. 
Decided to buy a dedicated machine for Openhab and started a fresh setup with OH3.2.0 (zigbee-addon 3.2.0)
Since, it is a nightmare. 

Some devices (like IKEA bulbs) do not work at all. After pairing (and when controller leave pairing mode), devices appears offline. 
I thought it came from the controller so I bought a BitronVideo BV2010/10 on top of the Elelabs one. Behavior was mostly the same. 
Decided to upgrade the firmware of the controller (the BV one) and then faced another issue : #726 
I moved back to the Elelabs one and decided to replace my bulbs. 

Everything was working quite well for some days and, suddenly, all the devices have the same behavior : going offline. 

I tried as much as I can to reproduce this with my old VM (still with addon 3.1.0) but this 'going offline' behavior does not seems to occur. 

After a restart of the host, I can usually send one command before the device goes offline: 

### Before restart : 

```
2022-03-02 17:05:48.473 [INFO ] [openhab.event.ItemCommandEvent      ] - Item 'LIGHT2ndBedRTZ3000oborybowTS0502A_Color' received command 39

==> /var/log/openhab/openhab.log <==
2022-03-02 17:05:48.474 [DEBUG] [ng.zigbee.handler.ZigBeeThingHandler] - 60A423FFFE92CFCB: Command for channel zigbee:device:6fbabd96fd:60a423fffe92cfcb:60A423FFFE92CFCB_1_color --> 39 [PercentType]
2022-03-02 17:05:48.476 [DEBUG] [transaction.ZigBeeTransactionManager] - 60A423FFFE92CFCB: Creating new Transaction Queue
2022-03-02 17:05:48.479 [DEBUG] [e.transaction.ZigBeeTransactionQueue] - 60A423FFFE92CFCB: Set profile to ZigBeeTransactionProfile [maxOutstandingTransactions=1, interTransactionDelay=50, maxRetries=2]
[...]

PR #741, opened 2022-03-04
README.md: Add note about max zigbee_networksize for Elelab controller

nailyk-fr:
* As discussed in #740 a too big `zigbee_networksize` can overload controller memory and create issue in the Zigbee network. 
 * Add a warning about this in the `supported-coordinators` table.


cdjackson:
I would not suggest to add this here. The problem is this will be dependant on firmware - not the ELE dongle. If the firmware is not compiled properly to take advantage of the memory, then this is the result. If ELE change their firmware, this could be fixed - other ember devices will have exactly the same limitation.

It's worth noting that this issue is already written in the README - so your exact problem is documented there already ;)

nailyk-fr:
Thanks for your answer ! 

Sorry I Wanted to create this one as a draft. 
I actually read this section about network size, but it was unclear to me, this is why I did not understand it was source of this issue. 
I will rework the PR.

cdjackson:
Thanks - improvements are definitely appreciated :) Often I will write stuff quickly when a feature is added, so having users update this in a more "user orientated" way is much appreciated.

PR #749, opened 2022-03-18
Simplify battery voltage converter support detection (similar to Chri…

FranzSchi:
Hi Chris,
i would recommend to have the same behavior for battery voltage as for the battery percent converter.
Thx
/Franz

Simplify battery voltage converter support detection (similar to Chris change for battery percent converter)

Signed-off-by: FranzSchi <Franz.Schindler@web.de>


FranzSchi:
UAT successful 

image

PR #754, opened 2022-04-01
Changed reporting default value Humidity to Integer and to 0,5%

FranzSchi:
Changed reporting default value Humidity to Integer and to 0,5%

Signed-off-by: Franz Schi <Franz.Schindler@web.de>
Signed-off-by: Franz Schi Franz.Schindler@web.de


FranzSchi:
What I am doing wrong always to fail the DCO.
#IneedAComputerTraining
#NotSkilledWithComputers

cdjackson:
Thanks @FranzSchi . I also don't see why the DCO was failing - the signatures on both the PR and the commit looked fine to me so I marked it as passed and merged.

Thanks.

PR #756, opened 2022-04-13
Bump ZSmart Systems ZigBee library to 1.4.5

cdjackson:


cdjackson:
@wborn do you understand why CI is failing here? This builds fine locally, but fails at the last hurdle in Jenkins. I don't think it's caused by the change in library version (at least the Java all builds so maven is pulling the correct version) and the error doesn't appear related to this.

It's also strange that Jenkins lists a huge number of commits (42 of them) whereas this is just a single commit - it's as though it's comparing against an old main branch from October, but this appears to be configured ok in Jenkins (AFAICT).

wborn:
> do you understand why CI is failing here?

It seems to build fine now on Jenkins after I retriggered it.

cdjackson:
Strange - I restarted it a couple of times and it resulted in the same issue. Anyway, it's working now so I'll merge. This will require the updated dependencies....

Issue #743, opened 2022-03-12
Allow configuration of reporting parameters in Channel Details for all Attributes that support reporting


FranzSchi:
### Outline

As a OpenHab user I want to be empowered to configure in the Channel Details the Reporting properties similar to.

zigbee subscribe 19620/1 0x001 0x0020 120 900 0
zigbee subscribe 19620/1 0x001 0x0021 120 900 0

like show here:
![image](https://user-images.githubusercontent.com/18656932/158018647-b3b837a5-6b61-490b-96c6-472f8bdaf7cc.png)

### Configuration

All device attributes that support (or has setup a reporting)

cdjackson:
As requested in the other thread, please provide a list of the converters that require this enhancement.

FranzSchi:
Hi @cdjackson,

thanks a lot for your kind support I really appreciate this.
It sometimes takes some times for me to understand the technical terms, I hope this answers the question.

As mentioned in the Post https://community.openhab.org/t/zigbee-consecutive-polling-of-attributes/134166/2?u=franzs
in the OH Forum. I would suggest to put in.

- Minimum Reporting Period
- Maximum Reporting Period
- Polling Period
- Reportable Change

Into this Converters:
ZigBeeConverterBatteryPercent
ZigBeeConverterBatteryVoltage

ZigBeeConverterMeasurementPower
ZigBeeConverterMeasurementRmsCurrent
ZigBeeConverterMeasurementRmsVoltage

ZigBeeConverterMeteringSummationDelivered
ZigBeeConverterMeteringSummationReceived

ZigBeeConverterRelativeHumidity
ZigBeeConverterTemperature

ZigBeeConverterWindowCoveringLift


Thanks a lot!
/Franz

FranzSchi:
Hi @cdjackson ,
I've seen that this was partly done for Plugs.
But as i remember it is still not applied to to the device unless a known init error is given is not resolved I do not remember the issue number anymore.
right?

PR #764, opened 2022-06-05
Add Thermostat PI Heating/Cooling Demand Channels

lucasec:
This adds the "PI Heating Demand" and "PI Cooling Demand" standard thermostat channels. These channels are represented as a percentage (Dimensionless unit), and indicate whether there is a current heating or cooling call from the thermostat. They are fully supported by auto-discovery for the generic device type.

My electric line voltage thermostats generally report 0% for no heat and 100% for full heating power. They also report values in between when they are limiting the current to the heater. I am not aware of how low voltage thermostats handle this spec, e.g. do they only report 0 and 100 or do they come up with some number in between even when they only electrically support "on" and "off".

I used `Number:Dimensionless` as this seems to be the generally accepted way to represent a percentage under the units of measurement system. I see some ongoing debate in https://github.com/openhab/org.openhab.binding.zigbee/pull/597 wrt. the relative humidity channel. If desired, I can submit a separate PR to update the humidity channel to the same type.

### DIFF
--- a/org.openhab.binding.zigbee/README.md
+++ b/org.openhab.binding.zigbee/README.md
@@ -408,6 +408,8 @@ The following channels are supported -:
 | thermostat_systemmode        | `THERMOSTAT` (0x0201)                    | Number                   |             |
 | thermostat_unoccupiedcooling | `THERMOSTAT` (0x0201)                    | Number                   |             |
 | thermostat_unoccupiedheating | `THERMOSTAT` (0x0201)                    | Number                   |             |
+| thermostat_heatingdemand     | `THERMOSTAT` (0x0201)                    | Number:Dimensionless     |             |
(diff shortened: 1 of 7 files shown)


cdjackson (commented):
Thanks. Looks generally good with just a couple of small comment.
I'd like to get the reporting configuration implemented so it's one less thing to do in #743

lucasec:
There seem to be a few surprises with the reporting config, trying to debug.

lucasec:
Disregard, it seems my zigbee stick got into some kind of loop and needed to be power cycled. Confirmed the reporting configuration is working as expected. Squashed commits to prepare for merge.

cdjackson:
Thanks. Looks good.

PR #765, opened 2022-06-06
Add units to Illuminance and Relative Humidity, fix missing units

lucasec:
This PR should include all code in https://github.com/openhab/org.openhab.binding.zigbee/pull/597, plus a few other places I found the units missing in dynamically defined channels.

It adds new units to the following channel types:
- Relative humidity is now defined as a `Number:Dimensionless`
- Illuminance is now defined as a `Number:Illuminance`

Since the addition of units may require things to be deleted/re-created, this may be a breaking change.


org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterMeasurementPower.java
```diff
@@ -145,7 +145,7 @@ public Channel getChannel(ThingUID thingUID, ZigBeeEndpoint endpoint) {
         return ChannelBuilder
                 .create(createChannelUID(thingUID, endpoint,
                         ZigBeeBindingConstants.CHANNEL_NAME_ELECTRICAL_ACTIVEPOWER),
-                        ZigBeeBindingConstants.ITEM_TYPE_NUMBER)
+                        ZigBeeBindingConstants.ITEM_TYPE_NUMBER_POWER)
```
lucasec, line 148: I'm assuming when we dynamically generate channels, e.g. for a generic device without a thing XML definition, we ignore the contents of `channels.xml` completely? Thus without changing the type here we were not actually getting the correct channel type?

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterRelativeHumidity.java
```diff
@@ -65,15 +74,17 @@ public boolean initializeDevice() {
...
             CommandResult bindResponse = bind(serverCluster).get();
             if (bindResponse.isSuccess()) {
                 // Configure reporting
                 ZclAttribute attribute = serverCluster
                         .getAttribute(ZclRelativeHumidityMeasurementCluster.ATTR_MEASUREDVALUE);
-                CommandResult reportingResponse = attribute
-                        .setReporting(REPORTING_PERIOD_DEFAULT_MIN, REPORTING_PERIOD_DEFAULT_MAX, 50).get();
-                handleReportingResponse(reportingResponse, POLLING_PERIOD_DEFAULT, REPORTING_PERIOD_DEFAULT_MAX);
+                CommandResult reportingResponse = attribute.setReporting(reporting.getReportingTimeMin(),
+                        reporting.getReportingTimeMax(), reporting.getReportingChange()).get();
```
lucasec, line 86: I'm not exactly clear on the lifecycle order—would the `CHANGE_DEFAULT` configured on line 115 be picked up here? As in the converter is initialized first then when we create the `reporting` it reads the already defined default parameter values?

lucasec (commented):
Adding a few open questions I have

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterMeasurementPower.java
cdjackson: No - the channels XML file is not ignored. The channel type definitions are still required. When you define a channel (either programatically or in the thing XML file) you have to specify the channel-type, but the channel can also override certain aspects of this.  I'm pretty sure that if you try and define a channel that doesn't have a channel type defined, it will error.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterRelativeHumidity.java
cdjackson: No - it won't - it will use whatever the default is in the `ZclReportingConfig` class.

lucasec: I'm assuming the device does not immediately get reconfigured later with the actual default either (unless the user goes in and changes/saves the channel options)?
So I probably need to refactor the pattern that was used in other converters somewhat to ensure it uses a reasonable default

Issue #775, opened 2022-10-01
Polling not working with Tradfri E27 bulb


weakfl:
### Outline

I have a Tradfri E27 dimmable light bulb that no longer answers to polling requests (it has worked fine in the past). I've tried changing the polling interval to 70s and back to 60s, and setting the interval seems to be successful. 

But polling still times out:

```
2022-10-01 09:34:25.144 [DEBUG] [ng.zigbee.handler.ZigBeeThingHandler] - 000B57FFFEB1BAB5: Polling [zigbee:device:c3cd543626:000b57fffeb1bab5:000B57FFFEB1BAB5_1_dimmer, zigbee:device:c3cd543626:000b57fffeb1bab5:000B57FFFEB1BAB5_1_colortemperature] channels...
[...]
2022-10-01 09:34:54.586 [DEBUG] [.zigbee.handler.ZigBeeIsAliveTracker] - IsAlive Tracker LastChance Timeout has been reached for thingUID=zigbee:device:c3cd543626:000b57fffeb1bab5
2022-10-01 09:34:54.590 [DEBUG] [ng.zigbee.handler.ZigBeeThingHandler] - 000B57FFFEB1BAB5: Polling stopped
2022-10-01 09:34:54.595 [DEBUG] [ng.zigbee.handler.ZigBeeThingHandler] - 000B57FFFEB1BAB5: Polling initialised at 61557ms
```

Reading attributes through the console however works as expected:

```
openhab> zigbee node 32147
IEEE Address     : 000B57FFFEB1BAB5
Network Address  : 32147
```

```
openhab> zigbee read 32147/1 0008 0                                                                                                                                                                                                                                      
Reading endpoint 7D93/1, cluster server cluster Level Control (0008), attributes Current Level
Response for cluster 0x0008
Attribute    0  Current Level                                       UNSIGNED_8_BIT_INTEGER          254
```

I'm also seeing some AshFrameHandler exceptions, probably unrelated though:

```
2022-10-01 09:32:11.231 [ERROR] [e.ember.internal.ash.AshFrameHandler] - AshFrameHandler Exception processing EZSP frame:
java.lang.IllegalArgumentException: No reader defined in ZigBeeDeserializer for UNSIGNED_64_BIT_INTEGER (0x27)
	at com.zsmartsystems.zigbee.serialization.DefaultDeserializer.readZigBeeType(DefaultDeserializer.java:345) ~[bundleFile:?]
[...]

PR #780, opened 2022-10-07
Revert channel link test to use ChannelUID

cdjackson:
In the base handler the call to the registry use the UID so it seems best to continue to use this here.


cdjackson:
#775

PR #782, opened 2022-10-15
Fix dependencies and headers in firmware provider

cdjackson:
Fixes build, reformats all headers and updates log messages.

@kaikreuzer sorry for the problem with this - I should have run another test before merging this.


kaikreuzer (approved):
No worries, many thanks for the quick clean-up!

PR #783, opened 2022-10-18
Add timestamp to duplicate command detection

dschall:
This change keeps track of the timestamp of the last transaction sent by a Tuya wireless button.
Sometimes, these devices send duplicate commands, and we need to ignore these duplicates.
After some inactivity, the transaction ids restart at 1, so we also need to track the timestamp of the last transaction and ignore only those duplicates that happened in a short time (currently set to 5 minutes).

### DIFF
--- a/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterTuyaButton.java
+++ b/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterTuyaButton.java
@@ -12,8 +12,10 @@
  */
 package org.openhab.binding.zigbee.internal.converter;
 
-import static java.lang.Integer.*;
-
+import java.time.Duration;
+import java.time.LocalDateTime;
+import java.time.Month;
+import java.time.temporal.TemporalUnit;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


cdjackson:
I'm a bit surprised that this is set to minutes. The zigbee library handles duplicate transactions, but that is normally caused by APS retries so it's always in the millisecond range (it's the normal requirement in the APS layer).

I assume this is a bug in the device itself if it's resending the same command after 5 minutes. I guess my other question is how reliable this is, how often it happens, and if the consequences of incorrectly dropping a frame here is worse that the consequences of the original duplicate issue?

dschall:
These are my log settings right now (omitting unrelated log settings):
```
ROOT                                               │ WARN
com.zsmartsystems.zigbee                           │ DEBUG
com.zsmartsystems.zigbee.dongle.ember.internal.ash │ INFO
org.openhab                                        │ INFO
org.openhab.binding.zigbee                         │ DEBUG
```

Does that cover the data you'd like to see?

cdjackson:
Yes, that should be fine - the package is `com.zsmartsystems.zigbee.aps` so that should be at `DEBUG` based on this.

Thanks.

dschall:
Hey @cdjackson , I finally captured a duplicate command.
It happend after the Tuya device ran out of battery and was left offline for a few days.
When I replaced the batteries and pushed a button, I got two events on OH, see log attached (timestamp 16:27:21.737, look for "TuyaButtonPressCommand")

[duplicate command after reconnect.txt](https://github.com/openhab/org.openhab.binding.zigbee/files/10173123/duplicate.command.after.reconnect.txt)

dschall:
PS: These Tuya thingys have other issues, mine eat through batteries like crazy.
I'll follow up on the thread with more logs:
https://community.openhab.org/t/adding-support-for-tuya-zs3l-tz3000-ts0043-ts0042-ts0041-to-zigbee-bindings/108895/61

cdjackson:
Sorry - I should have read your message better ;)

dschall:
Yeah, sorry, now that you're pointing it out..
Looks like my change would not even fix that, then.

On the positive side, it never happend during my testing (before the batteries were empty), so maybe we can live with this as a corner case when the buttons are pressed while the device is rejoining after being offline for a few days.
I'll collect logs about the battery drain. I noticed the Tuya device is asking for the TIME cluster every 10 minutes, and it's not implemented on OH side (I think, not sure if I'm reading the logs right).

PR #785, opened 2022-12-15
Fix Zigbee name stylization

jlaur:
Name was changed some years ago from ZigBee to [Zigbee](https://en.wikipedia.org/wiki/Zigbee). Maybe it's time to adapt.

Changed spelling in:
- Documentation
- Code comments.
- Log entries.
- User-oriented texts.
- Bundle names and descriptions.

Did not change spelling in:
- Class names etc.
- "QIVICON ZigBee-Funkstick" as it seems like a product name (probably old).
- all the places I've missed...

See also openhab/openhab-addons#13954

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -2,7 +2,7 @@
 
 Delete this section in your issue...
 
-This issue list is to report bugs or enhancements for the openHAB ZigBee binding. If you're not sure about something, please consider discussing it on the [forum](http://community.openhab.org) before raising an issue - more people follow the forum than monitor the issues list, so if you are having trouble with a device, or want to ask a question, you will have a much larger audience there.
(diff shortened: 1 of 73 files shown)


wborn (approved):
:+1:

cdjackson:
While I appreciate that this is not in line with the correct Zigbee name, I'm not super keen to change this. I don't mind changing the external facing names (ie text strings), but changing internal stuff like logging is potentially a pain as it has flow on effects to (for example) log processing scripts. I don't really think it matters that this is different to the Zigbee official name - and actually in theory Zigbee is trademarked and probably we shouldn't use it ;)

The styling `ZigBee` is used by all the external libraries and log processing scripts which is why it was used in the binding (since the libraries were written initially at the same time). At the time I was warned by the Alliance that I shouldn't use `Zigbee` as it was their trademark.

So, my preference is not to change this - or at least not to make code changes other than for external user facing strings...

jlaur:
OK, I'll not insist.

PR #787, opened 2022-12-26
Add support for occupied heating in thermostat client

cdjackson:
Adds support for client cluster for the occupied heating. This will add the occupied heating channel if the client thermostat cluster is found - no further checks are done. This might mean it's added when the device doesn't support it, but the alternative is that if we add checks to ensure it is actually supported, it might not be added for devices that don't report correctly.


kaikreuzer:
@cdjackson We usually merge always against main and then cherry-pick issues back to the 3.4.x branch and mark them with the "patch" label. This way, we can ensure that every fix exists on main and that our patch release notes gather the correct PRs.

cdjackson:
Sure, but at the moment I don't know if this PR works as I can't test it since I don't have the device in question. The user who can test it is using OH3.4. I would prefer that he tested it before I merge this into the master.

kaikreuzer:
Best send him a custom build jar for testing (even before merging such a fix).

cdjackson:
Well, i was hoping to use the normal snapshot which is easier for most people who are not so familiar with development processes as we can point them at standard processes. However at the moment it doesn't look like there is a OH3.4.x build available?

PR #793, opened 2023-01-15
Adjust to core changes (addon.xml)

J-N-K:
ONLY merge after the corresponding code PR (https://github.com/openhab/openhab-core/pull/3050) has been merged.

Signed-off-by: Jan N. Klug <github@klug.nrw>

### DIFF
--- a/org.openhab.binding.zigbee/src/main/resources/OH-INF/addon/addon.xml
+++ b/org.openhab.binding.zigbee/src/main/resources/OH-INF/addon/addon.xml
@@ -1,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<addon:addon id="zigbee" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xmlns:addon="https://openhab.org/schemas/addon/v1.0.0"
+	xsi:schemaLocation="https://openhab.org/schemas/addon/v1.0.0 https://openhab.org/schemas/addon-1.0.0.xsd">
+
+	<type>binding</type>
+	<name>ZigBee Binding</name>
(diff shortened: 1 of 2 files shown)


org.openhab.binding.zigbee/src/main/resources/OH-INF/addon/addon.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<addon:addon id="zigbee" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xmlns:addon="https://openhab.org/schemas/addon/v1.0.0"
+	xsi:schemaLocation="https://openhab.org/schemas/addon/v1.0.0 https://openhab.org/schemas/addon-1.0.0.xsd">
+
+	<type>binding</type>
+	<name>ZigBee Binding</name>
+	<description>This is the binding for ZigBee providing a vendor neutral ZigBee implementation supporting many different
+		ZigBee devices.</description>
+	<author>Chris Jackson</author>
```
kaikreuzer, line 10: Following the new rule that we do not use this field for individual authors, shouldn't we rather remove this line, just as we did for all bindings in openhab-addons?

J-N-K: This rule is not new, we already had it before, I don‘t know why it was not adopted here, so I left it as-is.

kaikreuzer: The reason is simple - we simply forgot it, because it was in a different repository.
But you are right, it is not directly related to this PR, so I'll create a separate PR for it.

kaikreuzer (approved):

PR #798, opened 2023-02-14
Fix build

J-N-K:
Signed-off-by: Jan N. Klug <github@klug.nrw>


kaikreuzer (approved):
Thanks for fixing the build!

Out of curiosity: Why do we need this new file all of a sudden?

J-N-K:
The issue is a bug in spotbugs that reports an error when using `foo instanceof Bar bar` checks and that needs a suppression (`SA_LOCAL_SELF_COMPARISON`). If we only add that suppression, then other issues pop up and to we need to add some more (not all that are added here), but for consistency I decided to copy the file from openhab-core to openhab-addons and that somehow caused the build to fail here because of the changes in the parent POM.

PR #810, opened 2023-10-14
[DOCS] Zigbee: Fix Link of Docker Installation

felixschndr:
Link to Docker installation guide was missing a `.html`


felixschndr:
Apparently the link was already correct in the GitHub repo but is not working on the [actual page](https://www.openhab.org/addons/bindings/zigbee/#configuring-a-coordinator) (at the end of the paragraph)...

cdjackson:
What link are you referring to exactly?  If its the one that says "See Docker Install Guide" then it's working fine for me.

felixschndr:
The link "See Docker Install Guide": The link brings the user to the webpage however the anker gets lost. The link points towards
```
https://www.openhab.org/docs/installation/docker#explanation-of-arguments-passed-to-docker
```
However it should be 
```
https://www.openhab.org/docs/installation/docker.html#explanation-of-arguments-passed-to-docker
```

So it brings one to the right page but doesn't scroll down to the section

PR #824, opened 2024-01-04
Update README.md - Sonoff Zigbee 3.0 USB Dongle Plus V2

Boldfor:
Updated the Sonoff Zigbee 3.0 USB Dongle Plus V2-entry for clarity.


Boldfor:
Please advise on whether I shall close this PR as well.

Boldfor:
@cdjackson, quick question: I remember you had a different opinion on my suggestion for the intro of the readme (the intention was: easier to understand for non-pros).

With this suggestion (which you also did not merge), however, I have problems to see the problem. Could you maybe elaborate?

cdjackson:
Two reasons - one you state in the Ember line that the TI chips don't work with this binding - this is true, but it's not related to Ember and it's already in the docs. Second you state that the CP2102 doesn't work - the CP2102 is also used with the Ember chip in new devices. I'm working with Sonoff on some issues, but the CP2102 is used extensively in other Ember dongles and works ok.

Boldfor:
Thanks.

Do you believe there's value in including these information (incl. with the ones that I included in my proposal) in the documentation?

PR #827, opened 2024-01-06
Set parent to 4.1.1-SNAPSHOT

wborn:
This may help @kaikreuzer


kaikreuzer (approved):
I should remember this when bumping the versions the next time...

wborn:
> I should remember this when bumping the versions the next time

It probably only matters if you want to build the branch yourself as the sandbox build has no issues with it, see:

https://ci.openhab.org/view/Sandbox/job/sandbox-openhab4-patch-release/

PR #850, opened 2024-07-19
Add support for PM2.5 channel

cdjackson:
This adds support for the PM2.5 channel. This is currently untested and must be built with the latest snapshot (1.4.15) of the ZSmart Systems ZigBee library since the PM2.5 cluster support is added there.


rdslw:
@cdjackson will it make to next 4.3 (4.3.6?) release of openhab, or only 5.0.0 ?

cdjackson:
This is only merged into main branch so it will only be in the 5.0.0 release.

PR #870, opened 2024-10-31
Add reporting configuration for temperature and humidity channels

cdjackson:


FranzSchi:
Hi Chris, this feature looks very valueable to me, similar to the configuration of power measurement.
Are you planning to merge this also in soon?

cdjackson:
Yes, this is all coded, but I've not had time to test. Hopefully I'll do this in the next week or so.

FranzSchi:
Thanks for your feedback, i thought about if those are also affected by the „long“ reporting config change.

cdjackson:
No - different attributes have different formats

cdjackson:
Unfortunately this won't work until https://github.com/openhab/openhab-webui/issues/2882 is fixed because the thing gets reinitialised rather than updating the configuration.

I will merge this in the coming days in the hope that the UI gets fixed.

FranzSchi:
Sorry to ask that,
But maybe there can be an improved description for this.
There is the Config
“Report On Change”
The minimum change of the attribute value needed to trigger a device state update

Is it like that that this is an absolute number or is that dependent on the Data type?
So simply said does 5 mean
5 Degrees, 0.5. Degrees?
Or in Volt does 5 Mean
5 Volt or 0.5 volt or 0..05 volt?

cdjackson:
This is on my list of things to do, but as it requires a larger, more widespread change to the binding to support the improved descriptions, I won't add it to this PR. I plan on adding a description that will be shown in the UI below the input box.

For many of these channels there is a /100 (so temperature and humidity work this way) so 50 would be 1/2% for example.

PR #883, opened 2025-01-01
Update README.md

curlyel:
Recently ordered ITEAD Sonoff ZBDongle-E came with Sofware flow control. Therefore the documentation should reflect this otherwise it would be misleading (as in my case ;-)

Added "Software flow control" to the ZBDongle-E line.


cdjackson:
Thanks

PR #855, opened 2024-09-15
[zigbee] Map colour temperature states to color HSB states

andrewfg:
Previously, when a lamp device was operated in color temperature mode, the color channel/item state within OH and its respective UIs would be shown as UNDEF.

Now, with this PR when a lamp is operated in color temperature mode, the respective color temperature is converted to an actual valid color that is derived from the color temperature point on the color temperature locus on the CIE color xy chart. See curved line on CIE chart below.

**NOTA BENE**: this calculated HSB state is **NOT** sent to the remote device (since that would lead to a feedback loop) -- the state is used solely within OH and its respective UIs. However I think it may be be necessary to set `autoUpdate=false` on the respective color channel (or implement another mechanism) in order to definitively rule out feedback loops.

image

The initial aim of this PR was simply to resolve #853 (which it probably does). However it also (therefore) eliminates a lot of state setting that was previously conditionally required due to the UNDEF state when in color temperature mode. And (therefore) it also eliminates the JUnit tests that previously tested that conditional state setting code.

As this change is a little more radical than others might have anticipated, I am putting this PR up here, with the request that you kindly let me have your thoughts on it. We may need a few iterations..

This PR depends on https://github.com/openhab/openhab-core/pull/4367

Signed-off-by: AndrewFG <software@whitebear.ch>

PR #889, opened 2025-01-02
Fix regression with managing on/off state in color handler

### LINKED ISSUE #888: Colour channels are updated when device is off
This appears to be introduced in #855. The channel state can be updated even when the device is off and this leads to incorrect status in OH

https://github.com/openhab/org.openhab.binding.zigbee/pull/855/files#r1901193799

cdjackson:
This fixes a regression introduced in #855 which removed the management of the on/off state of a device and therefore propagates incorrect states into openHAB

Closes #888

### DIFF
--- a/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterColorColor.java
+++ b/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterColorColor.java
@@ -23,6 +23,7 @@ import java.util.concurrent.Future;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
@@ -98,6 +99,8 @@ public class ZigBeeConverterColorColor extends ZigBeeBaseChannelConverter implem
     private int lastY = -1;
     private boolean xChanged = false;
     private boolean yChanged = false;
+
+    private final AtomicBoolean currentOnOffState = new AtomicBoolean(true);
 
     private double lastMired = -1;
 
@@ -441,6 +444,8 @@ public class ZigBeeConverterColorColor extends ZigBeeBaseChannelConverter implem
 
     private void updateOnOff(OnOffType onOff) {
         boolean on = onOff == OnOffType.ON;
+        currentOnOffState.set(on);
+
         // Extra temp variable to avoid thread sync concurrency issues on lastHSB
         HSBType oldHSB = lastHSB;
         HSBType newHSB = on ? lastHSB : new HSBType(oldHSB.getHue(), oldHSB.getSaturation(), PercentType.ZERO);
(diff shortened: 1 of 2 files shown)


org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterColorColor.java
```diff
@@ -452,15 +457,21 @@ private void updateBrightness(PercentType brightness) {
         HSBType oldHSB = lastHSB;
         HSBType newHSB = new HSBType(oldHSB.getHue(), oldHSB.getSaturation(), brightness);
         lastHSB = newHSB;
-        updateChannelState(newHSB);
+
```
andrewfg, line 460: Probably you should also update the HSB when the device is off.

Instead of this..
```
HSBType newHSB = new HSBType(oldHSB.getHue(), oldHSB.getSaturation(), brightness);
```
Try this..
```
HSBType newHSB = new HSBType(oldHSB.getHue(), oldHSB.getSaturation(), currentOnOffState.get() ? brightness :0);
```

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterColorColor.java
```diff
@@ -452,15 +457,21 @@ private void updateBrightness(PercentType brightness) {
...
+        }
     }
 
     private void updateColorHSB(DecimalType hue, PercentType saturation) {
         // Extra temp variable to avoid thread sync concurrency issues on lastHSB
         HSBType oldHSB = lastHSB;
         HSBType newHSB = new HSBType(hue, saturation, oldHSB.getBrightness());
         lastHSB = newHSB;
-        updateChannelState(newHSB);
+
```
andrewfg, line 471: Ditto..

Probably you should also update the HSB when the device is off.

Instead of this..
```
HSBType newHSB = new HSBType(oldHSB.getHue(), oldHSB.getSaturation(), oldHSB.getBrightness());
```
Try this..
```
HSBType newHSB = new HSBType(oldHSB.getHue(), oldHSB.getSaturation(), currentOnOffState.get() ? oldHSB.getBrightness() :0);
```

andrewfg (changes requested):
I see my error. But I think your code is also not optimal. See comments below.

org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterColorColor.java
andrewfg: But I must admit that my proposal does therefore mean that the brightness is not persisted across an on/off/on toggle. Hmm.

andrewfg: So probably you need to first set `lastHSB` to persist the actual brightness. And after that you can call `updateChannelState` with the modified brightness from the ternary operator.

cdjackson: I guess it's firstly important to note that this bug was introduced in #855. Prior to that this worked fine (or at least acceptably). My focus here is not to change the logic, but to fix the regression introduced in #855. This converter went through a lot of review by a number of people when it was written and I'd prefer not to change logic in this regression fix.

> Probably you should also update the HSB when the device is off.

The HSB is updated locally - I guess you mean it should be updated in OH. Personally, I would avoid this in this PR. It will be a change in functionality that could be annoying or confusing, and ultimately it doesn't actually matter if the device is not on.

andrewfg: Ok.

andrewfg (approved):
LGTM

cdjackson:
Thanks @andrewfg - and just to note that I'm not against discussing/making the proposed change in a separate PR if you think it's worthwhile - I just didn't want it hidden in this regression fix PR...

andrewfg:
@cdjackson Ok. Shall I open such PR? For OH 5.x only..

PR #891, opened 2025-01-03
Add dimmer/switch mode for Legrand with Netatmo dimmers switch w/o neutral

pgaufillet:
Implement Dimmer/Switch mode configuration for Legrand Céliane with Netatmo dimmer switches (https://github.com/openhab/org.openhab.binding.zigbee/issues/733) now that DATA_16_BIT ZCL type (id: 0x09) is fully supported. Successfully tested on OpenHAB 4.3.1.

### DIFF
--- a/org.openhab.binding.zigbee/src/main/resources/OH-INF/thing/legrand/dimmer-switch-without-neutral.xml
+++ b/org.openhab.binding.zigbee/src/main/resources/OH-INF/thing/legrand/dimmer-switch-without-neutral.xml
@@ -27,6 +27,15 @@
 				<label>MAC Address</label>
 			</parameter>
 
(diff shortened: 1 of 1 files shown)


cdjackson:
I'm not completely sure we should put this into 4.3.1, but since it's a small change I'm not against it if needed but we should also therefore produce a PR against main as well, or this will be lost...

So, either make this PR against main, OR, provide another PR against main...

pgaufillet:
I have just uploaded the second PR against main. I am not expecting to have it especially in 4.3.x, so feel free to reject it if you are not comfortable with it. 

That said, this small piece of configuration is the last part closing issue #733 after 2 years: I am really happy to push it 👍

cdjackson:
Thanks - that's merged into both branches. For 4.3 it's small and self contained, so I think it's fine.

PR #894, opened 2025-01-05
Update Color Channel Hue and Saturation when light is off

### LINKED ISSUE #893: Update the color channel HSB state when light is turned off.
See #889

Currently the binding only updates the color HSB channel when the light is on, where the B part of the HSB channel shows the actual brightness. When the light is turned off, the B part of the color HSB channel is not set to zero, but retains the prior B value.

EDIT: this has the effect that a Switch item linked to the color HSB channel still shows on when the light is actually off.

In future the B part of the color HSB channel in OH should toggle b_actual=>0=>b_actual as the light is toggled on=>off=>on.

andrewfg:
Resolves #893 

Signed-off-by: Andrew Fiddian-Green <software@whitebear.ch>


cdjackson:
Thanks @andrewfg

PR #900, opened 2025-01-31
Fix Nullable issue in OH5 build

holgerfriedrich:
As reported [here](https://community.openhab.org/t/oh-5-0-snapshot-builds-failing/162137), our OH5 sandbox builds currently fail when building the zigbee addon.

On the local machine it builds against 4.3 and succeeds.
When I change `pom.xml` to use `5.0-SNAPSHOT`, I can reproduce the issue seen in the [Jenkins builds](https://ci.openhab.org/view/Sandbox/job/sandbox-openhab5-release/).

The proposed fix is probably not the most elegant way to do it.
However, compilation passes with OH5 and 4.3.


cdjackson:
Thanks @holgerfriedrich

PR #916, opened 2025-08-08
Update parent to 5.1.0-SNAPSHOT

wborn:
It will cause issues eventually when the wrong parent is used.


wborn:
It's probably similar to: https://github.com/openhab/org.openhab.binding.zwave/pull/1913

Issue #932, opened 2026-01-17
OnCommand / OffCommand does not trigger "core.ItemCommandTrigger" on the item


joerg1985:
I have a IKEA button here with an On / Off button which does report itself as Dimmer.
A rule should trigger as soon as On / Off is pressed regardless of the current item state.

```
configuration: {}
triggers:
  - id: "3"
    configuration:
      itemName: Auto_Klima_Button_NON_COLOR_CONTROLLER_Level_Control
    type: core.ItemCommandTrigger
conditions: []
actions:
  - inputs: {}
    id: "1"
    configuration:
      icon: fire
      message: Test
    type: notification.SendExtendedBroadcastNotification
```

The button does handle the zigbee command and the status of the item does change.
But there is no command send to the item, so the rule does not trigger.

<img width="1383" height="235" alt="Image" src="https://github.com/user-attachments/assets/13930bcd-be7f-4cc0-908f-b835b6d8a27c" />

For most zigbee items like temperature this is okay, but a button pressed shoud in my mind send a command.

cdjackson:
Can you do a fingerprint for this device please (console `zigbee fingerprint 2623` should do it)

joerg1985:
I did setup a zigbee stick to my dev environment and will try to fix the issues with related to the switch level.

The first related issue is a `NPE` when moving the slider in the UI, so the new level is not send to the dimmer.
The `configLevelControl` field is null in client mode, so this will raise in `ZigBeeConverterSwitchLevel:431`.

cdjackson:
There is already an event class that can be used to pass through button events - maybe take a look at that?

PR #936, opened 2026-03-02
trigger channel events for ON/OFF

joerg1985:
This PR will add support for channel events to the new `Level Control Events` and `Switch Events` channels.
Allowing to create rules triggered everytime a button on a `Level Control` or `Switch` ignoring the state of the channel.

To have trigger different rules depending on the times the button is pressed there are the following events supported:
- `SINGLE_ON`: Pressed `On` once within a short time
- `DOUBLE_ON`: Pressed `On` twice within a short time
- `TRIPLE_ON`: Pressed `On` thrice within a short time
- `SINGLE_OFF`: Pressed `Off` once within a short time
- `DOUBLE_OFF`: Pressed `Off` twice within a short time
- `TRIPLE_OFF`: Pressed `Off` thrice within a short time
- `SINGLE_TOGGLE`: Pressed `Toggle` once within a short time
- `DOUBLE_TOGGLE`: Pressed `Toggle` twice within a short time
- `TRIPLE_TOGGLE`: Pressed `Toggle` thrice within a short time

This has been tested with a `IKEA RODRET`.

### DIFF
--- a/org.openhab.binding.zigbee/README.md
+++ b/org.openhab.binding.zigbee/README.md
@@ -491,6 +491,19 @@ then
 end
 ```
 
+### Dimmer events
+
+`Level Control Button Events` channels support the following events:
+
+- `SHORT_PRESSED`: Pressed once within a short time
+- `DOUBLE_PRESSED`: Pressed twice within a short time
+
+These events can be combined with `But only if` of the item state to run rules depending on the button pressed, e.g.
+- a rule triggered on `SHORT_PRESSED` with `But only if dimmer_state != 0` to run on `ON` button pressed once
+- a rule triggered on `SHORT_PRESSED` with `But only if dimmer_state = 0` to run on `OFF` button pressed once
+- a rule triggered on `DOUBLE_PRESSED` with `But only if dimmer_state != 0` to run on `ON` button pressed twice
+- a rule triggered on `DOUBLE_PRESSED` with `But only if dimmer_state = 0` to run on `OFF` button pressed twice
+
 ## Reporting and Polling
 
 ZigBee has a standard way of configuring how a device sends status reports to the binding - this is called Reporting. Reporting is configured using three pieces of information -:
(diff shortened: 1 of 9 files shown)


cdjackson:
Sorry I've been travelling a lot over the past few weeks so have had little time to look at this.

I've really only had a quick look over this, but are you aware of the existing button event functions for remotes etc? I'm wondering if we shouldn't be extending that if necessary rather than coming up with these special on/off, or level event handlers. Where do we stop with this? We could end up duplicating this for a bunch more clusters which feels a bit strange given that the events are the same.

I'm just wondering if we really need this functionality, then can we make it generic like the existing events? I'm also not so keen on the binding generating derived events (such as the multi press events), but if we can find a nice way to make it generic it would be better and it might be ok.

joerg1985:
> Sorry I've been travelling a lot over the past few weeks so have had little time to look at this.

No need to be sorry, we are all doing this on our free time.

> I've really only had a quick look over this, but are you aware of the existing button event functions for remotes etc? I'm wondering if we shouldn't be extending that if necessary rather than coming up with these special on/off, or level event handlers. Where do we stop with this? We could end up duplicating this for a bunch more clusters which feels a bit strange given that the events are the same.

Yes, i was aware of these, but the dimmer has two buttons, so i need a way to add more payload to the events.
This is not possible as fair as i can tell, so i did created new events.

> I'm just wondering if we really need this functionality, then can we make it generic like the existing events? I'm also not so keen on the binding generating derived events (such as the multi press events), but if we can find a nice way to make it generic it would be better and it might be ok.

I could drop the `TRIPLE_*` events and use the exisiting `SHORT_PRESSED` and `DOUBLE_PRESSED` events with different channels. So one channel for each button, how about `..._on_events`, `..._off_events`, `..._toggle_events`?

This would still allow to:
- run a rule as soon a button is pressed on a dimmer (see [comment](https://github.com/openhab/org.openhab.binding.zigbee/issues/932#issuecomment-3969429476) in the original issue ticket)
- use one button for more than one case

joerg1985:
I have now simplified the events, there are only two new channels for dimmers and the existing events are used instead of defining new once. So i hope this get's closer to a potential merge :D

joerg1985:
I have found a way to only use one event channel for both buttons and updated the PR, see the updated README.md for details.

joerg1985:
@cdjackson could you hava a look at this, i think the PR is now in a good state, mutch simpler than before. It does use the existent events and only one new channel.

PR #938, opened 2026-03-16
Fix for PM25 converter ClassCastException

HubbleAstroid:
Minor change to handling double value from PM25 to fix Ikea PM2.5 monitor which causes java.lang.ClassCastException.


org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/internal/converter/ZigBeeConverterPM25.java
```diff
@@ -172,7 +172,9 @@ public void attributeUpdated(ZclAttribute attribute, Object val) {
         if (attribute.getClusterType() == ZclClusterType.PM2_5_MEASUREMENT
                 && attribute.getId() == ZclPm25MeasurementCluster.ATTR_MEASUREDVALUE) {
             logger.debug("{}: ZigBee attribute reports {}", endpoint.getIeeeAddress(), attribute);
-            updateChannelState(new DecimalType(Math.pow(10.0, (Integer) val / 10000.0) - 1));
+
+            double pm25Value = ((Number) val).doubleValue();
+            updateChannelState(new DecimalType(Math.pow(10.0, pm25Value / 10000.0) - 1));
```
cdjackson, line 177: I'd suggest to make this all a single line to avoid creating the varialbe pm25Value

